PR #5, opened 2022-08-26
fix(tokenizer): correctly use character offsets

sno2:


tokenizer/src/lib.rs
```diff
@@ -92,11 +92,15 @@ pub struct TokenStream<'a> {
...
+    /// Creates a new token stream with the given buffer as the source.
+    ///
+    /// ## Notes
+    /// - The buffer length must not be over the size of a `u32`.
     pub fn new(buffer: &'a [u8]) -> Self {
         Self {
             buffer,
             index: 0,
-            len: u32::try_from(buffer.len()).expect("[todo: better error]"),
+            len: buffer.len() as u32,
```
Phosra, line 103: This is incorrect, the try_from was fine?

Phosra (changes requested):

tokenizer/src/lib.rs
sno2: It's the user's job (module loader) to ensure the length of the source is not over 4 gigabytes. I've added this to the function's documentation.

Phosra: The code should be written in a self-documenting manner, it ought to accept a pointer and `u32` if that's how we're doing it - the user should provide the u32 and accept that it's their code caused the bug, not ours.

Phosra: I believe that keeping the `try_from`, and returning a `Result<Self, TooLargeFile>` sounds more sensible.

sno2: I'd rather just mark the function as `unsafe`. I don't want to add a ton of useless checks in the parser/tokenizer and return `Result<T, E>` when they should have been done by the module loader and runtime.

MierenManz: Why are we limiting it to 4gb? I get that that is a lot of source code. But mega apps that are bundled can go over this limit

sno2: > Why are we limiting it to 4gb? 
I can hardly imagine a use-case for that large of a JavaScript file.
> But mega apps that are bundled can go over this limit
We'd need an example of something in production that actually gets this large before deoptimizing the entire token structure for such annoyingly large files.

Phosra: The checks would merely propagate the error back to the consumer, the checks are primarily going to be the `?` operator.

Alternatively, how about a accepting a public new type
```rust
// name it whatever 
struct U32Slice<'l>(&'l [u8]);

impl U32Slice {
 fn new(slice: U32Slice) -> Result<Self, TooLarge> {
  match slice.len().try_from() {
   Ok(_) => Self(slice),
   Err(_) => TooLarge
  }
 }
}
```
The user would need to construct it and handle the error, before this code receives it.

Phosra: > Why are we limiting it to 4gb? I get that that is a lot of source code. But mega apps that are bundled can go over this limit

File do not get this large..
TSC is under 10 MB, Facebook is about 7 MB from what I've heard.
All of the JS in Deno and Node.js shouldn't exceed 1 GB; how about you measure all of SpiderMonkey's source to get an estimate of how much code you're talking about here?

tokenizer/src/lib.rs
```diff
@@ -201,13 +205,13 @@ impl<'a> TokenStream<'a> {
                         let (offset0, cp0) = chars.next().unwrap();
 
                         if cp0.is_id_start() {
-                            self.index += offset0 as u32 + 1;
+                            self.index = start + offset0 as u32;
```
andreubotella, line 208: Isn't `offset0` always 0, since it's the offset at which the first Unicode character starts in the slice?

tokenizer/src/lib.rs
andreubotella: Why do we even need `len` as a field in `TokenStream`? The `buffer` slice already stores the length.

Phosra: > Why do we even need `len` as a field in `TokenStream`? The `buffer` slice already stores the length.

That's what this argument revolves around. OP wants to use a u32 so that offsets for each token would be a u32. (I'll let them further elaborate on the reasoning.)

I'm in favour of not having the duplicate lengths present, one way or another.

MierenManz: > I'm in favour of not having the duplicate lengths present, one way or another.

I agree. We should use buffer.len() and cast it to a u32 / convert it to a u32

Phosra: What should that struct layout look like?

tokenizer/src/lib.rs
```diff
@@ -201,13 +205,13 @@ impl<'a> TokenStream<'a> {
...
 
                             for (offset, cp) in chars {
+                                self.index = start + offset as u32;
```
andreubotella, line 211: This doesn't work when the last character of a source file belongs to an identifier. Since `offset` returns the offset where _the current character starts_, once we reach EOF `self.index` is off by one character. And in subsequent invocations of `next()`, `chars` in this loop is empty, which means `self.index` will be equal to `start`, causing an endless loop.

tokenizer/src/lib.rs
sno2: > Why do we even need `len` as a field in `TokenStream`? The `buffer` slice already stores the length.

I want it cached for every `next` call.  In the case of calling `len`, it would pull the len from the stack every iteration.

tokenizer/src/lib.rs
sno2: Oh, yeah could not wrap this section in my head it seems.  Thanks

tokenizer/src/lib.rs
andreubotella: I oppose this change. I think every bit of unsafe code needs a justification as to why it is needed, and for performance that would require benchmarks. Since we aren't anywhere close enough to being able to benchmark, I don't think unsafe for performance reasons is justified. Whenever we have some semblance of a complete parser to bytecode, we could then try any unsafe optimizations and see if they're justified.

The same goes for the use of `get_unchecked`.

Phosra: > > Why do we even need `len` as a field in `TokenStream`? The `buffer` slice already stores the length.
> 
> I want it cached for every `next` call.  In the case of calling `len`, it would pull the len into the stack every iteration.

Have you confirmed that this is the case? It makes no sense that `self.len` would be cached, but that `self.buffer` wouldn't?

Phosra: > While this change doesn't add `unsafe` code in itself, it would either cause wrong behavior in corner cases, or require `TokenStream::new` to be an unsafe function, which would itself force any callers of the tokenizer to use unsafe code. Panicking is the correct behavior IMO.

May you elaborate for specifically why you would prefer panicking here, as opposed to my suggestions to use an enum?
https://github.com/trynova/nova/pull/5#discussion_r956453359
https://github.com/trynova/nova/pull/5#discussion_r956458746

I expect that we would need to use these later anyways, so consumer code could gracefully handle the error, however they may choose.

Even so, I would prefer the panic over the silently lossy truncation.

andreubotella: > May you elaborate for specifically why you would prefer panicking here, as opposed to my suggestions to use an enum? [#5 (comment)](https://github.com/trynova/nova/pull/5#discussion_r956453359) [#5 (comment)](https://github.com/trynova/nova/pull/5#discussion_r956458746)

Right, those suggestions probably make more sense than panicking.

tokenizer/src/lib.rs
```diff
@@ -108,7 +112,7 @@ impl<'a> TokenStream<'a> {
         loop {
             let c = if self.index < self.len {
                 // TODO: find way to avoid upcast on 64bit machines
-                self.buffer[self.index as usize]
+                *unsafe { self.buffer.get_unchecked(self.index as usize) }
```
andreubotella, line 115: You could do:
```rust
let c = self.buffer.get(self.index as usize).copied().unwrap_or(0); 
```

PR #6, opened 2022-08-27
fix(tokenizer): Allow multi-line comments that aren't followed by a space

andreubotella:
Currently the start of a multi-line comment is only parsed as such if the tokenizer is in one of the junk states; that is, if it is followed by a space or newline. Starts of multi-line comments that occur at the beginning of the source text or immediately after the end of a token are misparsed as a division punctuator followed by a multiplication punctuator. This change fixes this.


tokenizer/src/lib.rs
```diff
@@ -485,3 +491,53 @@ impl<'a> TokenStream<'a> {
...
+            }
+        }
+    }
+    impl<'a> Iterator for TokenAndSliceIterator<'a> {
+        type Item = (TokenKind, &'a [u8]);
+
+        fn next(&mut self) -> Option<Self::Item> {
+            if self.current_token.is_none() {
+                self.current_token = Some(self.stream.next());
+            }
```
Phosra, line 522: It's a small issue, but if you could match the `current_token` and reassigning, then take the resulting token as a local, the later unwraps would be unnecessary.
```suggestion
            let current_token = match self.current_token {
                Some(token) => token,
                None => {
                    let token = self.stream.next();
                    self.current_token = Some(token);
                    token
            };
```

tokenizer/src/lib.rs
```diff
@@ -485,3 +491,53 @@ impl<'a> TokenStream<'a> {
...
+                stream: TokenStream::new(buffer),
+                slice: buffer,
+                current_token: None,
+            }
+        }
+    }
+    impl<'a> Iterator for TokenAndSliceIterator<'a> {
+        type Item = (TokenKind, &'a [u8]);
+
+        fn next(&mut self) -> Option<Self::Item> {
```
Phosra, line 519: I'm not seeing how `next` returns `None`?

MierenManz (commented):
Lgtm but I'm not the parsing expert here

aapoalas (approved):

PR #8, opened 2022-08-28
chore: Add Github Actions CI

andreubotella:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+      - main
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
```
MierenManz, line 18: Maybe seperate the test step and only have the timeout there? Just a suggestion. This seems fine for now aswell

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+    branches:
+      - main
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
```
MierenManz, line 17: Could you add a todo so that we run ci on all 3 platforms rather than Linux only?

MierenManz (approved):
Other than the suggestion and comment
Lgtm

.github/workflows/ci.yml
andreubotella: I've set the build's timeout to 40 minutes and the test step's to 20. Is that okay?

MierenManz:
Lgtm!

PR #9, opened 2022-08-29
add basic wasm decoder

MierenManz:


wasm/src/decoder/common.rs
```diff
@@ -0,0 +1,94 @@
...
+            0x60 => Ok(Self::Func),
+            0x40 => Ok(Self::Void),
+            _ => Err(()),
+        }
+    }
+}
+
+pub struct FnType {
+    pub params: Vec<ValueKind>,
+    pub result: Vec<ValueKind>,
```
Phosra, line 66: This is plural, how about `results`?

Phosra, line 66: Maybe these should be `Box<[ValueKind]>` to not indicate that the lengths will be mutated?

Phosra (commented):
I also don't think using plain u8 works for deciding, because Wasm is a variable-sized instruction set.

wasm/src/decoder/code_section.rs
```diff
@@ -0,0 +1,32 @@
+use super::common::CodeBlock;
+use super::util;
+use crate::error::Result;
+
+pub fn decode_code_section<R: crate::Reader>(reader: &mut R) -> Result<CodeBlock> {
+    let body_size = util::decode_u32(reader)? as u64;
+    let offset = reader.seek(std::io::SeekFrom::Current(0))?;
```
Phosra, line 7: Nit: I believe there's an alias to 0, as Start, or something of the sort

wasm/src/decoder/code_section.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    let v = util::decode_vec(reader, |x| {
+        Ok((util::decode_u32(x)?, util::decode_kind(x)?))
+    })?;
+
+    let mut locals = Vec::new();
+    for x in v {
+        for _ in 0..x.0 {
+            locals.push(x.1);
+        }
```
Phosra, line 17: Hmm, this could extend via an interior?
`locals.extend((0..x.0).map(|_| x.1)` or maybe `locals.resize(locals.len() + x.0, x.1)` or maybe this is overkill anyways

wasm/src/decoder/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+        0x01 => {
+            let vec = util::decode_vec(reader, type_section::decode_type_section)?;
+            Section::Type(vec)
+        }
+        // 0x02 => {
+        //     let vec = util::decode_vec(reader, import_section::decode_import_section)?;
+        //     Section::Import(vec)
+        // }
+        0x03 => {
+            let vec = util::decode_vec(reader, |r| Ok(read::unsigned(r)? as u32))?;
```
Phosra, line 80: What does `unsigned` yield here? If it reads larger than a u32, too many bytes may be read, and the upper bytes are truncated..

wasm/src/decoder/type_section.rs
```diff
@@ -0,0 +1,19 @@
+use super::common;
+use super::util;
+use crate::error::Error;
+use crate::error::Result;
+
+const FN_SIGNATURE: u8 = common::ValueKind::Func as u8;
+
+pub fn decode_type_section<R: std::io::Read>(reader: &mut R) -> Result<common::FnType> {
```
Phosra, line 8: Does this allow modules that lack functions?
(I don't know how this part is being used)

wasm/src/error.rs
```diff
@@ -0,0 +1,52 @@
+use leb128::read;
+
+pub type Result<T> = std::result::Result<T, crate::error::Error>;
+
+#[repr(u8)]
+#[derive(Debug)]
+pub enum Error {
+    IoError(std::io::Error),
+    ReadZeroBytes,
+    TooManyBytes(u8),
```
Phosra, line 10: On its own, I don't think it's obvious as to what the u8 represents here?
As in, it's not clear as to whether it was the expected amount, or how many were read, even though I'd expect the latter to be usize.

wasm/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+mod decoder;
+pub mod error;
+
+pub trait Reader: std::io::Read + std::io::Seek {}
+
+pub fn compile_module<R: Reader>(bytes: &mut R) -> Result<(), error::Error> {
```
Phosra, line 6: Maybe this should use `todo!()`, as it doesn't seem complete?

Phosra (commented):
I'd suggest more `todo!()` instead of commented out code, really

wasm/src/common.rs
```diff
@@ -0,0 +1,94 @@
...
+            0x02 => Ok(Self::Memory),
+            0x03 => Ok(Self::Global),
+            _ => Err(()),
+        }
+    }
+}
+
+#[repr(u8)]
+#[derive(Copy, Clone)]
+pub enum ValueKind {
```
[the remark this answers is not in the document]
Phosra: I later saw the `FnType`, so I think this is a non-issue

wasm/src/decoder/type_section.rs
MierenManz: yes. This only get's called when modules do have a type_section and thus a function saction

wasm/src/error.rs
MierenManz: Should I add a comment? something like `// TooManyBytes(expected)`

wasm/src/decoder/mod.rs
MierenManz: Gonna change this to `decode_u32` and not `read::unsigned`

wasm/src/decoder/code_section.rs
MierenManz: Was trying that. But I wasn't able to figure it out. Gonna change that

wasm/src/decoder/code_section.rs
MierenManz: `SeekFrom::Start` refers to the first byte and not the current one. And I do need the current one

MierenManz:
>I also don't think using plain u8 works for deciding, because Wasm is a variable-sized instruction set.

So change the instructions from `Vec<u8>` to `Vec<Instruction>` where `Instruction` is a enum of all possible instructions?

wasm/src/decoder/code_section.rs
Phosra: I'll do it in a follow-up PR

wasm/src/error.rs
Phosra: Could be `TooManyBytes { expected_bytes: u8 }`

MierenManz: Sgtm

wasm/src/decoder/code_section.rs
MierenManz: You should be able to just push to this pr branch

wasm/src/decoder/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+    leb128::read::unsigned(reader)?;
+    let section = match section_id_byte[0] {
+        0x01 => {
+            let vec = util::decode_vec(reader, type_section::decode_type_section)?;
+            Section::Type(vec)
+        }
+        // 0x02 => {
+        //     let vec = util::decode_vec(reader, import_section::decode_import_section)?;
+        //     Section::Import(vec)
+        // }
```
Phosra, line 78: This is an example where I'd suggest `todo!(...)/unimplemented!(...)`

wasm/src/decoder/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+        0x03 => {
+            let vec = util::decode_vec(reader, |r| Ok(read::unsigned(r)? as u32))?;
+            Section::Function(vec)
+        }
+        // 0x03 => Section::Function(section_data),
+        // 0x04 => Section::Table(section_data),
+        // 0x05 => Section::Memory(section_data),
+        // 0x06 => Section::Global(section_data),
+        // 0x07 => Section::Export(section_data),
+        // 0x09 => Section::Element(section_data),
```
Phosra, line 88: Here too

wasm/src/decoder/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+        0x0A => {
+            let vec = util::decode_vec(reader, code_section::decode_code_section)?;
+            Section::Code(vec)
+        }
+        // 0x0B => Section::Data(section_data),
```
Phosra, line 93: Ditto

Phosra (commented):
> > I'd suggest more todo!() instead of commented out code, really
> 
> There is a `unimplemented!()` somewhere

Phosra:
> > I also don't think using plain u8 works for deciding, because Wasm is a variable-sized instruction set.
> 
> So change the instructions from `Vec<u8>` to `Vec<Instruction>` where `Instruction` is a enum of all possible instructions?

Your call, but how are you planning on doing it otherwise with just `u8`?

MierenManz:
> Your call, but how are you planning on doing it otherwise with just u8?

Essentially have a transformer from a Vec<u8> directly to cranelift ir rather than to Instruction and from Instruction to ir

Phosra:
> > Your call, but how are you planning on doing it otherwise with just u8?
> 
> Essentially have a transformer from a Vec<u8> directly to cranelift ir rather than to Instruction and from Instruction to ir

I'm specifically talking about instructions that exceed the size of u8? And we might want it to be more general, because we may decide to eventually drop CraneLift

wasm/src/decoder/code_section.rs
Phosra: I'll save it for an optimisation PR when we have tests and benchmarking.

MierenManz:
Those would have additional bytes being read. The transformer wouldn't take only a u8 and convert it to a Instruction but rather take a `std::io::Read` or slice and consume the bytes necessary there and then convert that

wasm/src/decoder/code_section.rs
Phosra: Yup, I forgot the API

wasm/src/decoder/common.rs
MierenManz: your call

MierenManz: sgtm

MierenManz:
After this has landed I'm gonna remove the leb128 dep and just write a quick macro for decoding varint etc. Because right now it may return an error `TooManyBytes { expected: 10, found: 11 }` eventho you decode a u32

wasm/src/common.rs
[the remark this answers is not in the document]
MierenManz: Will add those later

wasm/src/common.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+impl From<ExternalKind> for u8 {
+    fn from(v: ExternalKind) -> Self {
+        v as Self
+    }
+}
+
+impl TryFrom<u8> for ExternalKind {
+    type Error = ();
```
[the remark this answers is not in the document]
MierenManz: would a new `Error` variant word. `UnknownExternalKind` and for `ValueKind` the `UnknownValueKind`?

wasm/src/common.rs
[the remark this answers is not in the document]
Phosra: We could have their decoding early, and just their usage as  `unimplemented!()`.

wasm/src/decoder/common.rs
Phosra: In .wat, it's singular "param" though too `(param i32 i32 i32)`, and arguably we should just spell it out for "parameters." I'd say they should both be plural.

wasm/src/decoder/util.rs
```diff
@@ -0,0 +1,50 @@
...
+    let length = read::unsigned(reader)? as usize;
+    if length > u32::MAX as usize {
+        return Err(Error::ArrayTooLarge);
+    }
+
+    let mut v = Vec::with_capacity(length);
+
+    for x in v.iter_mut().take(length) {
+        *x = func(reader)?;
+    }
```
Phosra, line 20: Wouldn't this iterator be empty?

wasm/src/decoder/util.rs
```diff
@@ -0,0 +1,50 @@
...
+        Ok(v) => Ok(v),
+        Err(_) => Err(Error::InvalidValueKind),
+    }
+}
+
+pub fn decode_u32<R: crate::Reader>(reader: &mut R) -> Result<u32> {
+    let s = reader.seek(std::io::SeekFrom::Current(0))?;
+    let length = read::unsigned(reader)?;
+    // This is so wacky to do. Replacing it with a better system should be done at some point
+    let end = reader.seek(std::io::SeekFrom::Current(0))?;
```
Phosra, line 39: I wouldn't expect this to even move the cursor, wouldn't you want to save the last position?

wasm/src/error.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl std::fmt::Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::IoError(e) => e.fmt(f)?,
+            Self::ReadZeroBytes => f.write_str("Tried to read from reader but got 0 bytes")?,
+            Self::TooManyBytes { expected, found } => f.write_str(&format!(
+                "Varint is too long! Expected {expected} bytes but found {found}"
+            ))?,
```
Phosra, line 26: Could avoid the allocation
```suggestion
            Self::TooManyBytes { expected, found } => write!(f, "Varint is too long! Expected {expected} bytes but found {found}")?,
```

wasm/src/error.rs
```diff
@@ -0,0 +1,58 @@
...
+    UnknownSectionID,
+    InvalidValueKind,
+    InvalidExternalKind,
+    InvalidSignatureType,
+    ArrayTooLarge,
+}
+
+impl std::fmt::Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
```
Phosra, line 21: every path ends in `?`, could that be moved to the end of the `match`?

wasm/src/decoder/util.rs
MierenManz: clippy didn't like my previous solution and recommended this to me instead.

previous solution was below
```rs
let mut v = Vec::with_capacity(length);

for i in 0..length {
    v[i] = func(reader)?;
}
```

wasm/src/decoder/util.rs
MierenManz: well we would need to determine how many bytes were read. and by subtracting the begin position from the end we should get the difference. Eg how many bytes were read than expected

wasm/src/decoder/util.rs
Phosra: It's because clippy lacks the knowledge that the Vec's *length is zero*.

This is just a bug, you want to push upon each iteration, not write into it via `*x =` or even `v[i] = `.

Phosra:
```rust
for _ in 0..length {
 v.push(func(reader)?);
}
```

MierenManz: oh yeah I see. Stupid mistake from myself. Weirdly enough I did do it correctly in the code section

```rs
for _ in 0..x.0 {
            locals.push(x.1);
        }
        ```

Phosra (approved):
I'm still doubtful about the seeking code, but I'll double-check that with some others later.

MierenManz:
https://doc.rust-lang.org/std/io/trait.Seek.html#method.stream_position Check out this method. It's an alias for the seek

Phosra:
Ah, I hadn't checked the return value of seek, that makes sense. LGTM

PR #10, opened 2022-08-30
Remove leb128 dep

MierenManz:
Removes leb128 dep in favor of our own 2 decode functions.
This allows us not to use a wacky `Seek::stream_pos` or `Seek::seek(SeekFrom::Current(0))`
But rather just return the consumed byte count directly

### DIFF
--- a/wasm/src/varint.rs
+++ b/wasm/src/varint.rs
@@ -1,0 +1,94 @@
+use crate::error::Error;
+
+const SEGMENT_BITS: u8 = 0x7F;
+const CONTINUE_BIT: u8 = 0x80;
+
+pub fn decode_u32<R: std::io::Read>(reader: &mut R) -> Result<(u32, u8), Error> {
+    let mut length = 0;
+    let mut value = 0;
+
+    loop {
+        let mut bytes: [u8; 1] = [0 ;1];
+        reader.read_exact(&mut bytes)?;
+        value |= ((bytes[0] & SEGMENT_BITS) as u32) << length;
+
+        length += 7;
+        if bytes[0] & CONTINUE_BIT == 0 {
+            break;
+        }
+
+        if length > 32 {
+            print!("{length}");
+            return Err(Error::TooManyBytes { expected: 5, found: length / 7 + 1})
+        }
+    }
+
+    length += length / 7;
+    Ok((value, length / 8))
+}
+
+pub fn decode_u64<R: std::io::Read>(reader: &mut R) -> Result<(u64, u8), Error> {
+    let mut length = 0;
+    let mut value = 0;
+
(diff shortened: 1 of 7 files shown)


MierenManz:
Also added 2 tests for it!

Phosra:
Reformat?

MierenManz:
yeah fmt was set wrong. It's now set to cargo

wasm/src/varint.rs
```diff
@@ -0,0 +1,99 @@
+use crate::error::Error;
+
+const SEGMENT_BITS: u8 = 0x7F;
+const CONTINUE_BIT: u8 = 0x80;
+
+pub fn decode_u32<R: std::io::Read>(reader: &mut R) -> Result<(u32, u8), Error> {
```
Phosra, line 6: Nit: It might be easier to read and understand this with a struct like
```rust
struct DecodedU32Result {
 value: u32,
 bytes_read: u8
}
```
Although... technically there's a mathematical relationship between those two values, so you could logarithmically check the u32 result to reconstruct the byte count consumed?

Or, is this byte sequence valid: `[0x80, 0x80, 0x80, 0x80, 0x00]`?

wasm/src/varint.rs
```diff
@@ -0,0 +1,99 @@
...
+        reader.read_exact(&mut bytes)?;
+        value |= ((bytes[0] & SEGMENT_BITS) as u32) << length;
+
+        length += 7;
+        if bytes[0] & CONTINUE_BIT == 0 {
+            break;
+        }
+
+        if length > 32 {
+            return Err(Error::TooManyBytes {
```
Phosra, line 21: This should be a debug print?

wasm/src/varint.rs
MierenManz: That is a valid technically. but basically never used because it results in `0`

Phosra (approved):

wasm/src/varint.rs
```diff
@@ -0,0 +1,148 @@
...
+        let mut bytes: [u8; 1] = [0; 1];
+        reader.read_exact(&mut bytes)?;
+        value |= ((bytes[0] & SEGMENT_BITS) as u64) << length;
+
+        length += 7;
+        if bytes[0] & CONTINUE_BIT == 0 {
+            break;
+        }
+
+        if length > 64 {
```
Phosra, line 56: Wait, would this overflow for `length == 64`?

MierenManz: Oh shoot you are right...
Should've been `>=` not `>`

MierenManz: Opened a pr for that

PR #13, opened 2022-08-31
Wasm instructions

MierenManz:


MierenManz:
I've thought of a different design that will likely work better than these instructions.
Closing for now. Leaving the branch open

PR #20, opened 2023-03-03
feat(ast): reserved words

load1n9:


MierenManz (changes requested):

ast/src/lib.rs
```diff
@@ -18,28 +18,47 @@ pub enum Keyword {
...
+    Enum,
+    Implements,
+    Interface,
+    Package,
+    Private,
+    Protected,
+    Public,
+
+    // deprecated reserved words
+    With,
```
Phosra, line 63: This isn't deprecated – it's just denied in strict mode, and I expect it would be a keyword. Please confirm.

ast/src/lib.rs
```diff
@@ -18,28 +18,47 @@ pub enum Keyword {
...
+}
```
Phosra, line 64: Add a blank line at the end for the formatting CI
```suggestion
}

```

Phosra (changes requested):

ast/src/lib.rs
Phosra: @load1n9 Did you confirm with the specification?

MierenManz (approved):
Lgtm

ast/src/lib.rs
load1n9: We talked about how we really didn't want it but for now we could keep it

Phosra: May I ask where this talk was? Why wouldn't we want it?

MierenManz: @Phosra https://discord.com/channels/1012084654389604372/1012084654389604376/1081302847930499202
First message of the convo

Phosra: Right. I disagreed at https://discord.com/channels/1012084654389604372/1012084654389604376/1081334881642094612.
No issue here, it's to be kept until we decide on something different later.

sno2:
Closing because this was for using the old parser/tokenizer setup.

PR #21, opened 2023-06-05
Basic Bytecode Impl with oxc's parser

sno2:


nova_vm/src/lib.rs
```diff
@@ -0,0 +1,576 @@
...
+    pub map: std::collections::HashMap<&'a str, u32>,
+    pub parent: Option<Box<Env<'a>>>,
+}
+
+impl<'a> VM<'a> {
+    pub fn interpret(&self) -> JsResult<()> {
+        let mut memory = Vec::<Value>::with_capacity(self.pc as usize);
+
+        for _ in 0..self.pc {
+            memory.push(Value::Undefined);
```
andreubotella, line 249: Why is this needed, since each instruction can specify the memory address it operates on?

andreubotella (commented):
I might not be too familiar with bytecode, but it seems odd to me that every instruction operates on an arbitrary address passed in the bytecode itself. Usually bytecode or machine code architectures are either stack-based or register-based, but you usually don't have instructions directly operating on memory addresses. This looks like an odd design to me, and it seems like it'd be much cleaner if it was a stack machine.

andreubotella:
It also seems like the bytecode building and the interpreting could be split apart, rather than having to share the `VM` struct. For example, the parsing could generate the `VM` that then gets used to evaluate the code.

Also, what is `VM` exactly? Is it an agent in the JS spec (= a V8 isolate)? Is it a realm? A parsed script or module?

sno2:
> I might not be too familiar with bytecode, but it seems odd to me that every instruction operates on an arbitrary address passed in the bytecode itself. Usually bytecode or machine code architectures are either stack-based or register-based, but you usually don't have instructions directly operating on memory addresses. This looks like an odd design to me, and it seems like it'd be much cleaner if it was a stack machine.

Yeah...I haven't built a bytecode interpreter before this so that makes more sense.

Do you mean something like this? (Still not really sure how these things are designed)

```
PushNumber 23
PushNumber 40
Add
```

And Add would pop `40` and `23` off a stack and then push `63`?

andreubotella:
> Do you mean something like this? (Still not really sure how these things are designed)
> 
> ```
> PushNumber 23
> PushNumber 40
> Add
> ```
> 
> And Add would pop `40` and `23` off a stack and then push `63`?

That's right.

Building such a bytecode might be tricker, though, since the values on top of the stack might not be the ones you need, and building the most optimal bytecode might mean moving instructions around. But you could have instructions to copy a value from elsewhere in the stack to the top, and a noop instruction to discard values from the top.

MierenManz:
Maybe rather than using noop for discarding a value a noop and drop might be better? (same as for wasm)

MierenManz:
> Yeah...I haven't built a bytecode interpreter before this so that makes more sense.
I've built semi usable bytecode interpreters before. So I could maybe help you with that

sno2:
Stale, might be able to use later.

PR #24, opened 2023-07-24
feat: engine setup

sno2:
This is greatly inspired by Kiesel and SerenetyOS's LibJS JavaScript engines.


nova_vm/src/small_string.rs
```diff
@@ -0,0 +1,43 @@
...
+        let mut data: [u8; 7] = [0, 0, 0, 0, 0, 0, 0];
+        data.copy_from_slice(value.as_bytes());
+        Self { data }
+    }
+}
+
+impl TryFrom<&str> for SmallString {
+    type Error = ();
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        if value.len() < 8 {
```
aapoalas, line 37: **issue:** The string cannot end in a null character. We cannot tell apart `"foo\0"` and `"foo"` since we think the string ends at the last non-intermediate null byte.

sno2: Thank you

nova_vm/src/small_integer.rs
```diff
@@ -0,0 +1,52 @@
...
+            Err(())
+        }
+    }
+}
+
+impl Into<i64> for SmallInteger {
+    fn into(self) -> i64 {
+        let Self { data } = self;
+        let mut bytes: [u8; 8] = [0, 0, 0, 0, 0, 0, 0, 0];
+        bytes.copy_from_slice(&data);
```
aapoalas, line 45: **issue:** This won't work for negative numbers: The top byte needs to have sign extension which this will not.

Also, `copy_from_slice` will assert that the lengths of the slices are the same so I think this should just assert every time (7 vs 8).

nova_vm/src/small_string.rs
```diff
@@ -0,0 +1,43 @@
...
+impl TryFrom<&str> for SmallString {
+    type Error = ();
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        if value.len() < 8 && !value.as_bytes().contains(&0) {
+            Ok(Self::from_str_unchecked(value))
+        } else {
+            Err(())
+        }
+    }
+}
```
aapoalas, line 43: The tests should perhaps be kept?

nova_vm/src/types.rs
```diff
@@ -0,0 +1,17 @@
...
+pub use spec::{Base, Reference, ReferencedName};
+
+impl From<Object> for Value {
+    fn from(value: Object) -> Self {
+        todo!()
+    }
+}
+
+#[derive(Debug)]
+pub struct Object;
```
aapoalas, line 14: **question:** I presume these are intended to become (checked) wrappers around `Value`?

nova_vm/src/types/language/value.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    /// 6.1.5 The Symbol Type
+    /// https://tc39.es/ecma262/#sec-ecmascript-language-types-symbol-type
+    Symbol(Handle<SymbolHeapData>),
+
+    /// 6.1.6.1 The Number Type
+    /// https://tc39.es/ecma262/#sec-ecmascript-language-types-number-type
+    Number(Handle<NumberHeapData>),
+    IntegerNumber(SmallInteger), // 56-bit signed integer.
+    FloatNumber(f32),
```
aapoalas, line 38: **praise:** I thought about adding this as well but ended up not doing so due to thinking it wouldn't really work for the most part. Hiding in NaN and Infinities inside it is a good idea! Is a signed zero also representable in f32?

aapoalas: Seems like it is so that can also be hidden in here! Nice!

aapoalas: Oh yeah, I was also thinking that maybe we could implement our own `f56` :D But that may be quite hard to do and wouldn't necessarily bring much benefit.

nova_vm/src/small_integer.rs
sno2: Yeah my implementation is broken. I had worked on it before you pushed and (wrongly) didn't look into it. Will copy your code.

nova_vm/src/small_string.rs
sno2: Yep, porting them now.

nova_vm/src/small_string.rs
sno2: Yeah my implementation is broken. I had worked on it before you pushed and (wrongly) didn't look into it. Will copy your code.

aapoalas (changes requested):
I'll mark as Request changes since I'm not particularly thrilled about 90% of my heap code being simply removed :) I'll try leave more comments tomorrow.

nova_vm/src/execution/environments/declarative_environment.rs
```diff
@@ -0,0 +1,29 @@
+use super::Environment;
+use crate::types::Value;
+use std::collections::HashMap;
+
+/// 9.1.1.1 Declarative Environment Records
+/// https://tc39.es/ecma262/#sec-declarative-environment-records
+#[derive(Debug)]
+pub struct DeclarativeEnvironment {
+    pub outer_env: Option<Environment>,
+    pub bindings: HashMap<&'static str, Binding>,
```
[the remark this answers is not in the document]
sno2: Nice suggestion. I had added the `&'static str` to avoid lifetimes while copying the spec entities but this would be better. Will work on implementing this.

nova_vm/src/types/language/value.rs
Phosra: > Oh yeah, I was also thinking that maybe we could implement our own `f56` :D But that may be quite hard to do and wouldn't necessarily bring much benefit.

We can :)
Actually it may be rather trivial byte manipulation and modification. As for it not bringing much benefit, yes, it may be rather slow to resize f64 <-> f56 frequently (it's not as straightforward as bit-wise truncation).

Honestly once we get anywhere with this, I would love to experiment with this for the purpose of comparing against loading the heap floats instead, ... with some real scripts against this.

nova_vm/src/small_integer.rs
```diff
@@ -11,18 +11,23 @@ impl std::fmt::Debug for SmallInteger {
 }
 
 impl SmallInteger {
-    pub const MIN: i64 = -(2 as i64).pow(56) / 2 + 1;
-    pub const MAX: i64 = (2 as i64).pow(56) / 2 - 1;
+    pub const MIN: i64 = -(2 as i64).pow(53) / 2 + 1;
```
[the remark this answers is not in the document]
Phosra: I56 and I53 should be separate enum tags.

They should both exist, so we can easily write appropriate code to, for example, check overflows for their own respective bit widths.

nova_vm/src/execution/environments/declarative_environment.rs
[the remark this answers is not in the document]
Phosra: The `ParallelVec`s in question should be sorted, and all code should appropriately use logarithmic lookups (stdlib has us covered here with slice methods).

If we're to do away with `HashMap`s entirely for some parts of the code\*, we should have a generic type built on `ParallelVec` for reuse.

\* There are other hash map types and algorithms and allocator configurations? Maybe those would be more suitible? A `HashMap` in an arena allocator for example. I believe there are already libraries for this.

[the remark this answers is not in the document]
Phosra: Or BTrees? https://doc.rust-lang.org/std/collections/struct.BTreeMap.html

nova_vm/src/types/language/value.rs
```diff
@@ -125,6 +125,14 @@ impl Value {
         matches!(self, Value::Symbol(_))
     }
 
+    pub fn is_empty_string(self) -> bool {
+        if let Value::SmallString(s) = self {
+            s.len() == 0
```
aapoalas, line 131: **nitpick:** It'd be faster to match against a const-time created empty string since then it becomes just a single equality comparison.

sno2: I do not understand. Can you please explain what that code would look like?

aapoalas: Something like 
```rs
match self {
  Value::SmallString([0,0,0,0,0,0,0]) => true,
  _ => false,
}
```

Phosra: This isn't possible. You do need to handle every possible type of string variant / structure.

Right now it's only implemented for `SmallString`.

Right now it could be
```rust
pub fn is_empty_string(self) -> bool {
 self == Value::SmallString([0; 7])
}
```with a bit-wise comparison, but we don't even have a comparison trait implementation yet.

But more importantly, this doesn't handle strings in general.

Comment (aapoalas) [reply nova_vm/src/types/language/value.rs:131]: A heap string is never empty, that's sort of the whole point.

File: nova_vm/src/execution/realm/intrinsics.rs [hunk 9]
```diff
@@ -0,0 +1,237 @@
...
+//       contain a mutable reference to [`Realm`] and be created via a
+//       `Realm::intrinsic()` method to guarantee safety.
+
+pub struct Intrinsics;
+
+impl Intrinsics {
+    /// %Array%
+    pub fn array() -> Object {
+        Object::new(Value::Object(Handle::new(
+            BuiltinObjectIndexes::ArrayConstructorIndex as u32,
```
aapoalas, line 16: This is incorrect: The constructor should be a function and the arrayconstructorindex is the index of the function's heap object data which the function heap data refers to. I had the `get_constructor_index` function somewhere on the heap constants file that translates the built-in object heap data index into the corresponding function index.

aapoalas:
I rebased and merged this in #28

PR #43, opened 2023-10-30
feat(ecmascript): more builtins

sno2:


nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -119,3 +152,128 @@ pub(crate) fn same_value_non_number<T: Copy + Into<Value>>(_agent: &mut Agent, x
...
+/// expressions. If LeftFirst is true, the x parameter corresponds to an
+/// expression that occurs to the left of the y parameter's corresponding
+/// expression. If LeftFirst is false, the reverse is the case and operations
+/// must be performed upon y before x. It performs the following steps when
+/// called:
+pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
+    agent: &mut Agent,
+    x: impl Into<Value> + Copy,
+    y: impl Into<Value> + Copy,
+) -> JsResult<Value> {
```
aapoalas, line 175: **nitpick:** I think it might be worthwhile to leave out the technical `undefined` property of this and return a `JsResult<Option<bool>>` instead. Converting to a `Value` from there is fairly trivial from there after all.

nova_vm/src/ecmascript/types/language/bigint.rs
```diff
@@ -11,3 +15,23 @@ pub enum BigInt {
...
     SmallBigInt(SmallInteger) = SMALL_BIGINT_DISCRIMINANT,
 }
+
+impl BigInt {
+    /// ### [6.1.6.2.12 BigInt::lessThan ( x, y )](https://tc39.es/ecma262/#sec-numeric-types-bigint-lessThan)
+    ///
+    /// The abstract operation BigInt::lessThan takes arguments x (a BigInt) and
+    /// y (a BigInt) and returns a Boolean. It performs the following steps when
+    /// called:
+    pub(crate) fn less_than(self, agent: &mut Agent, y: BigInt) -> bool {
```
aapoalas, line 25: **suggestion:** I haven't really been following this pattern myself as of yet, but I've been thinking that it might be a good idea to actually move all abstract operations into their own modules, so eg. `types/language/bigint/abstract_operations.rs` in this case.

That way we have kind of a closer control over the API of our `Value` "subclasses" and it replicates the spec to a T.

sno2: Interesting, what do you propose in the case of naming collisions? e.g. `Number::lessThan` vs abstract `LessThan`?

aapoalas: Module scoping, so `number::abstract_operations::less_than` (or perhaps rather `number::less_than`) vs general `ecmascript::abstract_operations::less_than`. If both are needed in the same file then they'd need to be scoped, but I'm not sure if that's particularly common.

aapoalas: But it's also fair to say that `Number::lessThan` in the spec could just as well be a method on `Number` enum.

nova_vm/src/ecmascript/types/language/value.rs
```diff
@@ -290,9 +290,12 @@ impl From<bool> for Value {
...
 
-impl From<Option<Value>> for Value {
-    fn from(value: Option<Value>) -> Self {
-        value.unwrap_or(Value::Undefined)
+impl<T> From<Option<T>> for Value
+where
+    T: Into<Value>,
+{
+    fn from(value: Option<T>) -> Self {
+        value.map_or(Value::Undefined, |v| v.into())
```
aapoalas, line 298: **thought:** I wonder if this is a good idea. There might be some places where we want to "default" into `Null` instead of `Undefined`.

That being said, we can just make a helper for it at that time.

PR #44, opened 2023-10-30
feat(ecmascript): Implement ArrayBuffer Object sub-type

aapoalas:
Implements `ArrayBuffer` properly as a sub-type of `Object` (and `Value`). This PR should serve as a good guideline on how `Object` sub-types are implemented at their most basic, though `ArrayBuffer` is probably the only one that is quite as simple as this.

### DIFF
--- a/nova_vm/src/ecmascript/builtins/array_buffer.rs
+++ b/nova_vm/src/ecmascript/builtins/array_buffer.rs
@@ -1,23 +1,53 @@
 //! ## [25.1 ArrayBuffer Objects](https://tc39.es/ecma262/#sec-arraybuffer-objects)
 
 mod data;
-
+use self::data::InternalBuffer;
+use super::{ordinary::ordinary_set_prototype_of_check_loop, Array};
 use crate::{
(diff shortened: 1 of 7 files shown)


nova_vm/src/ecmascript/builtins/array_buffer.rs
```diff
@@ -1,24 +1,54 @@
...
-use self::data::InternalBuffer;
+use std::ops::Deref;
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct ArrayBuffer(ArrayBufferIndex);
 
+impl From<ArrayBufferIndex> for ArrayBuffer {
+    fn from(value: ArrayBufferIndex) -> Self {
+        ArrayBuffer(value)
+    }
```
sno2, line 29: Yeah, defining all the `From<_>` for something is definitely better at the target level. It beats having a ton of `From<_> for Value` in the `value.rs` file.

sno2 (approved):
LGTM, this all plays together really nice :)

PR #46, opened 2023-10-30
chore(ecmascript): Run clippy

aapoalas:


sno2 (approved):
LGTM, we should probably figure out how to get this on CI. Might need to throw an ignore for unused declarations.

aapoalas:
> LGTM, we should probably figure out how to get this on CI. Might need to throw an ignore for unused declarations.

Agreed. Some day :)

PR #49, opened 2023-10-30
fix(ecmascript) fix Number::is_odd_integer

vimirage:
Previously had inverted results. Also optimised the floating-point implementations.


sno2 (approved):
LGTM thanks for fixing my code :)

aapoalas (approved):
Lgtm

PR #50, opened 2023-10-30
feat(vm): more abstract operations

lino-levan:


nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -225,7 +225,7 @@ pub(crate) fn to_number(agent: &mut Agent, argument: Value) -> JsResult<Number>
 
 /// ### [7.1.5 ToIntegerOrInfinity ( argument )](https://tc39.es/ecma262/#sec-tointegerorinfinity)
 // TODO: Should we add another [`Value`] newtype for IntegerOrInfinity?
-pub(crate) fn to_integer_or_infinty(agent: &mut Agent, argument: Value) -> JsResult<Number> {
+pub(crate) fn to_integer_or_infinity(agent: &mut Agent, argument: Value) -> JsResult<Number> {
```
lino-levan, line 228: drive by cleanup

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+    // 3. Return 𝔽(min(len, 2**53 - 1)).
+    Ok(match len {
+        Number::Integer(n) => Number::Integer(n).into(),
+        Number::Float(n) => n.into(),
+        Number::Number(n) => Number::Number(
+            agent
+                .heap
+                .alloc_number(agent.heap.get(n).min(2.0f64.powi(53) - 1.0)),
+        )
+        .into(),
```
lino-levan, line 518: Is this how I should do it? Is this safe?

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+    Ok(Value::Undefined)
+}
+
+/// ### [7.1.22 ToIndex ( value )](https://tc39.es/ecma262/#sec-toindex)
+pub(crate) fn to_index(agent: &mut Agent, argument: Value) -> JsResult<Value> {
+    // 1. Let integer be ? ToIntegerOrInfinity(value).
+    let integer = to_integer_or_infinity(agent, argument)?;
+
+    // 2. If integer is not in the inclusive interval from 0 to 2**53 - 1, throw a RangeError exception.
+    todo!();
```
lino-levan, line 553: I'm not sure how to implement this one.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+
+/// ### [7.1.21 CanonicalNumericIndexString ( argument )](https://tc39.es/ecma262/#sec-canonicalnumericindexstring)
+pub(crate) fn canonical_numeric_index_string(
+    agent: &mut Agent,
+    argument: String,
+) -> JsResult<Value> {
+    // 1. If argument is "-0", return -0𝔽.
+    match argument.as_str(agent) {
+        Some("-0") => {
+            return Ok(Number::Number(agent.heap.alloc_number(-0.0)).into());
```
lino-levan, line 530: Investigating removing this one

lino-levan: Fixed.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
         _ => Ok(Object::try_from(argument).unwrap()),
     }
 }
+
+/// ### [7.1.19 ToPropertyKey ( argument )](https://tc39.es/ecma262/#sec-topropertykey)
+pub(crate) fn to_property_key(agent: &mut Agent, argument: Value) -> JsResult<Value> {
```
aapoalas, line 476: **issue:** While it's not spec-wise exactly the same, I think we might want to replace steps 2 and 3 with mostly a conversion into `PropertyKey`. There are a few special cases (Boolean, BigInts larger than 2^53 - 1, and non-integer Numbers) that should take the normal step 3 route but aside from those since this is an internal operation instead of a public API we should aim to return the type that makes most sense to the engine. And in this case I do feel that `PropertyKey` makes the most sense :)

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+}
+
+/// ### [7.1.20 ToLength ( argument )](https://tc39.es/ecma262/#sec-tolength)
+pub(crate) fn to_length(agent: &mut Agent, argument: Value) -> JsResult<Value> {
+    // 1. Let len be ? ToIntegerOrInfinity(argument).
+    let len = to_integer_or_infinity(agent, argument)?;
+
+    // 2. If len ≤ 0, return +0𝔽.
+    match len {
+        Number::Integer(n) if n.into_i64() <= 0 => {
```
aapoalas, line 497: **question:** Huh, I've never seen this sort of format but apparently it works? Nice.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+            return Ok(0.0.into());
```
aapoalas, line 498: **nitpick:** I presume you can combine all the three matches into a single branch with `|`. Or maybe it doesn't work with this syntax?

We could maybe setup a `Default` impl for `Number` that would just gen a zero. Since it's a valid i53 it'll become an `Integer` variant though, so the `0.0` here may be a bit of a red herring to someone reading the code.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+        Number::Number(n) if *agent.heap.get(n) < 0.0 => {
+            return Ok(0.0.into());
+        }
+        _ => {}
+    }
+
+    // 3. Return 𝔽(min(len, 2**53 - 1)).
+    Ok(match len {
+        Number::Integer(n) => Number::Integer(n).into(),
+        Number::Float(n) => n.into(),
```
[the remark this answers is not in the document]
aapoalas: Yup, definitely seems like the min check is needed.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+        Number::Float(n) if n <= 0.0 => {
+            return Ok(0.0.into());
+        }
+        Number::Number(n) if *agent.heap.get(n) < 0.0 => {
+            return Ok(0.0.into());
+        }
+        _ => {}
+    }
+
+    // 3. Return 𝔽(min(len, 2**53 - 1)).
```
aapoalas, line 509: **issue:** The min-comparison seems to be entirely missing.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+    Ok(match len {
+        Number::Integer(n) => Number::Integer(n).into(),
+        Number::Float(n) => n.into(),
+        Number::Number(n) => Number::Number(
+            agent
+                .heap
+                .alloc_number(agent.heap.get(n).min(2.0f64.powi(53) - 1.0)),
```
aapoalas, line 516: **issue:** These APIs actually should probably be marked unsafe, yeah: We're only allowed to allocate into the heap numbers that cannot be represented as `Number::Integer` or `Number::Float`, ie. the three enum variants must have empty intersections.

Here you're allocating into the heap a number that _is_ a valid `Number::Integer` which is not allowed. For `2**53 - 1` specifically you could do `(2i64.pow(53) - 1).into()`.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+    })
+}
+
+/// ### [7.1.21 CanonicalNumericIndexString ( argument )](https://tc39.es/ecma262/#sec-canonicalnumericindexstring)
+pub(crate) fn canonical_numeric_index_string(
+    agent: &mut Agent,
+    argument: String,
+) -> JsResult<Value> {
+    // 1. If argument is "-0", return -0𝔽.
+    match argument.as_str(agent) {
```
aapoalas, line 528: **nitpick:** Here it's probably more efficient to do a `if argument == String::from_small_string("-0")`

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+        Some("-0") => {
+            return Ok(Number::Float(-0.0).into());
```
aapoalas, line 530: **nitpick:** I'm not sure why this is actually being allowed, I guess it's just how Rust works maybe, but we should avoid creating the enum variants directly. You can use `-0.0f32.into()` here.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+    // 2. Let n be ! ToNumber(argument).
+    let n = to_number(agent, argument.into()).unwrap();
+
+    // 3. If ! ToString(n) is argument, return n.
+    if to_string(agent, n.into()).unwrap() == argument {
+        return Ok(n.into());
+    }
+
+    // 4. Return undefined.
+    Ok(Value::Undefined)
```
aapoalas, line 544: **issue:** Here as well, similarly to with `ToPropertyKey`, we should opt for the most useful type in our repertoire which in this case would be `None`. It also seems to me that this operation never returns `Err` so the return type could just be `Option<Number>`.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,87 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+}
+
+/// ### [7.1.22 ToIndex ( value )](https://tc39.es/ecma262/#sec-toindex)
+pub(crate) fn to_index(agent: &mut Agent, argument: Value) -> JsResult<Value> {
```
aapoalas, line 548: **issue:** Spec says this returns either a non-negative Integer or a throw completion, so the return value should be `JsResult<Number>`.

aapoalas (changes requested):
Looking like a nice start! Some issues and some nitpicks left as comments.

Thanks for working on this <3

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
aapoalas: Yeah, seems so. Need to get them into a private namespace.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -471,3 +471,98 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+        // a. Return key.
+        return Ok(key);
+    }
+
+    // 3. Return ! ToString(key).
+    Ok(to_string(agent, key).unwrap().into())
+}
+
+/// ### [7.1.20 ToLength ( argument )](https://tc39.es/ecma262/#sec-tolength)
+pub(crate) fn to_length(agent: &mut Agent, argument: Value) -> JsResult<Value> {
```
aapoalas, line 491: **issue:** Should return a Number.

nova_vm/src/ecmascript/types/language/object/property_key.rs
```diff
@@ -122,9 +122,26 @@ impl TryFrom<Value> for PropertyKey {
     fn try_from(value: Value) -> Result<Self, Self::Error> {
         match value {
             Value::Integer(x) => Ok(PropertyKey::Integer(x)),
+            Value::Float(x)
```
aapoalas, line 125: **issue:** This should be unnecessary: The Venn diagram of the three different Number variants should have no overlap, meaning that for any `Value::Float(x)` there cannot be a **lossless** function `f` for which `Value::Integer(f(x))` is a valid `Value::Integer` value, and vice versa. Same of course goes for heap numbers.

So given any numeric value N there is only ever one proper Number variant for it.

Thus, for every `Value::Float(x)` it is always trivially true that `x` cannot be represented as a non-negative integer.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -483,3 +486,84 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+        return Some((-0.0).into());
+    }
+
+    // 2. Let n be ! ToNumber(argument).
+    let n = to_number(agent, argument.into()).unwrap();
+
+    // 3. If ! ToString(n) is argument, return n.
+    if to_string(agent, n.into()).unwrap() == argument {
+        return Some(n);
+    }
```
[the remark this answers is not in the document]
aapoalas: It says that the argument is a number or the (canonical, so correct according to spec as opposed to correct according to localization) string representation of a number.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -483,3 +486,84 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+    if let Ok(property_key) = PropertyKey::try_from(key) {
+        return Ok(property_key);
+    }
+
+    // 3. Return ! ToString(key).
+    Ok(to_string(agent, key).unwrap().into())
+}
+
+/// ### [7.1.20 ToLength ( argument )](https://tc39.es/ecma262/#sec-tolength)
+pub(crate) fn to_length(agent: &mut Agent, argument: Value) -> JsResult<Number> {
```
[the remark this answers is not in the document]
aapoalas: We do not have an `Integer` subenum yet. We should though.

That should probably also have a subenum of `UnsignedInteger` and `SignedInteger`: There are a few places in the spec where only positive integers are returned (this is one of them I think). V8 also has a similar distinction.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -483,3 +486,84 @@ pub(crate) fn to_object(_agent: &mut Agent, argument: Value) -> JsResult<Object>
...
+    // 1. Let key be ? ToPrimitive(argument, hint String).
+    let key = to_primitive(agent, argument, Some(PreferredType::String))?;
+
+    // 2. If Type(key) is Symbol, then
+    //    a. Return key.
+    // NOTE: This handles Symbols and other primitives because we use niche
+    // specializations for PropertyKey (e.g. integer indexes for arrays).
+    if let Ok(property_key) = PropertyKey::try_from(key) {
+        return Ok(property_key);
+    }
```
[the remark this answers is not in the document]
aapoalas: As mentioned in Discord, the point of this is to avoid stringifying numbers.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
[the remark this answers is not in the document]
aapoalas: That may be so, but we do not need to worry about that optimisation right now.

nova_vm/src/ecmascript/types/language/object/property_key.rs
aapoalas: I think -0 is not actually a proper indexed integer. The spec is a bit vague on this, -0 is a canonical numeric string but it's not a valid index.

So I think checking for it is not necessary. Need to probably check the spec with a bit finer-toothed a comb to be really sure though.

aapoalas: Yeah, this definitely seems to be how it works. I wonder why on earth the spec has the special casing for `-0`... Presumably it pops up _somewhere_ as some kind of a contradictory special case...

nova_vm/src/ecmascript/types/language/object/property_key.rs
```diff
@@ -122,9 +122,26 @@ impl TryFrom<Value> for PropertyKey {
...
+                if x.fract() == 0.0
+                    && (SmallInteger::MIN_NUMBER..=SmallInteger::MAX_NUMBER)
```
aapoalas, line 127: **nitpick:** Alright so we do need a special case for -0 apparently but it any other Float value being an Integer here should be an assert basically: As mentioned in the above discussion, the Venn diagram of the different number variants should not have any overlap: A safe integer should only ever be represented as a SmallInteger, never a Float with the exception of -0.

aapoalas (approved):
Left one nit

aapoalas (dismissed):

aapoalas (approved):

PR #51, opened 2023-10-31
fix(ecmascript) Fix Number::is_nonzero

vimirage:
https://github.com/trynova/nova/blob/main/nova_vm/src/ecmascript/types/language/number.rs#L165

`!n.is_sign_negative() && !n.is_sign_positive()` is always false.

### DIFF
--- a/nova_vm/src/ecmascript/types/language/number.rs
+++ b/nova_vm/src/ecmascript/types/language/number.rs
@@ -160,12 +160,9 @@ impl Number {
 
     pub fn is_nonzero(self, agent: &mut Agent) -> bool {
         match self {
-            Number::Number(n) => {
-                let n = *agent.heap.get(n);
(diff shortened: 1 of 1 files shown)


nova_vm/src/ecmascript/types/language/number.rs
```diff
@@ -160,12 +160,9 @@ impl Number {
 
     pub fn is_nonzero(self, agent: &mut Agent) -> bool {
         match self {
-            Number::Number(n) => {
-                let n = *agent.heap.get(n);
-                !n.is_sign_negative() && !n.is_sign_positive()
-            }
+            Number::Number(n) => 0.0 != *agent.heap.get(n),
             Number::Integer(_) => true,
```
aapoalas, line 164: Integer can definitely be (positive) zero as well.

PR #53, opened 2023-10-31
feat(ecmascript): is_loosely_equal and is_strictly_equal

eliassjogreen:
This pr implements the is_loosely_equal and is_strictly_equal operators. ~~is_loosely_equal still contains some `todo!`s for the stuff I was unsure about, mostly having to do with bigint related operations.~~

### DIFF
--- a/nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
+++ b/nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
@@ -5,7 +5,7 @@ use crate::ecmascript::{
     types::{bigint, Number, Value},
 };
 
-use super::type_conversion::{to_primitive, PreferredType};
+use super::type_conversion::{string_to_big_int, to_number, to_primitive, PreferredType};
 
 /// ### [7.2.1 RequireObjectCoercible ( argument )](https://tc39.es/ecma262/#sec-requireobjectcoercible)
 ///
@@ -276,4 +276,134 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
         let rny = nx.to_real(agent)?;
         Ok(Some(rnx < rny))
     }
+}
+
+/// [7.2.14 IsLooselyEqual ( x, y )](https://tc39.es/ecma262/#sec-islooselyequal)
+///
+/// The abstract operation IsLooselyEqual takes arguments x (an ECMAScript
+/// language value) and y (an ECMAScript language value) and returns either a
+/// normal completion containing a Boolean or a throw completion. It provides
+/// the semantics for the == operator. It performs the following steps when
+/// called:
+pub(crate) fn is_loosely_equal(
+    agent: &mut Agent,
+    x: impl Into<Value> + Copy,
+    y: impl Into<Value> + Copy,
+) -> JsResult<bool> {
+    let (x, y) = (x.into(), y.into());
+
+    // 1. If Type(x) is Type(y), then
+    if is_same_type(x, y) {
+        // a. Return IsStrictlyEqual(x, y).
+        return Ok(is_strictly_equal(agent, x, y));
+    }
+
+    // 2. If x is null and y is undefined, return true.
+    if x.is_null() && y.is_undefined() {
+        return Ok(true);
+    }
+
+    // 3. If x is undefined and y is null, return true.
+    if x.is_undefined() && y.is_null() {
+        return Ok(true);
+    }
+
+    // 4. Perform the following steps:
+    // a. If x is an Object, x has an [[IsHTMLDDA]] internal slot, and y is either undefined or null, return true.
+    // b. If x is either undefined or null, y is an Object, and y has an [[IsHTMLDDA]] internal slot, return true.
+
+    // 5. If x is a Number and y is a String, return ! IsLooselyEqual(x, ! ToNumber(y)).
+    if x.is_number() && y.is_string() {
+        let y = to_number(agent, y)?;
+        return is_loosely_equal(agent, x, y);
+    }
+
+    // 6. If x is a String and y is a Number, return ! IsLooselyEqual(! ToNumber(x), y).
+    if x.is_string() && y.is_number() {
+        let x = to_number(agent, x)?;
+        return is_loosely_equal(agent, x, y);
+    }
+
+    // 7. If x is a BigInt and y is a String, then
+    if x.is_bigint() && y.is_string() {
+        // a. Let n be StringToBigInt(y).
+        let n = string_to_big_int(agent, y)?;
+
+        // b. If n is undefined, return false.
+        if n.is_undefined() {
+            return Ok(false);
+        }
+
+        // c. Return ! IsLooselyEqual(x, n).
+        return is_loosely_equal(agent, x, n);
+    }
+
+    // 8. If x is a String and y is a BigInt, return ! IsLooselyEqual(y, x).
+    if x.is_string() && y.is_bigint() {
+        return is_loosely_equal(agent, y, x);
+    }
+
+    // 9. If x is a Boolean, return ! IsLooselyEqual(! ToNumber(x), y).
+    if x.is_boolean() {
+        let x = to_number(agent, x)?;
+        return is_loosely_equal(agent, x, y);
+    }
+
+    // 10. If y is a Boolean, return ! IsLooselyEqual(x, ! ToNumber(y)).
+    if y.is_boolean() {
+        let y = to_number(agent, y)?;
+        return is_loosely_equal(agent, x, y);
+    }
+
+    // 11. If x is either a String, a Number, a BigInt, or a Symbol and y is an Object, return ! IsLooselyEqual(x, ? ToPrimitive(y)).
+    if (x.is_string() || x.is_number() || x.is_bigint() || x.is_symbol()) && y.is_object() {
+        let y = to_primitive(agent, y, None)?;
+        return is_loosely_equal(agent, x, y);
+    }
+
+    // 12. If x is an Object and y is either a String, a Number, a BigInt, or a Symbol, return ! IsLooselyEqual(? ToPrimitive(x), y).
+    if x.is_object() && (y.is_string() || y.is_number() || y.is_bigint() || y.is_symbol()) {
+        let x = to_primitive(agent, x, None)?;
+        return is_loosely_equal(agent, x, y);
+    }
+
+    // 13. If x is a BigInt and y is a Number, or if x is a Number and y is a BigInt, then
+    if x.is_bigint() && y.is_number() {
+        // a. If x is not finite or y is not finite, return false.
+        // b. If ℝ(x) = ℝ(y), return true; otherwise return false.
+
+        todo!();
+    }
+
+    // 14. Return false.
+    Ok(false)
+}
+/// [7.2.14 IsStrictlyEqual ( x, y )](https://tc39.es/ecma262/#sec-isstrictlyequal)
+///
+/// The abstract operation IsStrictlyEqual takes arguments x (an ECMAScript
+/// language value) and y (an ECMAScript language value) and returns a Boolean.
+/// It provides the semantics for the === operator. It performs the following
+/// steps when called:
+pub(crate) fn is_strictly_equal(
+    agent: &mut Agent,
+    x: impl Into<Value> + Copy,
+    y: impl Into<Value> + Copy,
+) -> bool {
+    let (x, y) = (x.into(), y.into());
+
+    // 1. If Type(x) is not Type(y), return false.
+    if !is_same_type(x, y) {
+        return false;
+    }
+
+    // 2. If x is a Number, then
+    // NOTE: We need to convert both to a number because we use number
+    // type-safety.
+    if let (Ok(x), Ok(y)) = (x.to_number(agent), y.to_number(agent)) {
+        // a. Return Number::equal(x, y).
+        return x.equal(agent, y);
+    }
+
+    // 3. Return SameValueNonNumber(x, y).
+    same_value_non_number(agent, x, y)
 }
--- a/nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
+++ b/nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
@@ -283,8 +283,7 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
 /// The abstract operation IsLooselyEqual takes arguments x (an ECMAScript
 /// language value) and y (an ECMAScript language value) and returns either a
 /// normal completion containing a Boolean or a throw completion. It provides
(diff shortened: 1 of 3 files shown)


nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
         Ok(Some(rnx < rny))
     }
 }
+
+/// [7.2.14 IsLooselyEqual ( x, y )](https://tc39.es/ecma262/#sec-islooselyequal)
+///
+/// The abstract operation IsLooselyEqual takes arguments x (an ECMAScript
+/// language value) and y (an ECMAScript language value) and returns either a
+/// normal completion containing a Boolean or a throw completion. It provides
+/// the semantics for the == operator. It performs the following steps when
```
aapoalas, line 286: **nitpick:** I've been leaving out the `It performs the following steps when called:` part out of the doc comments since we do not actually write out the steps but instead implement them :)

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+    let (x, y) = (x.into(), y.into());
+
+    // 1. If Type(x) is Type(y), then
+    if is_same_type(x, y) {
+        // a. Return IsStrictlyEqual(x, y).
+        return Ok(is_strictly_equal(agent, x, y));
+    }
+
+    // 2. If x is null and y is undefined, return true.
+    if x.is_null() && y.is_undefined() {
```
aapoalas, line 302: **nitpick:** Just to compress the code down a bit I'd maybe put the inline-comments for steps 2 and 3 one after the other and combine the if checks for these into a single one.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+        return Ok(true);
+    }
+
+    // 3. If x is undefined and y is null, return true.
+    if x.is_undefined() && y.is_null() {
+        return Ok(true);
+    }
+
+    // 4. Perform the following steps:
+    // a. If x is an Object, x has an [[IsHTMLDDA]] internal slot, and y is either undefined or null, return true.
```
aapoalas, line 312: **thought:** I hate the HTML spec for doing this to the world.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+    // b. If x is either undefined or null, y is an Object, and y has an [[IsHTMLDDA]] internal slot, return true.
+
+    // 5. If x is a Number and y is a String, return ! IsLooselyEqual(x, ! ToNumber(y)).
+    if x.is_number() && y.is_string() {
+        let y = to_number(agent, y)?;
```
aapoalas, line 317: **issue:** Spec text says `! ToNumber(y)` so this should be `to_number(agent, y).unwrap()`.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+        return is_loosely_equal(agent, x, y);
+    }
+
+    // 6. If x is a String and y is a Number, return ! IsLooselyEqual(! ToNumber(x), y).
+    if x.is_string() && y.is_number() {
+        let x = to_number(agent, x)?;
```
aapoalas, line 323: **issue:** Same thing here.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -427,6 +427,18 @@ pub(crate) fn to_big_int(agent: &mut Agent, argument: Value) -> JsResult<BigInt>
     todo!()
 }
 
+/// ### [7.1.14 StringToBigInt ( str )](https://tc39.es/ecma262/#sec-stringtobigint)
+pub(crate) fn string_to_big_int(agent: &mut Agent, argument: Value) -> JsResult<Value> {
```
aapoalas, line 431: **issue:** Spec text says that this operation returns a `BigInt` or undefined. We'll convert that to the most convenient form we can which is then `Option<BigInt>`.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+    }
+
+    // 4. Perform the following steps:
+    // a. If x is an Object, x has an [[IsHTMLDDA]] internal slot, and y is either undefined or null, return true.
+    // b. If x is either undefined or null, y is an Object, and y has an [[IsHTMLDDA]] internal slot, return true.
+
+    // 5. If x is a Number and y is a String, return ! IsLooselyEqual(x, ! ToNumber(y)).
+    if x.is_number() && y.is_string() {
+        let y = to_number(agent, y)?;
+        return is_loosely_equal(agent, x, y);
```
aapoalas, line 318: **issue:** Same thing here as well, should unwrap according to spec.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+    }
+
+    // 6. If x is a String and y is a Number, return ! IsLooselyEqual(! ToNumber(x), y).
+    if x.is_string() && y.is_number() {
+        let x = to_number(agent, x)?;
+        return is_loosely_equal(agent, x, y);
```
aapoalas, line 324: **issue:** Same thing here as well, should unwrap according to spec.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+        // a. Let n be StringToBigInt(y).
+        let n = string_to_big_int(agent, y)?;
+
+        // b. If n is undefined, return false.
+        if n.is_undefined() {
+            return Ok(false);
+        }
+
+        // c. Return ! IsLooselyEqual(x, n).
+        return is_loosely_equal(agent, x, n);
```
aapoalas, line 338: **issue:** Same thing here as well, should unwrap according to spec.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+    }
+
+    // 8. If x is a String and y is a BigInt, return ! IsLooselyEqual(y, x).
+    if x.is_string() && y.is_bigint() {
+        return is_loosely_equal(agent, y, x);
```
aapoalas, line 343: **issue:** Same thing here as well, should unwrap according to spec.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+    }
+
+    // 9. If x is a Boolean, return ! IsLooselyEqual(! ToNumber(x), y).
+    if x.is_boolean() {
+        let x = to_number(agent, x)?;
+        return is_loosely_equal(agent, x, y);
```
aapoalas, line 349: **issue:** Same thing here as well, should unwrap according to spec.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+    }
+
+    // 8. If x is a String and y is a BigInt, return ! IsLooselyEqual(y, x).
+    if x.is_string() && y.is_bigint() {
+        return is_loosely_equal(agent, y, x);
+    }
+
+    // 9. If x is a Boolean, return ! IsLooselyEqual(! ToNumber(x), y).
+    if x.is_boolean() {
+        let x = to_number(agent, x)?;
```
aapoalas, line 348: **issue:** Same thing here as well, should unwrap according to spec.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+        return is_loosely_equal(agent, x, y);
+    }
+
+    // 10. If y is a Boolean, return ! IsLooselyEqual(x, ! ToNumber(y)).
+    if y.is_boolean() {
+        let y = to_number(agent, y)?;
```
aapoalas, line 354: **issue:** Same thing here as well, should unwrap according to spec.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+        return is_loosely_equal(agent, x, y);
```
aapoalas, line 355: **issue:** Same thing here as well, should unwrap according to spec.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+    }
+
+    // 11. If x is either a String, a Number, a BigInt, or a Symbol and y is an Object, return ! IsLooselyEqual(x, ? ToPrimitive(y)).
+    if (x.is_string() || x.is_number() || x.is_bigint() || x.is_symbol()) && y.is_object() {
+        let y = to_primitive(agent, y, None)?;
+        return is_loosely_equal(agent, x, y);
```
aapoalas, line 361: **issue:** Same thing here as well, should unwrap according to spec.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,133 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+    }
+
+    // 12. If x is an Object and y is either a String, a Number, a BigInt, or a Symbol, return ! IsLooselyEqual(? ToPrimitive(x), y).
+    if x.is_object() && (y.is_string() || y.is_number() || y.is_bigint() || y.is_symbol()) {
+        let x = to_primitive(agent, x, None)?;
+        return is_loosely_equal(agent, x, y);
```
aapoalas, line 367: **issue:** Same thing here as well, should unwrap according to spec.

aapoalas (changes requested):
Nice work, thanks for doing this! A few issues are repeating regarding interpretation of spec text, essentially. Just a tad bit of repetitive logic work.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
eliassjogreen: I despise it. [It's also the reason `document.all == undefined` 😭](https://tc39.es/ecma262/#sec-IsHTMLDDA-internal-slot). But considering that:
> There are no other known examples of this type of object and implementations should not create any with the exception of `document.all`.

I think it is probably fine to leave this part out, at least for now.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -277,3 +277,137 @@ pub(crate) fn is_less_than<const LEFT_FIRST: bool>(
...
+/// The abstract operation IsLooselyEqual takes arguments x (an ECMAScript
+/// language value) and y (an ECMAScript language value) and returns either a
+/// normal completion containing a Boolean or a throw completion. It provides
+/// the semantics for the == operator.
+pub(crate) fn is_loosely_equal(
+    agent: &mut Agent,
+    x: impl Into<Value> + Copy,
+    y: impl Into<Value> + Copy,
+) -> JsResult<bool> {
+    let (x, y) = (x.into(), y.into());
```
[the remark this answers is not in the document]
eliassjogreen: What do you mean by this? Like this?:
```ts
    fn inner(x: Value, y: Value) {
      ...
    }
    
    return inner(x.into(), y.into());
```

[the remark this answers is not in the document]
eliassjogreen: Do [these changes](https://github.com/trynova/nova/pull/53/commits/773d8cb13216e56bf75b5ddbbe1607d69584d03a) do what you mean?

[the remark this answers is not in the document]
aapoalas: The point sort of is that the entire function is generic for the parameters. It's not polymorphism but specialised monomorphism.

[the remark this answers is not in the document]
aapoalas: Discussed on Discord, I prefer the original way (keep the generics for the whole function in hopes of better compiler code generation), @vimirage prefers the suggested way (use separate internal function so that the function doesn't get compiled separately for each possible set of generic parameters but still allows for the convenience of not having to call function with `func(a.into(), b.into())` every time).

So, do as you prefer @eliassjogreen :)

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
aapoalas: Yeah, absolutely no reason to complicate our lives with this right now.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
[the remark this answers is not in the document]
eliassjogreen: I will do as you prefer, I find it easier to read and hopefully LLVM is kind to us when it comes to codegen.

sno2:
Thank you

PR #55, opened 2023-10-31
fix(ecmascript): Fix `Value::is_*` operations that didn't take all alternative variants into account

andreubotella:


aapoalas (approved):
Thanks. The `is_object` is somewhat annoying to upkeep... Clippy doesn't like it but it might be worth it to turn it into an explicit check where all variants are mentioned.

PR #58, opened 2023-10-31
fix(vm): Fix the ranges for `SmallInteger`

andreubotella:
Since `SmallInteger` represents numbers in two's complement, the minimum value for that type is -2^55, rather than -2^55 + 1, even though its maximum value is 2^55 - 1. This is because there are as many negative numbers representable in the type as non-negative numbers, but the non-negative numbers include zero. This patch updates `SmallInteger::MIN_BIGINT` to reflect this.

This patch also updates `SmallInteger::{MIN,MAX}_NUMBER` to remove the division by 2, to make them truly reflect Javascript's `Number.{MIN,MAX}_SAFE_INTEGER`. It also changes the ranges in `SmallInteger::from_i64_unchecked` and in the `TryFrom<i64>` impl to use `{MIN,MAX}_BIGINT`, rather than `{MIN,MAX}_NUMBER`.

### DIFF
--- a/nova_vm/src/engine/small_integer.rs
+++ b/nova_vm/src/engine/small_integer.rs
@@ -11,11 +11,11 @@ impl std::fmt::Debug for SmallInteger {
 }
 
 impl SmallInteger {
-    pub const MIN_BIGINT: i64 = -(2i64.pow(56)) / 2 + 1;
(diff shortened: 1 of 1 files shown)


aapoalas (approved):
LGTM, nice and thanks <3

PR #59, opened 2023-10-31
feat(vm): Add conversions from integer types to `SmallInteger`

andreubotella:
This patch also makes `SmallInteger::from_i64_unchecked` a private method, to prevent misuse.

-----------

This PR depends on #58 and should not land before it.


aapoalas (approved):
Nice stuff, thanks a lot! <3

Just needs a rebase and I'll merge it.

andreubotella:
Oh, I just realized I'd left some `unsafe` blocks from a previous version of this patch that made `from_i64_unchecked` unsafe and which I changed at the last minute.

aapoalas (approved):
Thank you <3

Issue #17, opened 2022-10-06
Wasm | varint - incorrect counts for expected bytes


vimirage:
https://github.com/trynova/nova/blob/f903792ad16f73275894fe710f6b39842fbfd6a0/wasm/src/varint.rs#L56-L61

U64 may definitely read over 5 bytes.

Furthermore, the `expected`/`found` situation (between both u32 and u64) is possibly incorrect (and somewhat weird), as for u32 one may have 5 bytes, with 0xFF for the first four, and 0x7F for the last - this is 4 continuation bits, and 5*7 = 35 bits in the integer value. 35 bits (2³⁵ - 1) will overflow the u32.

MierenManz:
It doesn't make sense to read over the count. Only that we do. Will change the error

PR #62, opened 2023-10-31
fix(webassembly) fix varint decoding message

ghost:
Resolves https://github.com/trynova/nova/issues/17

I think the code is all wrong. Eventually I'm going to read the entire spec and implement it myself here <.<


aapoalas (changes requested):
Test fails :(

> test varint::test::decode_32_over_u32_max ... FAILED

aapoalas (approved):

PR #64, opened 2023-11-03
feat(execution): implement more environments logic

sno2:
This will place us in a better position for a basic bytecode later.


aapoalas (approved):
Looks good to me.

nova_vm/src/ecmascript/execution/environments.rs
```diff
@@ -1,17 +1,68 @@
...
+/// of the current evaluation of the surrounding function.
+pub(super) type OuterEnv = Option<EnvironmentIndex>;
+
+macro_rules! create_environment_index {
+    ($name: ident, $index: ident) => {
+        #[derive(Debug, Clone, Copy)]
+        pub(crate) struct $index(NonZeroU32, PhantomData<$name>);
+
+        impl $index {
+            pub(crate) const fn from_u32_index(value: u32) -> Self {
```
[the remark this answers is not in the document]
sno2: It might happen...I'd rather just leave it as-is.

nova_vm/src/ecmascript/execution/environments.rs
```diff
@@ -1,17 +1,68 @@
...
+                assert!(value != u32::MAX);
+                // SAFETY: Number is not max value and will not overflow to zero.
+                // This check is done manually to allow const context.
+                Self(unsafe { NonZeroU32::new_unchecked(value + 1) }, PhantomData)
+            }
+
+            pub(crate) const fn from_usize_index(value: usize) -> Self {
```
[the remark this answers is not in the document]
sno2: Not sure, I just made the code use macros. I don't think I wrote that code.

nova_vm/src/ecmascript/execution/environments/declarative_environment.rs
```diff
@@ -1,56 +1,292 @@
...
 pub struct DeclarativeEnvironment {
-    pub outer_env: Option<EnvironmentIndex>,
-    pub bindings: HashMap<String, Binding>,
+    /// ### \[\[OuterEnv\]\]
+    ///
+    /// See [OuterEnv].
+    outer_env: OuterEnv,
+
+    /// The environment's bindings.
+    bindings: HashMap<Atom, Binding>,
```
aapoalas, line 23: **thought:** I'm a bit worried about how the `Atom` will work in the long run if this needs to interact with `Value::String`. But, that's a thing to ponder another day.

nova_vm/src/ecmascript/execution/environments/declarative_environment.rs
```diff
@@ -1,56 +1,292 @@
...
 }
 
-// TODO: Consider splitting binding this into ValueBinding
```
aapoalas, line 37: **nitpick:** IMO this TODO is still relevant and not entirely covered by a bitfield packing TODO. The idea for the splitting was to (possibly) do a struct-of-vectors type of thing so that `get_binding_value` and any iteration over the bindings (GC) would not need to load the metadata.

nova_vm/src/ecmascript/execution/environments/declarative_environment.rs
```diff
@@ -1,56 +1,292 @@
...
+    /// envRec takes arguments N (a String) and S (a Boolean) and returns either
+    /// a normal completion containing an ECMAScript language value or a throw
+    /// completion. It returns the value of its bound identifier whose name is
+    /// N. If the binding exists but is uninitialized a ReferenceError is
+    /// thrown, regardless of the value of S.
+    pub(crate) fn get_binding_value(
+        &self,
+        agent: &mut Agent,
+        name: &Atom,
+        _is_strict: bool,
```
aapoalas, line 227: **nitpick:** Since this function doesn't actually use the boolean and we're not implementing these as traits or anything, I'd just drop the parameter: We can leave it as a comment line.

nova_vm/src/ecmascript/execution/environments/global_environment.rs
```diff
@@ -1,68 +1,83 @@
...
+    /// Contains bindings for all declarations in global code for the associated
+    /// realm code except for FunctionDeclaration, GeneratorDeclaration,
+    /// AsyncFunctionDeclaration, AsyncGeneratorDeclaration, and
+    /// VariableDeclaration bindings.
+    declarative_record: DeclarativeEnvironment,
 
-    /// [[OuterEnv]]
+    /// ### \[\[VarNames\]\]
     ///
-    /// Per https://tc39.es/ecma262/#sec-the-environment-record-type-hierarchy:
```
aapoalas, line 53: **nitpick:** Could maybe move the spec quote about outer env into the struct description field. The null field doesn't hurt the size of the struct though so it's also an option to keep it here, just for documentation's sake.

nova_vm/src/ecmascript/execution/environments/global_environment.rs
```diff
@@ -1,68 +1,83 @@
...
+        GlobalEnvironment {
+            // 4. Set env.[[ObjectRecord]] to objRec.
+            object_record,
+
+            // 5. Set env.[[GlobalThisValue]] to thisValue.
             global_this_value: this_value,
-            declarative_record: Default::default(),
+
+            // 6. Set env.[[DeclarativeRecord]] to dclRec.
+            declarative_record,
```
aapoalas, line 74: **nitpick:** Ah, now I remember was why I'd inlined these. It's the entirely unnecessary `OuterEnv` fields that crop up in this case :(

aapoalas (approved):
Had some thoughts and maybe some nitpicks but nothing really worth thinking about too deeply.

nova_vm/src/ecmascript/execution/environments/declarative_environment.rs
sno2: True, we should probably make our own js parser...someday....

PR #66, opened 2023-11-04
feat(builtins): more `Number` constructor methods

sno2:
Now, we only need to finish the `Number` constructor, `Number.parseFloat`, and `Number.parseInt`.

### DIFF
--- a/nova_vm/src/ecmascript/builtins/number.rs
+++ b/nova_vm/src/ecmascript/builtins/number.rs
@@ -319,14 +319,9 @@ impl NumberConstructor {
         let maybe_number = arguments.get(0);
 
         // 1. If IsIntegralNumber(number) is true, then
-        if is_integral_number(agent, maybe_number) {
-            // a. If abs(ℝ(number)) ≤ 2**53 - 1, return true.
-            if maybe_number.to_real(agent).unwrap() <= (2.0f64.powi(53) - 1.0) {
-                return Ok(true.into());
-            }
-        }
-
+        //    a. If abs(ℝ(number)) ≤ 2**53 - 1, return true.
         // 2. Return false.
-        Ok(false.into())
+        // NOTE: Integers must be stored in `Value::Integer`.
+        Ok(matches!(maybe_number, Value::Integer(_)).into())
     }
 }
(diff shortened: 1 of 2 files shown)


nova_vm/src/ecmascript/builtins/number.rs
```diff
@@ -224,4 +265,68 @@ impl NumberConstructor {
         // 5. Set O.[[NumberData]] to n.
         // 6. Return O.
     }
+
+    /// ### [21.1.2.2 Number.isFinite ( number )](https://tc39.es/ecma262/#sec-number.isfinite)
```
aapoalas, line 269: **thought:** I had been using the same number of `#` characters as there are "levels" in the spec definition number, so `21.1.2.2` would be `####`.

That being said, it's not necessarily a very logical system so I definitely do not mind it not being used :D

nova_vm/src/ecmascript/builtins/number.rs
```diff
@@ -224,4 +265,68 @@ impl NumberConstructor {
...
+    // ### [21.1.2.5 Number.isSafeInteger ( number )](21.1.2.5 Number.isSafeInteger ( number ))
+    fn is_safe_integer(
+        agent: &mut Agent,
+        _this_value: Value,
+        arguments: ArgumentsList,
+    ) -> JsResult<Value> {
+        let maybe_number = arguments.get(0);
+
+        // 1. If IsIntegralNumber(number) is true, then
+        if is_integral_number(agent, maybe_number) {
```
aapoalas, line 322: **issue:** This can be optimised into just a `Number::Integer` check. At least until we support 32-bit environments where `Integer` cannot handle the full safe integer range, if we ever do, it should never be the case that any other number is a safe integer.

aapoalas (changes requested):
Looks good to me except for SafeInteger check which can be done with a direct enum check. Admittedly, that's a bit of a nitpick but eh.

nova_vm/src/ecmascript/builtins/number.rs
sno2: Yeah, I realized the meaning after I saw two headers being used. Although, I like the look of how `###` looks in the code and in intellisense as an `h3`. Do you think it would be all right if we just used `###` for everything, and people can use the numbering to understand the level?

aapoalas: Yeah, sure I guess.

I'd like to keep the module level headers different though, maybe.

aapoalas (approved):

PR #67, opened 2023-11-04
fix(builtins): make ordinary methods use `SameValue`

sno2:
I am unsure if this was intentional (or if I wrote this code), but I noticed this while doing a fly-by of the ordinary methods.


aapoalas (approved):
LGTM

PR #69, opened 2023-11-05
feat: better Agent <=> Realm relationship

sno2:
We now pass around realms as identifier which allows us to avoid locking thanks to AapoAlas's idea.

### DIFF
--- a/nova_vm/src/ecmascript/execution/realm.rs
+++ b/nova_vm/src/ecmascript/execution/realm.rs
@@ -55,7 +55,7 @@ impl<'ctx, 'host> Realm<'ctx, 'host> {
     fn create(agent: &mut Agent<'ctx, 'host>) -> RealmIdentifier<'ctx, 'host> {
         // TODO: implement spec
         let realm = Self {
-            global_env: GlobalEnvironmentIndex::from_u32_index(0),
+            global_env: GlobalEnvironmentIndex::from_index(0),
             global_object: Object::Object(ObjectIndex::from_u32_index(0)),
             host_defined: None,
             intrinsics: Intrinsics::default(),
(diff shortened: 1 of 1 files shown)


nova_vm/src/ecmascript/execution/realm.rs
```diff
@@ -96,14 +95,13 @@ impl<'ctx, 'host> Realm<'ctx, 'host> {
         let this_value = this_value.unwrap_or(global_object);
 
         // 4. Set realmRec.[[GlobalObject]] to globalObj.
-        self.global_object = global_object;
+        agent.heap.get_realm_mut(realm_id).global_object = global_object;
```
aapoalas, line 98: **thought:** I don't know if it would be useful but we might benefit from eg. having `get_realm_mut` and similar kind of APIs at the `agent.heap.realms` level, and having Environments related APIs at the `agent.heap.environments` level etc.

The reason being that then it would be possible to "split up" the borrow of the `agent` like this:
```rs
let &mut Heap {
  realms,
  environments,
  objects,
  ...
} = agent.heap;
```

Now all of those can be individually borrowed at the same time. As long as the internal APIs do not require access to the other parts of the agent / heap then they can be used this way.

But that's not a thing we need to worry about right now :)

aapoalas (approved):
Looks good to me.

nova_vm/src/ecmascript/execution/realm.rs
sno2: Interesting, that could help out. I meant for `get_realm` to be synonymous with [_the current Realm Record_](https://tc39.es/ecma262/#current-realm) in the specification. However, we could get away with accessing the heap after that. If only we could force functions like these to be inlined like macros so that borrows could be checked more precisely.

aapoalas (approved):

PR #70, opened 2023-11-06
feat(ecmascript): operations on iterator objects

eliassjogreen:
Adds some of the [operations on iterator objects](https://tc39.es/ecma262/#sec-operations-on-iterator-objects).

- [x] get_iterator_from_method
- [ ] get_iterator (step 1.b.iv remains)
- [x] iterator_next
- [x] iterator_complete
- [x] iterator_value
- [x] iterator_step
- [ ] iterator_close
- [ ] if_abrupt_close_iterator
- [ ] async_iterator_close
- [ ] create_iter_result_object
- [ ] create_list_iterator_record
- [x] iterator_to_list

### DIFF
--- a/nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
+++ b/nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
@@ -257,7 +257,7 @@ pub(crate) fn create_iter_result_object(agent: &mut Agent, value: Value, done: b
 /// of ECMAScript language values) and returns an Iterator Record. It creates
 /// an Iterator (27.1.1.2) object record whose next method returns the
 /// successive elements of list.
-pub(crate) fn create_list_iterator_record(agent: &mut Agent, list: Vec<Value>) -> JsResult<Value> {
+pub(crate) fn create_list_iterator_record(agent: &mut Agent, list: &[Value]) -> JsResult<Value> {
     // 1. Let closure be a new Abstract Closure with no parameters that captures list and performs the following steps when called:
     // a. For each element E of list, do
     // i. Perform ? GeneratorYield(CreateIterResultObject(E, false)).
--- a/nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
+++ b/nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
@@ -7,14 +7,22 @@ use super::{
 use crate::{
     ecmascript::{
         abstract_operations::operations_on_objects::get_method,
-        execution::{
-            agent::{ExceptionType, JsError},
-            Agent, JsResult,
-        },
-        types::{Function, Object, OrdinaryObject, PropertyKey, Value},
+        execution::{agent::ExceptionType, Agent, JsResult},
+        types::{Function, Object, PropertyKey, String, Value},
     },
     heap::WellKnownSymbolIndexes,
 };
+
+/// [7.4.1 Iterator Records](https://tc39.es/ecma262/#sec-iterator-records)
+///
+/// An Iterator Record is a Record value used to encapsulate an Iterator or
+/// AsyncIterator along with the next method.
+#[derive(Debug)]
+pub(crate) struct IteratorRecord {
+    iterator: Object,
+    next_method: Value,
+    done: bool,
+}
 
 /// [7.4.2 GetIteratorFromMethod ( obj, method )](https://tc39.es/ecma262/#sec-getiteratorfrommethod)
 ///
@@ -26,7 +34,7 @@ pub(crate) fn get_iterator_from_method(
     agent: &mut Agent,
     obj: Value,
     method: Function,
-) -> JsResult<Object> {
+) -> JsResult<IteratorRecord> {
     // 1. Let iterator be ? Call(method, obj).
     let iterator = call(agent, method.into(), obj, None)?;
 
@@ -36,12 +44,15 @@ pub(crate) fn get_iterator_from_method(
     };
 
     // 3. Let nextMethod be ? Get(iterator, "next").
-    let property = PropertyKey::from_str(&mut agent.heap, "next");
-    let next_method = get(agent, iterator, property)?;
+    let next_method = get(agent, iterator, String::from_small_string("next").into())?;
 
     // 4. Let iteratorRecord be the Iterator Record { [[Iterator]]: iterator, [[NextMethod]]: nextMethod, [[Done]]: false }.
     // 5. Return iteratorRecord.
-    todo!()
+    Ok(IteratorRecord {
+        iterator,
+        next_method,
+        done: false,
+    })
 }
 
 /// [7.4.3 GetIterator ( obj, kind )](https://tc39.es/ecma262/#sec-getiterator)
@@ -49,7 +60,11 @@ pub(crate) fn get_iterator_from_method(
 /// The abstract operation GetIterator takes arguments obj (an ECMAScript
 /// language value) and kind (sync or async) and returns either a normal
 /// completion containing an Iterator Record or a throw completion.
-pub(crate) fn get_iterator(agent: &mut Agent, obj: Value, is_async: bool) -> JsResult<Object> {
+pub(crate) fn get_iterator(
+    agent: &mut Agent,
+    obj: Value,
+    is_async: bool,
+) -> JsResult<IteratorRecord> {
     // 1. If kind is async, then
     let method = if is_async {
         // a. Let method be ? GetMethod(obj, @@asyncIterator).
@@ -94,10 +109,10 @@ pub(crate) fn get_iterator(agent: &mut Agent, obj: Value, is_async: bool) -> JsR
 
     // 3. If method is undefined, throw a TypeError exception.
     let Some(method) = method else {
-        Err(agent.throw_exception(
+        return Err(agent.throw_exception(
             ExceptionType::TypeError,
             "Iterator method cannot be undefined",
-        ))
+        ));
     };
 
     // 4. Return ? GetIteratorFromMethod(obj, method).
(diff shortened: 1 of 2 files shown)


nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,103 @@
...
+) -> JsResult<Object> {
+    // 1. Let iterator be ? Call(method, obj).
+    let iterator = call(agent, method.into(), obj, None)?;
+
+    // 2. If iterator is not an Object, throw a TypeError exception.
+    let iterator = if let Ok(iterator) = to_object(agent, iterator) {
+        iterator
+    } else {
+        return Err(agent.throw_exception(ExceptionType::TypeError, "Iterator is not object"));
+    };
```
sno2, line 31: Note that you can just do this:

```suggestion
    let Ok(iterator) = to_object(agent, iterator) else {
        return Err(agent.throw_exception(ExceptionType::TypeError, "Iterator is not object"));
    };
```

sno2: Yeah, nifty in our project with how we dance around with enums :)

aapoalas: I didn't know this either, and indeed VERY! handy! Man, I love this right now. Thank you @sno2 !

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,302 @@
...
+    // 4. Return ? GetIteratorFromMethod(obj, method).
+    if let Some(method) = method {
+        get_iterator_from_method(agent, obj, method)
+    } else {
+        // 3. If method is undefined, throw a TypeError exception.
+        Err(agent.throw_exception(
+            ExceptionType::TypeError,
+            "Iterator method cannot be undefined",
+        ))
+    }
```
sno2, line 104: little cleanup

```suggestion
    // 3. If method is undefined, throw a TypeError exception.
    let Some(method) = method else {
        Err(agent.throw_exception(
            ExceptionType::TypeError,
            "Iterator method cannot be undefined",
        ))
    };

    // 4. Return ? GetIteratorFromMethod(obj, method).
    get_iterator_from_method(agent, obj, method)
```

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    // 3. Repeat, while next is not false,
+    while next.is_true() {
+        // a. Set next to ? IteratorStep(iteratorRecord).
+        next = iterator_step(agent, iterator_record)?;
+
+        // b. If next is not false, then
+        if !next.is_false() {
+            // SAFETY: Because iterator_step returns either false or an object this is always safe.
+            let next = unsafe { next.try_into().unwrap_unchecked() };
```
sno2, line 290: IMO it would be better if we changed `iterator_step`'s return type to `Option<Object>` and have the `false` state be `None`.

This would make this unsafe code obsolete and you would be able to do an `if let Some(next) = next {` above here.

sno2 (changes requested):
A few suggestions, otherwise LGTM 👍

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,302 @@
...
+/// [7.4.2 GetIteratorFromMethod ( obj, method )](https://tc39.es/ecma262/#sec-getiteratorfrommethod)
+///
+/// The abstract operation GetIteratorFromMethod takes arguments obj (an
+/// ECMAScript language value) and method (a function object) and returns
+/// either a normal completion containing an Iterator Record or a throw
+/// completion.
+pub(crate) fn get_iterator_from_method(
+    agent: &mut Agent,
+    obj: Value,
+    method: Function,
```
aapoalas, line 28: **issue:** When the spec says "iterator record" it does not refer to an object of a given form, not as far as I can tell anyhow.

It refers to a particular struct form, essentially, which contains the fields described [here](https://tc39.es/ecma262/#sec-iterator-records). So in effect this should make it much easier to implement some of these steps, eg. steps 4 and 5 here are just returning a struct.

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,302 @@
...
+
+/// [7.4.4 IteratorNext ( iteratorRecord [ , value ] )](https://tc39.es/ecma262/#sec-iteratornext)
+///
+/// The abstract operation IteratorNext takes argument iteratorRecord (an
+/// Iterator Record) and optional argument value (an ECMAScript language value)
+/// and returns either a normal completion containing an Object or a throw
+/// completion.
+pub(crate) fn iterator_next(
+    agent: &mut Agent,
+    iterator_record: Object,
```
aapoalas, line 115: **issue**: As above, the iterator record should be a struct. I think we'd take it by reference here.

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+/// [7.4.5 IteratorComplete ( iterResult )](https://tc39.es/ecma262/#sec-iteratorcomplete)
+///
+/// The abstract operation IteratorComplete takes argument iterResult (an
+/// Object) and returns either a normal completion containing a Boolean or a
+/// throw completion.
+pub(crate) fn iterator_complete(agent: &mut Agent, iter_result: Object) -> JsResult<Value> {
+    // 1. Return ToBoolean(? Get(iterResult, "done")).
+    let property = PropertyKey::from_str(&mut agent.heap, "done");
```
aapoalas, line 134: **nitpick:** This can be done with a SmallString without involving the heap in the creation which might make this easier for the compiler to inline.

Not sure if it's necessary though.

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,302 @@
...
+
+/// [7.4.7 IteratorStep ( iteratorRecord )](https://tc39.es/ecma262/#sec-iteratorstep)
+///
+/// The abstract operation IteratorStep takes argument iteratorRecord (an
+/// Iterator Record) and returns either a normal completion containing either
+/// an Object or false, or a throw completion. It requests the next value from
+/// iteratorRecord.\[\[Iterator\]\] by calling iteratorRecord.\[\[NextMethod\]\]
+/// and returns either false indicating that the iterator has reached its end
+/// or the IteratorResult object if a next value is available.
+pub(crate) fn iterator_step(agent: &mut Agent, iterator_record: Object) -> JsResult<Value> {
```
aapoalas, line 158: Here the same thing, custom struct by reference.

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,302 @@
...
+
+/// [7.4.8 IteratorClose ( iteratorRecord, completion )](https://tc39.es/ecma262/#sec-iteratorclose)
+///
+/// The abstract operation IteratorClose takes arguments iteratorRecord (an
+/// Iterator Record) and completion (a Completion Record) and returns a
+/// Completion Record. It is used to notify an iterator that it should perform
+/// any actions it would normally perform when it has reached its completed state.
+pub(crate) fn iterator_close(
+    agent: &mut Agent,
+    iterator_record: Object,
```
aapoalas, line 182: Ditto

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,302 @@
...
+    todo!()
+}
+
+/// [7.4.9 IfAbruptCloseIterator ( value, iteratorRecord )](https://tc39.es/ecma262/#sec-ifabruptcloseiterator)
+///
+/// IfAbruptCloseIterator is a shorthand for a sequence of algorithm steps that use an Iterator Record.
+pub(crate) fn if_abrupt_close_iterator(
+    agent: &mut Agent,
+    value: JsResult<Value>,
+    iterator_record: Object,
```
aapoalas, line 205: Ditto

aapoalas (changes requested):
Need to implement the IteratorRecord struct. That should make implementing these algorithms a lot simpler :)

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
eliassjogreen: Oh, ok. Was planning on asking in the discord when I had time but seems like that wont be needed, thanks! Where should I define this struct? I guess it belongs in this file considering it was defined in this section of the spec?

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
eliassjogreen: Sounds good!

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
aapoalas: Yeah. It seems like a stupid place to define it but since this is the place in the spec then let's follow it.

It might turn out to be a good place to define it.

sno2 (dismissed):

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,327 @@
...
+        "The iterator result was not an object",
+    )))
+}
+
+/// [7.4.5 IteratorComplete ( iterResult )](https://tc39.es/ecma262/#sec-iteratorcomplete)
+///
+/// The abstract operation IteratorComplete takes argument iterResult (an
+/// Object) and returns either a normal completion containing a Boolean or a
+/// throw completion.
+pub(crate) fn iterator_complete(agent: &mut Agent, iter_result: Object) -> JsResult<Value> {
```
aapoalas, line 157: **nitpick:** Not your fault obviously since the return value type comes from `to_boolean` but this should probably be `JsResult<bool>`.

aapoalas (dismissed):
Looks good to me, thank you! <3

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -0,0 +1,327 @@
...
+) -> JsResult<Object> {
+    // 1. If value is not present, then
+    // a. Let result be ? Call(iteratorRecord.[[NextMethod]], iteratorRecord.[[Iterator]]).
+    // 2. Else,
+    // a. Let result be ? Call(iteratorRecord.[[NextMethod]], iteratorRecord.[[Iterator]], « value »).
+    let result = call(
+        agent,
+        iterator_record.next_method,
+        iterator_record.iterator.into(),
+        value.as_ref().map(|v| std::slice::from_ref(v)),
```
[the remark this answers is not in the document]
eliassjogreen: Yeah I think so. This was the cleanest way I found to convert the `Option<Value>` to `Option<&[Value]>`.

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
eliassjogreen: Yeah leaving this for the future then.

aapoalas (approved):

sno2 (approved):

PR #73, opened 2023-11-10
Minor clean up for Value

ghost:
Discriminants.
There's less code now.
It's not any more fragile.

Furthermore, `value_discriminant` isn't necessary here.

### DIFF
--- a/nova_vm/src/ecmascript/types/language/value.rs
+++ b/nova_vm/src/ecmascript/types/language/value.rs
@@ -19,6 +19,8 @@ use crate::{
 
 use super::{BigInt, Number};
 
+// TODO: Handle this and enum subtypes via a macro for compile-time guarantees and safety
+
 /// 6.1 ECMAScript Language Types
 /// https://tc39.es/ecma262/#sec-ecmascript-language-types
 #[derive(Debug, Clone, Copy, PartialEq, Default)]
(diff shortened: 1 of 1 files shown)


aapoalas (changes requested):
I'm not really in favour of this: This now means that we have effectively two places for defining the order of items in the enun: The defining place is now the discriminant values which currently are just integers in increasing order. Then there is the enum itself which is what one would usually look for first.

If eg. I wanted to change the order of Number and Integer (so as to get all PropertyKey enum values in a row) I would have to manually change each of the integers being reordered, instead of just moving one enum entry.

And if we add an object entry into the enum and want to keep alphabetical order then all the integers need to be shifted over by one.

If we we're doing some sort of interesting bitmasking with the discriminants then I'd support this but as long as it's just integers in increasing order (from 1) then I do not see much benefit to doing this manually.

aapoalas:
> Such as? Our existing optimisations would still work.

Such as having a particular bit only be set for object-like types, or having all PropertyKey identifiers be after one another so that checking for PropertyKey-ness can be done with a range check on the discriminant.

> It's that they won't know which discriminant to use for overwriting heap values.
> 
> Even a minor incovenience is good, but the idea is that an attacker cannot generate anything. https://discord.com/channels/1012084654389604372/1012084654389604376/1170127670516916306 Or for example, if Wasm is disabled, but an attacker hopes to create a Wasm function w/ invalid assembly and execute it through another flaw we have, they won't know which discriminant to write. If the attacker gets to read only a single arraybuffer as a binary pointer, and they want to turn it into.. a dom gpu buffer or something to access some dom apis that are restricted (think: some apis accept objects that can't be created without privileges), they can't know which discriminant to use. More likely, they'll have to shy away or risk a runtime crash. This is a boon for security.

An ArrayBuffer will not become a SharedArrayBuffer or a DOM GPU buffer by simply changing its discriminant; the backing data is different. Especially for exotic objects related to DOM APIs defined in the HTML spec we're then talking about host defined objects which is something that the Value discriminant will have nothing to do with. And even if it did, again changing the discriminant wouldn't change the backing data so an attacker would also need to change the backing data somehow in such a way as to match the actual expected API, possibly also including a `dyn Object` pointer if host defined objects end up trait-defined.

And if the attacker can do that, then they can do that absolutely regardless of the discriminant of the enum and we likely have much bigger issues, as this is full on foreign code execution.

sno2:
Stale and account deleted.

PR #79, opened 2023-12-26
feat(heap): Remove self-referential lifetimes

aapoalas:
Our Heap currently has two lifetimes flying around:
1. `'host`: This is the lifetime of the Host Defined hooks.
2. `'ctx`: This is a somewhat vaguely defined lifetime copied over from Boa JS engine, where `&'ctx Context` is roughly equal to `&'agent Agent` for us.

Copying a page from Boa JS engine, they set `&'static dyn HostDefined` so we can just make HostDefined as static. This is likely entirely fine as it's quite unlikely for any embedder to have dynamically changing Host Defined hooks. This allows us to drop the `'host` lifetime.

The `'ctx` lifetime is more complicated. Its main place of usage ends up being the `&'ctx str` source text of Scripts and the `Program<'ctx>` oxc parsing result that refers to the source text. This seems like a nice system but it essentially means that the user of the engine lib must keep all script sources alive until they drop the `Agent`, and the script sources end up living outside of the Heap. This probably isn't what we want (I think that eg. V8 takes ownership of script sources into the heap). I think what we want is for the script sources to be in the Heap, and then to have intra-Heap references from eg. ECMAScriptFunctions to the `Script`. Preferably we'd not make those references actual pointers but rather eg. `oxc_span::Span` structs.

But at least some actual references do remain, as the `Program<'_>` struct does refer (at least at a lifetime level if not by actual pointers) to the source text. And both the Program and the script source are in the Heap. This makes the Heap self-referential, which is not representable in a safe or exact manner with Rust lifetimes. Instead, the lifetimes must be made `'static` and in some ways that is also correct: As long as the `ECMAScriptFunction` struct exists in the heap, it keeps the `Script` struct in the heap which then means that the `Program<'_>` and the `String` source text are both alive in the heap: From the point of view of the `ECMAScriptFunction` the lifetime of is `'static`.


sno2 (dismissed):

nova_vm/src/ecmascript/scripts_and_modules/script.rs
```diff
@@ -68,10 +68,15 @@ pub struct Script<'ctx, 'host> {
     ///
     /// Field reserved for use by host environments that need to associate
     /// additional information with a script.
-    pub(crate) host_defined: Option<HostDefined<'host>>,
+    pub(crate) host_defined: Option<HostDefined>,
+
+    /// Source text of the script
+    ///
+    /// Parsing a script takes ownership of the text.
+    source_text: String,
```
sno2, line 76: It doesn't matter _that_ much, but `Box<str>` might be a better choice to avoid doing any accidental mutations.

PR #82, opened 2024-01-03
feat(gc): Implement heap sweep and compaction

aapoalas:
Implement sweeping and compacting of heap vectors. The algorithm is single-threaded for now but the compacting can be made parallel trivially. Marking can be made parallel with relative ease as well: The marking is currently at its core a "vector of own indexes to vectors of other indexes" transformation algorithm. Making it parallel would require adding mpsc channels for receiving the indexes from other threads but otherwise it would be fairly trivial.

Making the marking concurrent is the true eventual goal: That will require making the Heap itself concurrent-safe. That is not so simple: It could be done in a simple-ish sort of way with Atomics and RwLocks and Mutexes and whatnot, but that would likely end up being way too slow at least in the long run. Also, not particularly interesting. It may end up be what I'll do as a stepping stone but eventually I want to make the Heap at-least-mostly RCU based. That will be fun. It is work for another day.


nova_vm/src/heap/heap_bits.rs
```diff
@@ -54,33 +82,105 @@ pub struct WorkQueues {
 
 impl HeapBits {
     pub fn new(heap: &Heap) -> Self {
+        let mut modules = Vec::with_capacity(heap.modules.len());
```
sno2, line 85: If I understand correctly, we should be able to do this:

```rust
let modules = vec![false; heap.modules.len()];
```

instead of this:

```rust
let mut modules = Vec::with_capacity(heap.modules.len());
modules.resize(modules.capacity(), false);
```

for all of these.

nova_vm/src/heap/heap_gc.rs
```diff
@@ -165,220 +120,244 @@ pub fn heap_gc(heap: &mut Heap) {
...
-                marked.store(true, Ordering::Relaxed);
+                *marked = true;
                 let data = heap.symbols.get(index).unwrap().as_ref().unwrap();
                 if let Some(string_index) = data.descriptor {
                     queues.push_value(string_index.into());
                 }
             }
         });
-        let mut e_2_4: Box<[ElementIndex]> = queues.e_2_4.drain(..).collect();
+        let mut e_2_4: Box<[(ElementIndex, u32)]> = queues.e_2_4.drain(..).collect();
```
sno2, line 147: Macros over all of these element queues could make this code less ugly, but it isn't strictly necessary.

sno2 (approved):
I mostly looked for Rust-specific issues as I don't exactly understand all of the logic, but this looks good with the test 🚀

nova_vm/src/heap/heap_bits.rs
aapoalas: I knew there had to be a better way, I just couldn't remember anything :D

nova_vm/src/heap/heap_gc.rs
aapoalas: Yeah, I'm unfortunately really bad at macros so I don't really use them :D

sno2 (dismissed):

PR #85, opened 2024-01-22
feat(vm): Complete GlobalDeclarationInstantiation and partially handle EvaluateCall

aapoalas:
Fully complete the GlobalDeclarationInstantiation spec abstract operation used for script evaluation and create a partial handling of the EvaluateCall syntax directed operation / bytecode instruction.

### DIFF
--- a/nova_vm/src/ecmascript/types/language/value.rs
+++ b/nova_vm/src/ecmascript/types/language/value.rs
@@ -175,7 +175,7 @@ impl Value {
     }
 
     pub fn is_function(self) -> bool {
-        matches!(self, |Value::BoundFunction(_)| Value::BuiltinFunction(_)
+        matches!(self, Value::BoundFunction(_) | Value::BuiltinFunction(_)
             | Value::ECMAScriptFunction(_))
     }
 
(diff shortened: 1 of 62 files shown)


nova_vm/src/ecmascript/types/language/value.rs
```diff
@@ -181,6 +174,11 @@ impl Value {
         )
     }
 
+    pub fn is_function(self) -> bool {
+        matches!(self, |Value::BoundFunction(_)| Value::BuiltinFunction(_)
```
sno2, line 178: I have no idea how this leading pipe happened, but I don't think it's correct...
```suggestion
        matches!(self, Value::BoundFunction(_) | Value::BuiltinFunction(_)
```

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -154,10 +185,14 @@ pub(crate) struct JumpIndex {
     pub(crate) index: usize,
 }
 
-pub(crate) trait Compile {
+pub(crate) trait CompileEvaluation {
```
sno2, line 188: 👍

sno2 (dismissed):

PR #88, opened 2024-01-27
feat(ecmascript): Increase ArrayBuffer max size to usize

aapoalas:
I originally had used u32 to match what V8 did at the time but they've upped their max ArrayBuffer size and there's little reason why we wouldn't as well: This does grow the ArrayBuffer heap data to 150% of its previous size but that's life.


sno2 (changes requested):
Nice, I have applied a few Cargo autofixes related to redundant casting in `feat/array-buffer-max-size-fixups`: https://github.com/trynova/nova/commit/6c7c9a43b520ec13a66d0317aa28e007545cbf28

sno2 (approved):

PR #90, opened 2024-01-28
feat(ecmascript): Implement BuiltinFunction [[Call]] method

aapoalas:
Fairly simple stuff: Just piping up a few steps to get builtin calls working.

### DIFF
--- a/nova_vm/src/ecmascript/builtins/builtin_function.rs
+++ b/nova_vm/src/ecmascript/builtins/builtin_function.rs
@@ -4,11 +4,11 @@ use crate::{
     ecmascript::{
         execution::{agent::ExceptionType, Agent, ExecutionContext, JsResult, RealmIdentifier},
         types::{
-            BuiltinFunctionHeapData, Function, IntoFunction, IntoObject, IntoValue, Object,
-            PropertyDescriptor, String, Value,
+            BuiltinFunctionHeapData, Function, InternalMethods, IntoFunction, IntoObject,
+            IntoValue, Object, PropertyDescriptor, String, Value,
         },
     },
-    heap::{indexes::BuiltinFunctionIndex, GetHeapData},
+    heap::{indexes::BuiltinFunctionIndex, CreateHeapData, GetHeapData},
 };
 
 #[derive(Debug, Clone, Copy, Default)]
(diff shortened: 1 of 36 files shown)


nova_vm/src/ecmascript/builtins/builtin_function.rs
```diff
@@ -60,53 +81,196 @@ impl BuiltinFunctionArgs {
     }
 }
 
-/// ### [10.3.3 CreateBuiltinFunction ( behaviour, length, name, additionalInternalSlotsList \[ , realm \[ , prototype \[ , prefix \] \] \] )](https://tc39.es/ecma262/#sec-createbuiltinfunction)
+impl BuiltinFunctionIndex {
```
sno2, line 84: To be frank, I'm not a fan of placing these as methods on the indexes. It seems to go against your original ideas of having standalone functions for these operations.

nova_vm/src/ecmascript/builtins/builtin_function.rs
```diff
@@ -60,53 +81,196 @@ impl BuiltinFunctionArgs {
...
+            // 10. Perform SetFunctionLength(func, length).
+            length: args.length as u8,
+            // 8. Set func.[[Realm]] to realm.
+            realm,
+            object_index,
+        }));
 
     // 13. Return func.
-    Function::from(func)
+    BuiltinFunctionIndex::last(&agent.heap.builtin_functions)
```
sno2, line 273: Is there any way this could be a heap `create()` method?

nova_vm/src/ecmascript/builtins/builtin_function.rs
aapoalas: You're absolutely correct. I should've instead created a struct that encapsulates the index and implemented the InternalMethods trait on it. I didn't because ECMAScriptFunction didn't and that was probably done just out of pure laziness.

I'll fix this to the proper form.

nova_vm/src/ecmascript/builtins/builtin_function.rs
aapoalas: Yeah, should be just a matter of changing the call. Good point.

PR #94, opened 2024-02-07
feat(vm): Lexical declarations

aapoalas:
This PR adds support for lexical declarations, ie. `let` and `const`. At the same time the bytecode received the possibility of entering and exiting lexical scopes (lexical environments) as well as creating and initializing bindings in them. A smarter engine would probably precalculate these but we're not like the others! Ha! In your face!

### DIFF
--- a/nova_vm/src/ecmascript/syntax_directed_operations/scope_analysis.rs
+++ b/nova_vm/src/ecmascript/syntax_directed_operations/scope_analysis.rs
@@ -189,6 +189,7 @@ impl<'a> LexicallyDeclaredNames<'a> for LabeledStatement<'_> {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
 pub(crate) enum LexicallyScopedDeclaration<'a> {
     Variable(&'a VariableDeclarator<'a>),
     Function(&'a Function<'a>),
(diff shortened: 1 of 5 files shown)


nova_vm/src/ecmascript/syntax_directed_operations/scope_analysis.rs
```diff
@@ -1038,11 +1057,17 @@ impl<'a> TopLevelLexicallyDeclaredNames<'a> for Statement<'_> {
 /// The syntax-directed operation TopLevelLexicallyScopedDeclarations takes no
 /// arguments and returns a List of Parse Nodes.
 trait TopLevelLexicallyScopedDeclarations<'a> {
-    fn top_level_lexically_scoped_declarations<F: FnMut(&'a Declaration<'a>)>(&'a self, f: &mut F);
+    fn top_level_lexically_scoped_declarations<F: FnMut(LexicallyScopedDeclaration<'a>)>(
+        &'a self,
+        f: &mut F,
```
sno2, line 1062: Is this a double reference? Could just `f: impl FnMut(LexicallyScopedDeclaration<'a>)` work?

nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -338,6 +339,74 @@ impl Vm {
...
+                        .lexical_environment
+                        .get_outer_env(agent)
+                        .unwrap();
+                    agent
+                        .running_execution_context_mut()
+                        .ecmascript_code
+                        .as_mut()
+                        .unwrap()
+                        .lexical_environment = old_env;
+                }
```
sno2, line 389: Are all of these dupes required?

PR #99, opened 2024-02-10
feat: support getter logic, `SameValueNonNumber(String, String)`, `+`, `!`, and `~`

sno2:
Kind of messy- I was just going through a few of the TODOs.

### DIFF
--- a/nova_vm/src/ecmascript/types/language/string.rs
+++ b/nova_vm/src/ecmascript/types/language/string.rs
@@ -102,6 +102,24 @@ impl String {
         }
     }
 
+    /// If x and y have the same length and the same code units in the same
+    /// positions, return true; otherwise, return false.
+    pub fn have_same_length_and_code_units(agent: &mut Agent, x: String, y: String) -> bool {
+        match (x, y) {
+            (String::String(x), String::String(y)) => {
+                let x = agent.heap.get(x);
+                let y = agent.heap.get(y);
+                x == y
+            }
+            (String::String(x), String::SmallString(y))
+            | (String::SmallString(y), String::String(x)) => {
+                let x = agent.heap.get(x);
+                x.as_str() == Some(y.as_str())
+            }
+            (String::SmallString(x), String::SmallString(y)) => x == y,
+        }
+    }
+
     /// ### [6.1.4.1 StringIndexOf ( string, searchValue, fromIndex )](https://tc39.es/ecma262/#sec-stringindexof)
     pub fn index_of(self, agent: &mut Agent, search_value: Self, from_index: i64) -> i64 {
         // TODO: Figure out what we should do for invalid cases.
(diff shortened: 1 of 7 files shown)


nova_vm/src/ecmascript/types/language/string.rs
```diff
@@ -102,6 +102,24 @@ impl String {
         }
     }
 
+    /// If x and y have the same length and the same code units in the same
+    /// positions, return true; otherwise, return false.
+    pub fn have_same_length_and_code_units(agent: &mut Agent, x: String, y: String) -> bool {
```
aapoalas, line 107: **suggestion:** Should this be just  `eq` or `equals` or `string_equals`?

nova_vm/src/ecmascript/types/language/string.rs
```diff
@@ -102,6 +102,24 @@ impl String {
...
+        match (x, y) {
+            (String::String(x), String::String(y)) => {
+                let x = agent.heap.get(x);
+                let y = agent.heap.get(y);
+                x == y
+            }
+            (String::String(x), String::SmallString(y))
+            | (String::SmallString(y), String::String(x)) => {
```
aapoalas, line 115: **issue:** It should be impossible to enter this branch at present since a small string should never end up heap allocated.

nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -100,6 +100,26 @@ impl Vm {
...
+                    // 2. Let oldValue be ? ToNumeric(? GetValue(expr)).
+                    let old_value = to_numeric(agent, vm.result.take().unwrap())?;
+
+                    // 3. If oldValue is a Number, then
+                    if let Ok(old_value) = Number::try_from(old_value) {
+                        // a. Return Number::bitwiseNOT(oldValue).
+                        vm.result = Some(Number::bitwise_not(agent, old_value)?.into());
+                    }
+                    // 4. Else,
+                    else {
```
aapoalas, line 113: **nitpick:** This sort of comment-in-the-middle-of-if-else can be really confusing sometimes as it makes the `if` branch seem else-less. I've started just moving the `// 4. Else` comment inside the `else` branch, trading some spec language beauty for (hopefully) more readable code :)

nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -100,6 +100,26 @@ impl Vm {
...
+                        // a. Assert: oldValue is a BigInt.
+                        let Ok(old_value) = BigInt::try_from(old_value) else {
+                            unreachable!();
+                        };
+
+                        // b. Return BigInt::bitwiseNOT(oldValue).
+                        vm.result = Some(BigInt::bitwise_not(agent, old_value).into());
```
aapoalas, line 120: **nitpick:** Consider using `.into_value()` explicitly. Doesn't really matter but I believe that's what we've generally been doing in the `vm.result` assignments.

aapoalas (changes requested):
Kind of nitpicky but could make the `HeapString == SmallString` comparison branch explicitly `unreachable!()`: It's currently unreachable already but it's just that much better to be consistent about the unreachability of such concerns in my opinion.

aapoalas (approved):

PR #101, opened 2024-03-02
feat(vm): ArrayCreate

aapoalas:
Kind of a dirty implementation of `ArrayCreate`. The `array_set_length` and related abstract operations and functions are particularly dirty as they need to work around the `length` special case.

### DIFF
--- a/nova_vm/src/ecmascript/builtins/array.rs
+++ b/nova_vm/src/ecmascript/builtins/array.rs
@@ -2,21 +2,28 @@
 //!
 //! https://tc39.es/ecma262/#sec-array-exotic-objects
 
+pub(crate) mod abstract_operations;
 mod data;
 
 use std::ops::Deref;
 
-use super::{create_builtin_function, ArgumentsList, Behaviour, Builtin, BuiltinFunctionArgs};
+use super::{
+    array_set_length, create_builtin_function, ordinary::ordinary_define_own_property,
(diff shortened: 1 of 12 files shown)


nova_vm/src/ecmascript/builtins/array.rs
```diff
@@ -178,33 +211,118 @@ impl InternalMethods for Array {
...
+        agent: &mut Agent,
+        property_key: PropertyKey,
+        property_descriptor: PropertyDescriptor,
     ) -> JsResult<bool> {
-        todo!()
+        if property_key == PropertyKey::SmallString(SmallString::try_from("length").unwrap()) {
+            array_set_length(agent, self, property_descriptor)
+        } else if let PropertyKey::Integer(index) = property_key {
+            let index = index.into_i64();
+            if !(0 <= index && index <= (i64::pow(2, 32) - 2)) {
```
sno2, line 252: Just eye candy. Also, did you mean to exclude the maximum `u32` value? This range can be modified to `..=` to include all valid `u32` values.

```suggestion
            if !(0..u32::MAX as i64).contains(&index) {
```

sno2 (dismissed):
LGTM, nice

nova_vm/src/ecmascript/builtins/array.rs
aapoalas: Yeah I think I did mean that :) Just translating spec text without thinking about it properly

PR #106, opened 2024-04-03
feat(vm): Start building integration test capabilities

aapoalas:
This required quite a bit of work here and there:
1. Intrinsics needed to be set to the global object.
2. Errors needed to be somewhat properly supported.
3. Throwing of errors needed to be supported.

Catching errors is not supported, and a lot of the Object static and prototype methods are still unimplemented. But it's a start.


sno2 (approved):
LGTM, very cool 🎉

PR #179, opened 2024-05-13
feat(ecmascript): Add object property assignment

andreubotella:
This patch implements object property assignment (`a.b = c`).

Additionally, in order to make an assertion in `ordinary_set_with_own_descriptor` not fail, it also implements `PropertyStorage::has` on objects, as well as fixes a stack overflow in `ElementArray::has` caused by the fact that the element arrays were being stored on the stack as part of the computation.


aapoalas (approved):
LGTM! Very nice, thank you <3

PR #181, opened 2024-05-13
chore(vm): Refactor `Vm::execute` to allow for try-catch

andreubotella:
The current implementation of `Vm::execute` uses the `?` operator to propagate JS exceptions to the caller. This currently only works because we don't implement try-catch, since `?` does not allow catching an `Err` result value within the current function.

This patch adds an `execute_instruction` method, which handles the execution of a single instruction, which lets the `execute` method handle its non-successful result value.

However, this would not work with `return` statements, since they are implemented by returning a value from `execute`. To fix this, we add a `ContinuationKind` enum, which we use as the successful return type of `execute_instruction`, to be able to tell apart what the spec calls normal vs return completions.

### DIFF
--- a/nova_vm/src/engine/bytecode/vm.rs
+++ b/nova_vm/src/engine/bytecode/vm.rs
@@ -24,7 +24,17 @@ use crate::ecmascript::{
     },
 };
 
-use super::{Executable, Instruction, InstructionIter};
+use super::{instructions::Instr, Executable, Instruction, InstructionIter};
+
+/// Indicates how the execution of an instruction should affect the remainder of
+/// execution that contains it.
+#[must_use]
+enum ContinuationKind {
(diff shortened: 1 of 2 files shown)


nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -24,7 +24,14 @@ use crate::ecmascript::{
     },
 };
 
-use super::{Executable, Instruction, InstructionIter};
+use super::{instructions::Instr, Executable, Instruction, InstructionIter};
+
+enum NonThrowCompletion {
```
aapoalas, line 29: **suggestion:** Could mark this as `must_use`.

aapoalas (approved):
LGTM, great idea and execution! <3

PR #182, opened 2024-05-13
feat(vm): Implement try-catch

andreubotella:
Do not merge before #181.

### DIFF
--- a/nova_vm/src/engine/bytecode/executable.rs
+++ b/nova_vm/src/engine/bytecode/executable.rs
@@ -1159,6 +1159,44 @@ impl CompileEvaluation for ast::ThrowStatement<'_> {
     }
 }
 
+impl CompileEvaluation for ast::TryStatement<'_> {
+    fn compile(&self, ctx: &mut CompileContext) {
+        if self.finalizer.is_some() {
+            todo!();
+        }
+
+        let jump_to_catch = ctx
+            .exe
+            .add_instruction_with_jump_slot(Instruction::PushExceptionJumpTarget);
+        self.block.compile(ctx);
+        ctx.exe.add_instruction(Instruction::PopExceptionJumpTarget);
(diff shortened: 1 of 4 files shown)


nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -1159,6 +1159,44 @@ impl CompileEvaluation for ast::ThrowStatement<'_> {
...
+        let jump_to_catch = ctx
+            .exe
+            .add_instruction_with_jump_slot(Instruction::PushExceptionJumpTarget);
+        self.block.compile(ctx);
+        ctx.exe.add_instruction(Instruction::PopExceptionJumpTarget);
+        let jump_to_end = ctx.exe.add_instruction_with_jump_slot(Instruction::Jump);
+
+        let catch_clause = self.handler.as_ref().unwrap();
+        ctx.exe.set_jump_target_here(jump_to_catch);
+        if let Some(exception_param) = &catch_clause.param {
```
aapoalas, line 1177: **question:** Any idea how we should handle exiting declarative environments caused by throwing?

eg.
```js
try {
  if (foo) { // <-- this env needs to be popped off
    let a = null;
    null.maybeItWorks();
  }
} catch(e) {
}
```

aapoalas (approved):
Absolutely splendid! I had one question about how declaration environment popping should work with throws.

nova_vm/src/engine/bytecode/executable.rs
andreubotella: I hadn't taken that into account. I guess exception jump targets could store their declarative environment, and that would be restored when an exception is thrown.

aapoalas: Oh yeah, that would work and make a lot of sense. So `PushExceptionJumpTarget` and `PopExceptionJumpTarget` should push/pop the jump target together with the parent environment to restore when an exception is thrown.

PR #185, opened 2024-05-16
feat(ecmascript): Implement logical expressions and nullish coalescing

andreubotella:


nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -561,6 +561,46 @@ impl CompileEvaluation for ast::BinaryExpression<'_> {
...
+        match self.operator {
+            oxc_syntax::operator::LogicalOperator::Or => {
+                ctx.exe.add_instruction(Instruction::LogicalNot);
+            }
+            oxc_syntax::operator::LogicalOperator::And => {}
+            oxc_syntax::operator::LogicalOperator::Coalesce => {
+                ctx.exe.add_instruction(Instruction::IsNullOrUndefined);
+            }
+        }
+        let jump_to_return_left = ctx
```
aapoalas, line 583: **thought:** I wonder if it would be worthwhile to add a `JumpIfNotWithStore` kind of instruction...

nova_vm/src/engine/bytecode/instructions.rs
```diff
@@ -75,6 +75,9 @@ pub enum Instruction {
     IsLooselyEqual,
     /// Store IsStrictlyEqual() as the result value.
     IsStrictlyEqual,
+    /// Store true as the result value if the current result value is null or
+    /// undefined, false otherwise.
+    IsNullOrUndefined,
```
aapoalas, line 80: <3

aapoalas (approved):
Veery nice! Thank you <3

nova_vm/src/engine/bytecode/executable.rs
andreubotella: I think it might be better to have a version of `JumpIfNot` that switches on `vm.result` without taking it, since the only reason we need to duplicate the result into the stack is because the result gets lost after the jump.

aapoalas: Wouldn't we also need to restore it from the stack in the `Not` and `IsNullOrUndefined` cases, to undo their work?

andreubotella: Oh, that's right.

In any case, I think as long as the result is independent from the stack, there will have to be a way to duplicate the result into the stack and to load the top of the stack into the result without popping. In that sense, `LoadCopy` seems a lot more general than `JumpIfNotWithStore`.

aapoalas: Yeah, I like `LoadCopy` a lot, it's a great addition/idea by you.

PR #186, opened 2024-05-16
feat(ecmascript): Implement comparison and `in` operators

andreubotella:
Previously the only supported comparison operators were `<`, `===` and `!==`. This patch adds support for the rest of comparison operators (namely `<=`, `>`, `>=`, `==` and `!=`), as well as the `in` operator, and also unifies bytecode generation for all binary operators.

### DIFF
--- a/nova_vm/src/engine/bytecode/executable.rs
+++ b/nova_vm/src/engine/bytecode/executable.rs
@@ -471,69 +471,6 @@ impl CompileEvaluation for ast::UnaryExpression<'_> {
 
 impl CompileEvaluation for ast::BinaryExpression<'_> {
     fn compile(&self, ctx: &mut CompileContext) {
-        match self.operator {
-            BinaryOperator::LessThan => {
-                // 13.10.1 Runtime Semantics: Evaluation
-                // RelationalExpression : RelationalExpression < ShiftExpression
-
-                // 1. Let lref be ? Evaluation of RelationalExpression.
-                self.left.compile(ctx);
-
-                // 2. Let lval be ? GetValue(lref).
-                if is_reference(&self.left) {
-                    ctx.exe.add_instruction(Instruction::GetValue);
-                }
-                ctx.exe.add_instruction(Instruction::Load);
-
-                // 3. Let rref be ? Evaluation of ShiftExpression.
-                self.right.compile(ctx);
-
-                // 4. Let rval be ? GetValue(rref).
-                if is_reference(&self.right) {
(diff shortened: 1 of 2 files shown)


nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -551,13 +488,47 @@ impl CompileEvaluation for ast::BinaryExpression<'_> {
...
 
-        ctx.exe.add_instruction(Instruction::Load);
-
-        // 5. Return ? ApplyStringOrNumericBinaryOperator(lval, opText, rval).
-        ctx.exe
-            .add_instruction(Instruction::ApplyStringOrNumericBinaryOperator(
-                self.operator,
-            ));
+        match self.operator {
+            BinaryOperator::LessThan => {
```
aapoalas, line 492: **praise:** Oh yeah <3 I'd been looking at and confused by this code for a good while.

nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -418,17 +420,66 @@ impl Vm {
...
+                    Some(true) => true,
+                    _ => false,
+                };
+                vm.result = Some(result.into());
+            }
+            Instruction::LessThanEquals => {
+                let lval = vm.stack.pop().unwrap();
+                let rval = vm.result.take().unwrap();
+                let result = match is_less_than::<false>(agent, rval, lval)? {
+                    Some(true) | None => false,
```
aapoalas: **nitpick:** This could be done in the same vein as above: `Some(false) => true, _ => false`. Though admittedly, that can also be kind of confusing to read.

Actually, isn't it all just doable as `call(agent, lval, rval)? == Some(cond)`?

nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -418,17 +420,55 @@ impl Vm {
             Instruction::LessThan => {
                 let lval = vm.stack.pop().unwrap();
                 let rval = vm.result.take().unwrap();
-                let result = is_less_than::<true>(agent, lval, rval)
-                    .unwrap()
-                    .unwrap_or_default();
+                let result = matches!(is_less_than::<true>(agent, lval, rval)?, Some(true));
```
aapoalas, line 423: **nitpick:** These don't require `matches!()` but can instead be just `call(agent, lval, rval) == Some(true)` or or `!= Some(true)`.

aapoalas (approved):
Absolutely splendid! Had one nit.

aapoalas (approved):

PR #187, opened 2024-05-16
feat(ecmascript): String concatenation

andreubotella:


nova_vm/src/ecmascript/types/language/string.rs
```diff
@@ -161,6 +161,78 @@ impl String {
...
+                        std::string::String::with_capacity(agent[*idx].len() + string.len(agent));
+                    result.push_str(agent[*idx].as_str());
+                    result.push_str(string.as_str(agent));
+                    status = Status::String(result)
+                }
+                Status::SmallString { data, len } => {
+                    let string_len = string.len(agent);
+                    if *len + string_len <= 7 {
+                        let String::SmallString(smstr) = string else {
+                            unreachable!()
```
aapoalas, line 205: **issue:** I think this is currently reachable with `"foo" + "\0"`.

aapoalas (changes requested):
I think the small string assumption is wrong. That being said, we could also just leave it as a `todo!()` for now and come back to it later.

aapoalas (approved):
LGTM <3

PR #189, opened 2024-05-16
feat(ecmascript): Property access and assignment on ECMAScript functions

andreubotella:

PR #190, opened 2024-05-16
feat(ecmascript): `name` and `prototype` properties of ECMAScript functions

andreubotella:
This is a precondition for using ECMAScript functions as constructors.

Do not merge before #189.

### DIFF
--- a/nova_vm/src/ecmascript/syntax_directed_operations/function_definitions.rs
+++ b/nova_vm/src/ecmascript/syntax_directed_operations/function_definitions.rs
@@ -9,7 +9,7 @@ use crate::{
             Agent, ECMAScriptCodeEvaluationState, EnvironmentIndex, JsResult,
             PrivateEnvironmentIndex,
         },
-        types::{Function, PropertyKey, String, Value},
+        types::{Function, PropertyKey, String, Value, BUILTIN_STRING_MEMORY},
     },
     engine::{Executable, FunctionExpression, Vm},
 };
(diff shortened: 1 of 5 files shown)


nova_vm/src/ecmascript/syntax_directed_operations/function_definitions.rs
```diff
@@ -28,7 +29,7 @@ pub(crate) fn instantiate_ordinary_function_object(
     // FunctionDeclaration : function BindingIdentifier ( FormalParameters ) { FunctionBody }
     if let Some(id) = &function.id {
         // 1. Let name be StringValue of BindingIdentifier.
-        let _name = id.name.clone();
+        let name = id.name.clone();
```
aapoalas, line 32: **nitpick:** `clone()` here is no longer relevant. (This used to be an owned string, now it's a reference.)

nova_vm/src/ecmascript/syntax_directed_operations/function_definitions.rs
```diff
@@ -62,13 +65,15 @@ pub(crate) fn instantiate_ordinary_function_object(
             env,
             private_env,
         };
+        let f = ordinary_function_create(agent, params);
 
         // 3. Perform SetFunctionName(F, "default").
-        // set_function_name(f, "default");
+        let pk_name = PropertyKey::from_static_str(agent, "default");
```
aapoalas, line 71: **issue:** We have `"default"` already available as a string in `BUILTIN_STRING_MEMORY`. We should use that instead of trying to reallocate the string.

aapoalas (changes requested):
Niiiice; had one nit and one issue.

aapoalas (approved):

PR #191, opened 2024-05-17
feat(ecmascript): Support getters and setters in `PropertyStorage`

andreubotella:
Although currently JS user code can't define getters and setters, there is one usage in internal Nova code in the creation of the `arguments` object for non-lexical-this functions. This change makes it possible to call such functions without panicking.

### DIFF
--- a/nova_vm/src/ecmascript/builtins/ordinary.rs
+++ b/nova_vm/src/ecmascript/builtins/ordinary.rs
@@ -435,7 +435,7 @@ fn validate_and_apply_property_descriptor(
             }
         }
         // e. Else if current.[[Writable]] is false, then
-        else if let Some(false) = current.writable {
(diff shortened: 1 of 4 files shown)


nova_vm/src/ecmascript/builtins/ordinary.rs
```diff
@@ -442,7 +435,7 @@ fn validate_and_apply_property_descriptor(
             }
         }
         // e. Else if current.[[Writable]] is false, then
-        else if let Some(true) = current.writable {
+        else if let Some(false) = current.writable {
```
aapoalas, line 438: **nitpick:** `if current.writable == Some(false)`.

aapoalas (approved):
Awesome! <3

PR #193, opened 2024-05-18
fix(ecmascript): Make sure built-in functions have `name` and `length`

andreubotella:
When built-in functions are created, the `initial_name` and `length` fields of `BuiltinFunctionHeapData` are populated. This means that, if there is no backing object when the function is created, the `name` and `length` properties will have the value given by those fields. This is the case even if a backing object needs to be created later on, since those fields will be added to the backing object.

However, if properties are added to the built-in function as it is being constructed, then there will be a backing object without the `name` and `length` properties, so they will appear to not be set. This is always the case for built-in constructor functions, since they will always have a `prototype` property.

The long-term fix for this would be to avoid duplicating heap data fields in the backing object somehow. However, in the meantime, this patch makes sure that all built-in functions that are created with a backing object have the `name` and `length` properties in it.

### DIFF
--- a/nova_vm/src/ecmascript/builders/builtin_function_builder.rs
+++ b/nova_vm/src/ecmascript/builders/builtin_function_builder.rs
@@ -274,16 +274,27 @@ impl<'agent, P, L, B, Pr> BuiltinFunctionBuilder<'agent, P, L, CreatorName, B, P
     }
 }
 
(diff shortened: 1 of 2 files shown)


aapoalas (approved):
Nice fix, thank you <3

PR #195, opened 2024-05-19
feat(ecmascript): Allow object creation with a different prototype

andreubotella:
Per the spec, OrdinaryCreateFromConstructor should get the `prototype` property of the passed constructor, and if that is an object, it should create a new object whose prototype is that property; otherwise, the prototype would be set to the passed intrinsic default prototype.

In the current implementation of this method, the `prototype` of the passed constructor is obtained, but then it is ignored. The given intrinsic default prototype is used to determine in which heap to store the object, which gives it the right default prototype, but it does not check whether the constructor's `prototype` property matches that default prototype.

It seems like this gets the right result for all of the times where Nova currently does object construction. However, allowing creating an object with a prototype different from the heap's default would be needed for constructing ECMAScript functions.

As part of that, this patch changes `get_prototype_from_constructor` to return a `JsResult<Option<Object>>`, guaranteeing that it will not give the caller an object which is the `ProtoIntrinsics`'s corresponding prototype. It also changes `ordinary_object_create_with_intrinsics` to take an optional prototype object aside from the optional `ProtoIntrinsics`. 

It also changes `ordinary_create_from_constructor` to use the prototype obtained from the constructor, as well as to remove the `_internal_slots_list` parameter. In the spec this parameter is used to create the object with the right internal slots, but in Nova the equivalent of this is choosing in which heap the object is created, which the `ProtoIntrinsics` argument already does.

### DIFF
--- a/nova_vm/src/ecmascript/builtins/ordinary.rs
+++ b/nova_vm/src/ecmascript/builtins/ordinary.rs
@@ -815,15 +815,24 @@ pub(crate) fn ordinary_own_property_keys(_agent: &Agent, _object: Object) -> Vec
 /// called by any algorithm that subsequently modifies the internal methods of
 /// the object in ways that would make the result non-ordinary. Operations that
 /// create exotic objects invoke MakeBasicObject directly.
+///
+/// NOTE: In this implementation, `proto_intrinsics` determines the heap in
+/// which the object is placed, and therefore its heap data type and its
+/// internal slots. If `prototype` is None, the object will be created with the
+/// default prototype from the intrinsics, otherwise with the given prototype.
+/// To create an object with null prototype, both `proto_intrinsics` and
+/// `prototype` must be None.
 pub(crate) fn ordinary_object_create_with_intrinsics(
     agent: &mut Agent,
-    prototype: Option<ProtoIntrinsics>,
+    proto_intrinsics: Option<ProtoIntrinsics>,
+    prototype: Option<Object>,
 ) -> Object {
-    let Some(prototype) = prototype else {
+    let Some(proto_intrinsics) = proto_intrinsics else {
+        assert!(prototype.is_none());
         return agent.heap.create_null_object(&[]).into();
     };
 
-    match prototype {
+    let object = match proto_intrinsics {
         ProtoIntrinsics::Array => agent.heap.create(ArrayHeapData::default()).into(),
         ProtoIntrinsics::ArrayBuffer => agent.heap.create(ArrayBufferHeapData::default()).into(),
         ProtoIntrinsics::BigInt => agent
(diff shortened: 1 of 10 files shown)


nova_vm/src/ecmascript/builtins/ordinary.rs
```diff
@@ -957,7 +966,13 @@ pub(crate) fn ordinary_object_create_with_intrinsics(
         ProtoIntrinsics::WeakMap => agent.heap.create(WeakMapHeapData::default()).into_object(),
         ProtoIntrinsics::WeakRef => agent.heap.create(WeakRefHeapData::default()).into_object(),
         ProtoIntrinsics::WeakSet => agent.heap.create(WeakSetHeapData::default()).into_object(),
+    };
+
+    if let Some(prototype) = prototype {
+        object.internal_set_prototype(agent, Some(prototype));
```
aapoalas, line 972: **praise:** This is such a nice and clean solution, great idea! I'd been thinking that this function would maybe eventually somehow pass the `prototype` to all of the individual heap data constructors and they'd get to decide if they create the backing object immediately or not. But that was a huge hassle which is why I avoided doing it for the time being.

This, however <3 A great, greeeeat idea.

nova_vm/src/ecmascript/builtins/ordinary.rs
```diff
@@ -972,24 +987,28 @@ pub(crate) fn ordinary_object_create_with_intrinsics(
 /// \[\[Prototype]]. internalSlotsList contains the names of additional
 /// internal slots that must be defined as part of the object. If
 /// internalSlotsList is not provided, a new empty List is used.
+///
+/// NOTE: In this implementation, `intrinsic_default_proto` also defines which
```
aapoalas, line 991: **praise:** These notes are a great idea as well.

nova_vm/src/ecmascript/builtins/ordinary.rs
```diff
@@ -1003,188 +1022,73 @@ pub(crate) fn ordinary_create_from_constructor(
...
-            ProtoIntrinsics::WeakRef => (
-                intrinsics.weak_ref().into_function(),
-                intrinsics.weak_ref_prototype().into_object(),
-            ),
-            ProtoIntrinsics::WeakSet => (
-                intrinsics.weak_set().into_function(),
-                intrinsics.weak_set_prototype().into_object(),
-            ),
+        let intrinsic_constructor = match intrinsic_default_proto {
+            ProtoIntrinsics::AggregateError => intrinsics.aggregate_error().into_function(),
```
aapoalas, line 1041: **praise:** Loads better this way <3

aapoalas (approved):
Absolutely awesome <3 This makes me wax philosophical about collaboration and cooperation: A different pair of eyes sees such simple and great solutions to issues or pain points that another pair of eyes has learned to actively avoid thinking about.

PR #196, opened 2024-05-19
feat(ecmascript): Allow constructing ECMAScript functions

andreubotella:
As part of making this work, this patch merges the internal slots `[[ConstructorKind]]` and `[[IsClassConstructor]]`, as well as the presence or absence of the `[[Construct]]` internal method (which hadn't been implemented in Nova yet) into a single `ConstructorStatus` enum.

### DIFF
--- a/nova_vm/src/ecmascript/builtins/ecmascript_function.rs
+++ b/nova_vm/src/ecmascript/builtins/ecmascript_function.rs
@@ -534,7 +534,6 @@ impl InternalMethods for ECMAScriptFunction {
         //   a. If result.[[Value]] is an Object, return result.[[Value]].
         if let Ok(value) = Object::try_from(value) {
             return Ok(value);
-            return Ok(Object::try_from(value).unwrap());
         }
         //   b. If kind is base, return thisArgument.
         if is_base {
@@ -838,7 +837,7 @@ pub(crate) fn make_constructor(
         Function::ECMAScriptFunction(idx) => {
             let data = &mut agent[idx];
             // a. Assert: IsConstructor(F) is false.
-            assert!(!data.ecmascript_function.constructor_status.is_constructor());
+            debug_assert!(!data.ecmascript_function.constructor_status.is_constructor());
             // b. Assert: F is an extensible object that does not have a "prototype" own property.
             // TODO: Handle Some() object indexes?
             assert!(data.object_index.is_none());
--- a/nova_vm/src/ecmascript/builtins/ecmascript_function.rs
+++ b/nova_vm/src/ecmascript/builtins/ecmascript_function.rs
@@ -43,7 +43,8 @@ use crate::{
 };
 
 use super::{
-    create_unmapped_arguments_object, ordinary::ordinary_object_create_with_intrinsics,
+    create_unmapped_arguments_object,
+    ordinary::{ordinary_create_from_constructor, ordinary_object_create_with_intrinsics},
     ArgumentsList,
 };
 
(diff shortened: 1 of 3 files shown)


nova_vm/src/ecmascript/builtins/ecmascript_function.rs
```diff
@@ -454,11 +474,76 @@ impl InternalMethods for ECMAScriptFunction {
...
+        let result = ordinary_call_evaluate_body(agent, self, arguments_list);
+        // 9. Remove calleeContext from the execution context stack and restore callerContext as the running execution context.
+        agent.execution_context_stack.pop();
+        // 10. If result is a return completion, then
+        // 11. Else,
+        //   a. ReturnIfAbrupt(result).
+        let value = result?;
+        // 10. If result is a return completion, then
+        //   a. If result.[[Value]] is an Object, return result.[[Value]].
+        if value.is_object() {
```
aapoalas, line 535:
```suggestion
        if let Ok(value) = Object::try_from(value) {
            return Ok(value);
```

nova_vm/src/ecmascript/builtins/ecmascript_function.rs
```diff
@@ -454,11 +474,76 @@ impl InternalMethods for ECMAScriptFunction {
...
+            return Ok(Object::try_from(value).unwrap());
+        }
+        //   b. If kind is base, return thisArgument.
+        if is_base {
+            return Ok(this_argument.unwrap());
+        }
+        todo!("Derived classes");
```
aapoalas, line 542: **nitpick:** I guess all the steps here can already be implemented.

nova_vm/src/ecmascript/builtins/ecmascript_function.rs
```diff
@@ -98,10 +99,27 @@ impl From<ECMAScriptFunction> for Function {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
-pub enum ConstructorKind {
-    Base,
-    Derived,
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConstructorStatus {
```
aapoalas, line 103: **praise:** Great idea <3

nova_vm/src/ecmascript/builtins/ecmascript_function.rs
```diff
@@ -751,15 +835,15 @@ pub(crate) fn make_constructor(
         Function::BoundFunction(_) => unreachable!(),
         // 1. If F is an ECMAScript function object, then
         Function::ECMAScriptFunction(idx) => {
-            // a. Assert: IsConstructor(F) is false.
-            // TODO: How do we separate constructors and non-constructors?
             let data = &mut agent[idx];
+            // a. Assert: IsConstructor(F) is false.
+            assert!(!data.ecmascript_function.constructor_status.is_constructor());
```
aapoalas, line 840: **nitpick:** This can probably be a debug assert.

aapoalas (dismissed):
Awesome <3

nova_vm/src/ecmascript/builtins/ecmascript_function.rs
andreubotella: Not yet. I think for derived classes, `constructor_env` would be a `FunctionEnvironment`, and `get_this_binding()` wasn't implemented yet for those. I do implement it in #197 though.

nova_vm/src/ecmascript/builtins/ecmascript_function.rs
andreubotella: Done.

PR #197, opened 2024-05-19
feat(ecmascript): Implement the `this` keyword

andreubotella:
This patch also fixes a bug in which functions created with the `InstantiateOrdinaryFunctionExpression` bytecode instruction were created with a lexical this mode, when it should be either strict or global. The lexical this mode is only used for arrow functions, and it causes the `this` resolution to be wrong in other functions.

Do not merge before #196.

### DIFF
--- a/nova_vm/src/ecmascript/execution/environments.rs
+++ b/nova_vm/src/ecmascript/execution/environments.rs
@@ -556,7 +556,7 @@ impl Environments {
     }
 }
 
-/// ### {9.4.3 GetThisEnvironment ( )}(https://tc39.es/ecma262/#sec-getthisenvironment)
+/// ### [9.4.3 GetThisEnvironment ( )](https://tc39.es/ecma262/#sec-getthisenvironment)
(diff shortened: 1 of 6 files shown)


nova_vm/src/ecmascript/execution/environments.rs
```diff
@@ -555,3 +555,29 @@ impl Environments {
             .expect("ObjectEnvironmentIndex pointed to a None")
     }
 }
+
+/// ### {9.4.3 GetThisEnvironment ( )}(https://tc39.es/ecma262/#sec-getthisenvironment)
```
aapoalas, line 559: **nitpick:** This should be `[]` brackets instead of braces, right?

aapoalas (approved):
LGTM <3

nova_vm/src/ecmascript/execution/environments.rs
andreubotella: Fixed.

PR #200, opened 2024-05-27
feat(nova_cli): Add a `print` global function

andreubotella:
In order to implement this, this patch also publicly exposes `Realm` (since using that type is needed to call `initialize_host_defined_realm`), and it also adds a public `Value::to_string` method.


aapoalas (approved):
LGTM! Nice stuff, first embedding API usage :)

I have a feeling we might want to get rid of the Realm exposing in these functions somehow, if at all possible. But that's for some other day.

PR #202, opened 2024-05-28
feat(nova_vm): Add symbol descriptive strings

andreubotella:


nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -62,7 +63,10 @@ impl StringConstructor {
             // 2. Else,
             // a. If NewTarget is undefined and value is a Symbol, return SymbolDescriptiveString(value).
             if new_target.is_none() && value.is_symbol() {
```
aapoalas, line 65: **suggestion:** This could be:
```rs
if let Some(Value::Symbol(value)) = value {
  return Ok(Symbol::from(value).descriptive_string(agent).into_value());
}
```

And actually, it might not be a bad idea to change `Value::Symbol(SymbolIndex)` into a `Value::Symbol(Symbol)`. I started thinking yesterday that that's probably going to be what we should do.

nova_vm/src/ecmascript/types/language/symbol.rs
```diff
@@ -113,3 +113,22 @@ impl IndexMut<Symbol> for Heap {
             .expect("Symbol slot empty")
     }
 }
+
+impl Symbol {
+    /// [20.4.3.3.1 SymbolDescriptiveString ( sym )](https://tc39.es/ecma262/#sec-symboldescriptivestring)
```
aapoalas, line 118: **nitpick:** Missing `###` to make this a title.

aapoalas (dismissed):
LGTM <3 Had one suggestion and one nitpick.

PR #203, opened 2024-05-28
feat(nova_cli): Print uncaught exceptions

andreubotella:
This patch prints a string representation of an uncaught exception in `nova_cli eval` before exiting the process with error code 1.

Since the `ToString()` operation is not guaranteed to succeed, this instead adds a `Value::string_repr` which never returns `Err`, and which accounts for symbols as well as objects whose stringification throws.

Do not merge before #202.

### DIFF
--- a/nova_cli/src/main.rs
+++ b/nova_cli/src/main.rs
@@ -80,7 +80,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 Err(error) => {
                     eprintln!(
                         "Uncaught exception: {}",
-                        error.value().string_repr(&mut agent).as_str(&mut agent)
+                        error.value().string_repr(&mut agent).as_str(&agent)
                     );
                     std::process::exit(1);
                 }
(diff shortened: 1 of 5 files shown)


nova_cli/src/main.rs
```diff
@@ -71,10 +71,19 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            if verbose {
-                println!("{:?}", result);
+            match script_evaluation(&mut agent, script) {
+                Ok(result) => {
+                    if verbose {
+                        println!("{:?}", result);
+                    }
+                }
+                Err(error) => {
+                    eprintln!(
```
aapoalas, line 81: **thought:** I guess eventually this should go into some host hook, maybe?

aapoalas (approved):
LGTM

nova_cli/src/main.rs
aapoalas: Huh, okay. I hadn't actually realised there's a split between how V8 does tasks and how the spec defines them.

aapoalas (approved):

PR #229, opened 2024-06-15
feat(tests): Add test262 test runner

andreubotella:
This patch adds a test runner for test262. Since the runner will run multiple tests, it isn't added as a cargo test package, but instead as a separate crate outputting a `test262` binary, which expects to be always run from a Nova checkout.

This patch also updates the CI configuration to run the test262 tests and detect cases where the expectations no longer match.


aapoalas (approved):
I don't have the time to review right now but since this adds a ton of tests and "successfully" runs them, I'm very happy with this! :)

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+    Pass,
+    Fail,
+    Crash,
+    Timeout,
+}
+
+/// These directory names are filtered out by default.
+const SKIP_DIRS: &[&str] = &[
+    "annexB", "intl402", "staging",
+    //
```
aapoalas, line 27: **nitpick:** Orphan comment line.

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+    //"built-ins",
+    //"language",
+];
+
+fn is_test_file(file_name: &str) -> bool {
+    file_name.ends_with(".js") && !file_name.contains("_FIXTURE")
```
aapoalas, line 33: **question:** What're these `_FIXTURE` files? Consider maybe adding a comment about them here.

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+#[derive(Debug)]
+struct Test262Runner {
+    runner_base_path: PathBuf,
+    tests_base: PathBuf,
+    nova_harness_path: PathBuf,
+    nova_cli_path: PathBuf,
+    expectations: HashMap<PathBuf, TestExpectation>,
+    treat_crashes_as_failures: bool,
+
+    num_tests_run: usize,
```
aapoalas, line 45: **suggestion:** Should we maybe split the above configuration parts into their own struct and keep these mutable parts in a separate part? There might be some magical benefit to that :)

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+            if entry.file_type().unwrap().is_file()
+                && Self::filter_file(filters, &entry.file_name())
+            {
+                self.run_test(&entry.path());
+            }
+        }
+    }
+
+    fn filter_dir(filters: &[PathBuf], os_file_name: &OsStr) -> Option<Box<[PathBuf]>> {
+        if SKIP_DIRS.contains(&os_file_name.to_str().unwrap()) {
```
aapoalas, line 85: **suggestion:** Maybe add a few minor inline comments here, eg.

> Always skip SKIP_DIRS directories

and

> Filters being empty means that all directories should be visited.

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+
+            if entry.file_type().unwrap().is_file()
+                && Self::filter_file(filters, &entry.file_name())
+            {
+                self.run_test(&entry.path());
+            }
+        }
+    }
+
+    fn filter_dir(filters: &[PathBuf], os_file_name: &OsStr) -> Option<Box<[PathBuf]>> {
```
aapoalas, line 84: **nitpick:** `os_folder_name` maybe?

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+        if SKIP_DIRS.contains(&os_file_name.to_str().unwrap()) {
+            return None;
+        }
+        if filters.is_empty() {
+            return Some(vec![].into_boxed_slice());
+        }
+
+        let mut child_filters = Vec::with_capacity(filters.len());
+        for filter in filters {
+            if filter == OsStr::new("") || filter == os_file_name {
```
aapoalas, line 94: **question:** What is the purpose of this if check case and the early return? Is this some error condition? Maybe add a small comment here.

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+            return false;
+        }
+        if filters.is_empty() {
+            return true;
+        }
+        filters.iter().any(|filter| filter == os_file_name)
+    }
+
+    fn run_test(&mut self, path: &PathBuf) {
+        //println!("Running test {:?}", path);
```
aapoalas, line 120: **nitpick:** Leftover debug code.

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+                .chain(metadata.includes);
+            for include_relpath in includes_iter {
+                let mut include = harness.clone();
+                include.push(include_relpath);
+                command.arg(include);
+            }
+        }
+        command.arg(path);
+
+        //println!("{:?}", command);
```
aapoalas, line 151: **nitpick:** Here too.

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+        let relpath = path.strip_prefix(&self.tests_base).unwrap();
+        let expectation = self
+            .expectations
+            .get(relpath)
+            .copied()
+            .unwrap_or(TestExpectation::Pass);
+
+        if test_result != expectation {
+            if self.treat_crashes_as_failures
+                && matches!(test_result, TestExpectation::Fail | TestExpectation::Crash)
```
aapoalas, line 164: **suggestion:** This if case could also use a comment to explain what we're doing here. (Oh, I think I got it now as I was writing this comment.)

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+                && matches!(expectation, TestExpectation::Fail | TestExpectation::Crash)
+            {
+                return;
+            }
+            self.unexpected_results
+                .insert(relpath.to_path_buf(), test_result);
+        }
+    }
+
+    fn handle_test_output(
```
aapoalas, line 174: **nitpick:** This is actually doing the real running of the command, and could reflect that in its name. Not at all important though.

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+
+        let mut bytes = vec![];
+        let mut buffer = [0u8; 1024];
+
+        // Consume until the start of the YAML declaration
+        let mut matcher = BytesMatcher::new(YAML_START);
+        loop {
+            let read_bytes = match reader.read(&mut buffer) {
+                Ok(read_bytes) => read_bytes,
+                Err(error) if error.kind() == ErrorKind::Interrupted => continue,
```
aapoalas, line 396: **question:** What is this continue error condition?

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+                    read_bytes, 0,
+                    "Expected YAML declaration end before the end of the file"
+                );
+
+                let read_slice = &buffer[..read_bytes];
+                if let Some(idx) = matcher.run(read_slice) {
+                    if idx > YAML_END.len() {
+                        bytes.extend_from_slice(&read_slice[..idx - YAML_END.len()]);
+                    } else {
+                        bytes.truncate(bytes.len() + idx - YAML_END.len());
```
aapoalas, line 436: **suggestion:** Leave a comment about what this condition is.

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+}
+
+#[derive(Debug, ClapParser)]
+#[command(name = "test262")]
+#[command(about = "A test262 runner for Nova.", long_about = None)]
+struct Cli {
+    #[arg(short, long)]
+    update_expectations: bool,
+    #[arg(long = "dont-treat-crashes-as-failures", action = clap::ArgAction::SetFalse)]
+    treat_crashes_as_failures: bool,
```
aapoalas, line 456: **suggestion:** Add a comment about what this flag is meant for.

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+        runner_base_path,
+        tests_base,
+        nova_harness_path,
+        nova_cli_path,
+        treat_crashes_as_failures: cli.treat_crashes_as_failures,
+        expectations,
+        num_tests_run: 0,
+        unexpected_results: Default::default(),
+    };
+    if filters_are_valid {
```
aapoalas, line 535: **suggestion:** If filters are invalid, we should maybe panic with an error message?

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+        runner.run(&cli.filters);
+    }
+
+    //println!();
```
aapoalas, line 539: **nitpick:** Commented out code.

tests/test262_runner.rs
```diff
@@ -0,0 +1,589 @@
...
+            if result == TestExpectation::Pass {
+                expectations.remove(&path);
+            } else {
+                expectations.insert(path, result);
+            }
+        }
+
+        // We convert to a JSON value first because that way the paths are
+        // ordered alphabetically.
+        // See https://stackoverflow.com/questions/67789198.
```
aapoalas, line 584: **praise:** Nice comment <3

aapoalas (approved):
Had a bunch of nits to pick here and there, and suggestions that make fairly little difference anywhere but all in all excellent stuff! Absolutely wonderful to get test262 running! <3

tests/test262_runner.rs
andreubotella: In a follow-up I will make the test runner parallel, and there these fields will live in a mutable thread-local struct, which then gets joined. But for now, while the test runner is serial, I think we can keep this as a temporary unoptimal solution.

tests/test262_runner.rs
andreubotella: It seems like filters can be invalid in the following situations:
- The path doesn't exist (e.g.  `harnes` [sic] rather than `harness`).
- One of the intermediate components of the path isn't a directory (e.g. `harness/assert-false.js/test`).
- The final resolved path is outside the `test262/test` directory (e.g. `../test.js`, or `/etc/fstab`)

I think for the path not existing, it probably shouldn't panic. Not too sure about the other ones.

PR #237, opened 2024-06-19
fix(small_string): Support null-terminated small strings

andreubotella:
Currently small strings use the zero byte as filler between the end of the string and the end of the buffer. This makes it impossible to represent strings ending in `\0` as a small string, even if they fit the length requirements.

However, since small strings are always UTF-8, it is possible to use a byte like 0xFF as the filler byte, since valid UTF-8 (or WTF-8) strings will never contain that byte. This makes it possible to represent the null character at the end of a small string while distinguishing it from the end of the string.

### DIFF
--- a/small_string/lib.rs
+++ b/small_string/lib.rs
@@ -1,5 +1,7 @@
 #[derive(Clone, Copy, PartialEq, Eq, Hash)]
 pub struct SmallString {
+    /// The string will be padded to 7 bytes with the 0xFF byte, which is never
+    /// contained in valid UTF-8 or WTF-8.
     bytes: [u8; 7],
 }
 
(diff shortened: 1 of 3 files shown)


small_string/lib.rs
```diff
@@ -11,16 +13,16 @@ impl std::fmt::Debug for SmallString {
...
 
     pub fn len(&self) -> usize {
-        // Find the last non-null character and add one to its index to get length.
+        // Find the last non-0xFF character and add one to its index to get length.
         self.bytes
             .as_slice()
             .iter()
             .rev()
-            .position(|&x| x != 0)
+            .position(|&x| x != 0xFF)
```
aapoalas, line 25: **nitpick:** We could now turn this logic back around to look for the first 0xFF. Would be a bit easier to understand.

aapoalas (dismissed):
LGTM!

PR #248, opened 2024-06-22
feat(tests): Add `eval-test` subcommand to the test262 runner

andreubotella:
This patch adds the `eval-test` subcommand to the test262 runner, which runs an individual test262 test in the same way the regular test262 runner would, except that it prints its output.

To be able to do this, this patch creates a `BaseTest262Runner` struct, which is used to run a single test and get its result, splitting it from the `Test262Runner` struct which walks through the test directories and compares the test results with a map of expectations.


aapoalas (approved):
LGTM!

PR #260, opened 2024-06-26
feat(ecmascript): Implement internal methods for string exotic objects

andreubotella:


nova_vm/src/ecmascript/builtins/primitive_objects.rs
```diff
@@ -183,7 +190,212 @@ impl InternalSlots for PrimitiveObject {
...
-impl InternalMethods for PrimitiveObject {}
+impl InternalMethods for PrimitiveObject {
+    fn internal_get_own_property(
+        self,
+        agent: &mut Agent,
+        property_key: PropertyKey,
+    ) -> JsResult<Option<PropertyDescriptor>> {
+        // For non-string primitive objects:
+        // 1. Return OrdinaryGetOwnProperty(O, P).
+        if let Some(backing_object) = self.get_backing_object(agent) {
```
aapoalas, line 201: **issue:** String primitive objects can also have a backing store.

nova_vm/src/ecmascript/builtins/primitive_objects.rs
```diff
@@ -183,7 +190,212 @@ impl InternalSlots for PrimitiveObject {
...
+            let mut keys = Vec::with_capacity(len + 1);
+
+            // Insert keys for every index into the string.
+            keys.extend(
+                (0..len)
+                    .map(|idx| PropertyKey::Integer(SmallInteger::try_from(idx as u64).unwrap())),
+            );
+
+            let backing_object_keys;
+            let (integer_keys, other_keys) = match self.get_backing_object(agent) {
```
aapoalas, line 366: **praise:** Bravo, this is awesome! I would not have found such a neat solution to this, at ALL!

nova_vm/src/ecmascript/types/language/string.rs
```diff
@@ -310,6 +310,44 @@ impl String {
         }
     }
 
+    pub fn get_property_descriptor(
```
aapoalas, line 313: **nitpick:** I'd keep this `pub(crate)` as it's not really a nice public API.

aapoalas (changes requested):
Awesome stuff! There was one internal method that seemingly has the order of checks reversed with regards to backing objects and string primitive objects.

nova_vm/src/ecmascript/types/language/string.rs
andreubotella: Done.

nova_vm/src/ecmascript/builtins/primitive_objects.rs
andreubotella: Done.

aapoalas: Ah, oh yeah I see. I guess it'd be a little more performant for strings to first check the index and length property descriptors first, as those are the most likely thing to be looked up. But! Primitive objects are stupid and rare and their performance doesn't matter at all, so here the deduplication in code that we achieve is definitely the better choice.

aapoalas (approved):

nova_vm/src/ecmascript/builtins/primitive_objects.rs
andreubotella: Also, this is only suboptimal for string objects with backing objects, so that makes this deduplication even less worth the trouble.

PR #263, opened 2024-06-27
fix(ecmascript): Fix bugs related to negative zero

andreubotella:


aapoalas (approved):
Ooh, very nice catches! Thank you <3

PR #264, opened 2024-06-27
fix(ecmascript): Only implicit return in concise arrow functions

andreubotella:
If a function ends in a non-return expression, currently Nova will use that as the return value for the function. In JS this should only be the case for concise arrow functions – i.e, arrow functions which have a single expression, rather than a block, as their body – and the return value for any other functions should be `undefined`.

Aside from arrow functions, one big reason why Nova was returning implicitly is that the return value of `Vm::execute` was being used both as the return value for functions, as well as the result of script evaluation, which is used in script tests. To avoid breaking them, this function adds the `ExecutionResult` enum, which lets the return value of `Vm::execute` distinguish between JS returns vs non-returns.


nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -172,12 +172,6 @@ impl Executable {
             stmt.compile(&mut ctx);
         }
 
-        if ctx.exe.instructions.last() != Some(&Instruction::Return.as_u8()) {
```
aapoalas, line 175: **suggestion:** Would it be simpler to keep injecting this but only based on a boolean parameter (or a boolean in `exe`) that is passed `true` for scripts and concise body arrow functions?

aapoalas (approved):
I had an alternative suggestion on how to do this, but this is entirely fine to me as well.

nova_vm/src/engine/bytecode/executable.rs
andreubotella: Done.

PR #269, opened 2024-06-30
fix(bytecode): Assignments should return the right-hand side

andreubotella:
Per the spec, the value of `a = b` should be `b`, which is needed for assignment chaining (e.g. `a = b = c`). However, this currently returns `undefined` for bare assignment, although not for assignment operators. This patch fixes that.


aapoalas (dismissed):
LGTM <3

aapoalas (approved):

PR #270, opened 2024-06-30
fix(ecmascript): Fix `Object.is{Extensible,Sealed,Frozen}`

andreubotella:
The implementation of `Object.isSealed` and `Object.isFrozen`, although it did not panic, it just returned the passed object, which is never a valid behavior per the spec. Although `Object.isExtensible` was implemented per the spec for objects, for primitives it returned the passed value rather than `false`. This patch fixes these issues.


aapoalas (approved):
LGTM <3 The "implementations" were probably some half-hearted starts that I wrote and then forgot to fill in :D

aapoalas (approved):

PR #272, opened 2024-06-30
feat(ecmascript): remainder and shift operators

eliassjogreen:
Wanted to try out the new test262 runner now that I am back from my vacation so heres a little PR implementing the remainder and shift operators :)


nova_vm/src/ecmascript/types/language/number.rs
```diff
@@ -18,6 +18,7 @@ use crate::{
 };
 
 pub use data::NumberHeapData;
+use num_traits::PrimInt;
```
aapoalas, line 21: **issue:** This is probably an accidental import from writing `Primitive`?

aapoalas (changes requested):
Looks good to me except for one probably mistaken import, and then the expectations seem to be off. Might it be that you have so much memory on your machine that it's fine with using 4 GB of RAM in a single array?

eliassjogreen:
Heh yeah, my machine has 32g so I guess it ran without a problem 😅.

nova_vm/src/ecmascript/types/language/number.rs
```diff
@@ -801,7 +858,50 @@ impl Number {
...
+        Number::from(lnum.signed_shr(shift_count))
+    }
+
+    /// ### [6.1.6.1.11 Number::unsignedRightShift ( x, y )](https://tc39.es/ecma262/#sec-numeric-types-number-unsignedRightShift)
+    ///
+    /// The abstract operation Number::lessThan takes arguments x (a Number)
+    /// and y (a Number) and returns a Boolean or undefined.
+    pub fn unsigned_right_shift(agent: &mut Agent, x: Self, y: Self) -> Self {
+        // 1. Let lnum be ! ToUint32(x).
+        let lnum = to_int32(agent, x.into_value()).unwrap();
```
aapoalas, line 897: **issue:** Should be ToUint here

aapoalas (changes requested):
Found one mistaken place.

aapoalas (approved):
Looks good to me, thank you very much! <3

PR #273, opened 2024-06-30
feat(ecmascript): Implement `delete` for object properties

andreubotella:


aapoalas (approved):
**praise:** That's a damn great line count change! Almost 10x removal rate <3

PR #277, opened 2024-07-02
feat(test262): Add a set of skipped tests

andreubotella:
This is meant to reduce the effects of tests that tend to time out on the test262 test runner.


aapoalas (approved):
LGTM, good change <3

PR #278, opened 2024-07-02
feat(ecmascript): Implement `eval`

eliassjogreen:
Lots of code is re-used/copied from `global_declaration_instantiation` and could probably be refactored to look a lot nicer and prevent duplicate code.

Should hopefully fix **a lot** of test262 tests which use `eval`.

Co-author: @load1n9


nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -112,15 +127,457 @@ impl BuiltinIntrinsic for GlobalObjectUnescape {
...
+    // 11. Perform the following substeps in an implementation-defined order, possibly interleaving parsing and error detection:
+    // a. Let script be ParseText(x, Script).
+    let allocator = Allocator::default();
+    let source_text = x.as_str(agent).to_owned();
+    let parser = Parser::new(&allocator, &source_text, SourceType::default());
+    let ParserReturn {
+        errors,
+        program: script,
+        ..
+    } = parser.parse();
```
eliassjogreen, line 204: **todo:** `ParseText` still needs to be implemented.

nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -112,15 +127,457 @@ impl BuiltinIntrinsic for GlobalObjectUnescape {
...
+
+            // TODO:
+            // v. Let classFieldInitializerName be F.[[ClassFieldInitializerName]].
+            // vi. If classFieldInitializerName is not empty, set inClassFieldInitializer to true.
+        }
+    }
+
+    // 11. Perform the following substeps in an implementation-defined order, possibly interleaving parsing and error detection:
+    // a. Let script be ParseText(x, Script).
+    let allocator = Allocator::default();
```
eliassjogreen, line 197: *question:* Should we always create a new allocator for `ParseText` or is it something that should be reused?

aapoalas: So currently we actually rely on the allocator outliving the Agent. Here that is not the case, so this will lead to use-after-free bugs when eg. the eval'd text returns a function, and that then gets called outside the eval.

The way this should all eventually work is for the allocator to actually live with the Agent, though even that actually poses a bit of a question as the oxc allocator doesn't actually support deallocating things. So even if a script or module gets dropped, the allocator would still contain the whole AST for it.

nova_vm/src/ecmascript/builtins/global_object.rs
aapoalas: I think `ParseText` is just this `parser.parse()`, ie. turn the source string into an AST.

nova_vm/src/ecmascript/execution/environments/private_environment.rs
```diff
@@ -36,12 +36,12 @@ pub struct PrivateEnvironment {
     /// The PrivateEnvironment Record of the nearest containing class. null if
     /// the class with which this PrivateEnvironment Record is associated is
     /// not contained in any other class.
-    outer_private_environment: Option<PrivateEnvironmentIndex>,
+    pub outer_private_environment: Option<PrivateEnvironmentIndex>,
```
aapoalas, line 39: **nitpick:** `pub(crate)`

nova_vm/src/ecmascript/builtins/global_object.rs
eliassjogreen: Yep pretty much, but it also handles any exceptions which may have been thrown and converts them to `SyntaxError`s.

nova_vm/src/ecmascript/builtins/global_object.rs
eliassjogreen: Aha, ok. Should I add a `allocator` property to the agent for now or do you have any other solution?

aapoalas: Hmm... A temporary ugly solution would be to leak the allocator. But adding the allocator property to the agent is also fine.

nova_vm/src/ecmascript/execution/environments/private_environment.rs
```diff
@@ -36,12 +36,12 @@ pub struct PrivateEnvironment {
...
     /// the class with which this PrivateEnvironment Record is associated is
     /// not contained in any other class.
-    outer_private_environment: Option<PrivateEnvironmentIndex>,
+    pub(crate) outer_private_environment: Option<PrivateEnvironmentIndex>,
 
     /// ### \[\[Names\]\]
     ///
     /// The Private Names declared by this class.
-    names: HashMap<String, PrivateName>,
+    pub names: HashMap<String, PrivateName>,
```
aapoalas, line 44: **nitpick:** Sorry, here too. Always default to (at most) `pub(crate)`. Only do `pub` if absolutely necessary.

nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -112,15 +134,511 @@ impl BuiltinIntrinsic for GlobalObjectUnescape {
...
+        if let VarScopedDeclaration::Variable(d) = d {
+            // i. For each String vn of the BoundNames of d, do
+            let mut bound_names = vec![];
+            d.id.bound_names(&mut |identifier| {
+                bound_names.push(identifier.name.clone());
+            });
+            for vn in bound_names {
+                // 1. If declaredFunctionNames does not contain vn, then
+                if !declared_function_names.contains(&vn) {
+                    let vn = String::from_str(agent, vn.as_str());
```
aapoalas, line 486: **suggestion:** I think you can take ownership of `vn` and use `from_string`.

nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -112,15 +134,511 @@ impl BuiltinIntrinsic for GlobalObjectUnescape {
...
+                        const_bound_names.push(String::from_str(agent, identifier.name.as_str()))
+                    });
+                } else {
+                    decl.id.bound_names(&mut closure)
+                }
+            }
+            LexicallyScopedDeclaration::Function(decl) => decl.bound_names(&mut closure),
+            LexicallyScopedDeclaration::Class(decl) => decl.bound_names(&mut closure),
+            LexicallyScopedDeclaration::DefaultExport => {
+                bound_names.push(String::from_static_str(agent, "*default*"))
```
aapoalas, line 534: **suggestion:** I think this is `BUILTIN_STRINGS_MEMORY._default_`.

nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -112,15 +134,511 @@ impl BuiltinIntrinsic for GlobalObjectUnescape {
...
+        errors,
+        program: script,
+        ..
+    } = parser.parse();
+
+    // b. If script is a List of errors, throw a SyntaxError exception.
+    if !errors.is_empty() {
+        // Make sure `script` can't borrow `source_text` so we can return it.
+        drop(script);
+        // TODO: Include error messages in the exception.
```
aapoalas, line 220: **suggestion:** We could unleak the allocator Box here and in the `is_empty()` branch and collect them: That way we at least do not leak memory that will never be used.

aapoalas (dismissed):
Looks good to me, and awesome work getting this big of a feature done!

I'll merge this once all the comments are resolved: You are absolutely free to resolve them without code changes, as none of these are **issues** or otherwise marked **(blocking)**. ie. I am explicitly saying: I would change this or that personally, but I am absolutely okay with them not being changed.

nova_vm/src/ecmascript/builtins/global_object.rs
eliassjogreen: It's a `Atom` so I think that would just do the same thing: `String::from_string(agent, vn.into_string());`

aapoalas: Oh yeah, true. My bad.

aapoalas (approved):

PR #280, opened 2024-07-04
feat(vm): Implement direct `eval`

andreubotella:
In JS, the behavior of the `eval` function depends on whether it is called directly through the `eval` identifier ("direct eval"), or whether it is called in any other way ("indirect eval"). The latter runs code in the global scope, whereas the former runs in the scope of the function that calls it.

In the spec, the `eval` function always performs an indirect eval when run, but the syntax to call a function whose callee reference is the `eval` identifier is special-cased, so that if the identifier happens to reference the `eval` intrinsic, then a direct eval is performed.

PR #278 implemented the internal `PerformEval` spec operation, including the parts relevant to direct eval, but it only exposed indirect eval. This patch implements direct eval by adding a `DirectEvalCall` instruction which performs direct eval if the `eval` identifier references the current realm's `eval` intrinsic, and otherwise calls that identifier's value instead.

This patch regresses two tests that were previously passing. These were tests that started passing in the previous indirect eval PR, and which need the eval to run in strict mode, whereas this PR implements it to always run in non-strict mode. Nova does not yet distinguish strict versus non-strict mode, and performing direct eval in strict mode would cause a higher number of test regressions.


eliassjogreen (approved):
Looks good to me!

aapoalas (approved):
LGTM!

PR #281, opened 2024-07-06
refactor(ecmascript): Improve conversions between Number and Rust numeric types

andreubotella:
The `Number` ECMAScript type had a `From<i64>` implementation, although not all `i64` values can be converted to a `Number`. This implementation, in fact, casts the value into the safe integer range, which is not what is typically expected of the `From<T>` trait.

This patch instead turns this implementation into a `TryFrom<i64>` impl, which returns an `Err` when the value cannot be safely cast.

Additionally, since `Number` did not yet provide utility methods for converting to/from `usize`, which would be common for arrays and strings, this patch adds `TryFrom<usize>` and the `into_usize()` method. Notably, `number.into_usize()` is not equivalent to `number.into_i64() as usize` because negative numbers clamp to 0, rather than clamping to `i64::MIN` and subsequently being bytecast.

Furthermore, this patch also adds documentation to the `into_i64()` and `into_usize()` methods describing the conversion details in cases where the number isn't a valid member of the output type.


aapoalas (approved):
Very good improvement to add some documentation, and the Try trait had bothered me as well. LGTM!

PR #283, opened 2024-07-06
fix(ecmascript): Make `Value::is_object()` support all object types

andreubotella:
The `Value::is_object()` method currently checks whether the value is certain specific types of object, rather than checking all of them. This patch changes the method to instead check that it's not a primitive. This also makes it more future-proof, since new object types get added to the language regularly, whereas new primitives are a lot more rare.

In test262 tests, this seems to primarily affect primitive object comparison: `new Boolean(true) === new Boolean(true)` was false before this patch, since strict equality checks that both sides have the same JS type, which relies on `is_object()`.


aapoalas (commented):
Looks good to me, but a suggestion: Should we instead make `is_object` an inline alias for `Object::try_from` (maybe mapped to Option)?

Basically every place in the spec always asks IsObject? and then uses the parameter as an object. In Rust that would fairly well translate into a type narrowing instead of just a bool return value.

andreubotella:
> Looks good to me, but a suggestion: Should we instead make `is_object` an inline alias for `Object::try_from` (maybe mapped to Option)?
> 
> Basically every place in the spec always asks IsObject? and then uses the parameter as an object. In Rust that would fairly well translate into a type narrowing instead of just a bool return value.

Not all places that use `Value::is_object` immediately cast it to an `Object`. Instead, the ones that do should probably be changed to branch on whether `Object::try_from` succeeds.

But even then, the `is_object` function body should probably use `Object::try_from` indeed.

aapoalas (approved):
LGTM

PR #284, opened 2024-07-06
feat(ecmascript): Make `{Boolean,String,Number}.prototype` primitive objects

andreubotella:
Per the spec, the Boolean, String and Number prototype objects are themselves primitive objects of the corresponding kind, their value being `false`, the empty string, and zero, respectively. This patch implements this.


aapoalas (approved):
LGTM! <3

PR #285, opened 2024-07-07
fix(ecmascript): Fix `length` & `name` deletion in built-in functions

andreubotella:
Since the `length` and `name` properties in built-in functions are stored in the built-in function heap data, the `internal_delete` implementation needs to create a backing object if one of those properties is being deleted.

In this implementation, however, a backing object was being created *with the property that was being deleted*, rather than with the other property. The result of this is that deleting the `length` property would not delete `length` but would delete `name`, and vice versa.

This patch fixes it by instead relying on `self.create_backing_object()`, which did not yet exist when that implementation was written.


aapoalas (approved):
LGTM

PR #298, opened 2024-07-11
refactor(ecmascript): Add default implementations to `HostHooks`

andreubotella:
The host hooks defined in the JS spec include multiple disparate things, from letting the host control on-demand code evaluation, to integration with HTML's concept of an incumbent realm, to enqueuing jobs for the host to later run.

Nova's current model only allows an all-or-nothing approach: either you implement `HostHooks` by yourself, in which case you don't have any of the spec-provided default implementations, or you use `DefaultHostHooks`, which does not let you customize them.

This does not work for host hooks such as `HostEnqueuePromiseJob` and other scheduling hooks, since the spec does not give default implementations. This makes sense, since it is up to the host, not to the JS engine, to determine job scheduling.

This patch moves the implementation of `HostHooks` methods from `DefaultHostHooks` into the trait. It also adds documentation to indicate that in the future, scheduling host hooks without a default implementation will be a no-op in `DefaultHostHooks`, and recommends against the usage of `DefaultHostHooks` for anything but the simplest synchronous use cases.

### DIFF
--- a/nova_vm/src/ecmascript/execution/agent.rs
+++ b/nova_vm/src/ecmascript/execution/agent.rs
@@ -52,8 +52,17 @@ impl JsError {
 // pub struct PreAllocated;
 
 pub trait HostHooks: std::fmt::Debug {
-    fn host_ensure_can_compile_strings(&self, callee_realm: &mut Realm) -> JsResult<()>;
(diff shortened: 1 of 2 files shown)


aapoalas (approved):
LGTM<3

PR #299, opened 2024-07-12
refactor(ecmascript): Refactor the promise resolving functions

andreubotella:
The promise resolving functions are the `resolve` and `reject` functions that are passed to the callback passed to the `Promise` constructor. These functions are not built-ins, and they cannot be implemented as `BuiltinFunction`s because they are tied to the constructed promise – instances of them are created for each promise, and their heap data must reference that promise.

In Nova, these functions were implemented as two separate variants of `Value`: `BuiltinPromiseResolveFunction` and
`BuiltinPromiseRejectFunction`. However, only `BuiltinPromiseRejectFunction` had a heap vector; resolve functions didn't. This would be clearly unworkable.

Furthermore, since both kinds of resolving function are identical in everything except for their behavior when called, it might be better to instead use a single `BuiltinPromiseResolvingFunction` value variant and heap vector, with an enum to distinguish them.

This patch also updates the heap data for `BuiltinPromiseResolvingFunction` to use a `PromiseCapability` rather than a `Promise`, and to remove `already_resolved`. Since #297, `PromiseCapability` is the object that allows a promsie to be resolved, and it also (along with the promise's internal state) tracks the equivalent of the spec's already resolved record.


aapoalas (approved):
Excellent idea! The resolving functions were just left unfinished as I ran out of steam the night I was working on that :D But this is a great improvement: Definitely having two heap vectors for these is probably not worth it.

PR #300, opened 2024-07-12
feat(ecmascript): Basic (synchronous) `Promise` implementation

andreubotella:
This patch has a basic implementation of `Promise` that only covers synchronous operations. It implements the following APIs:

- `Promise` constructor
- `Promise.resolve`
- `Promise.reject`
- `Promise.prototype.then`
- `Promise.prototype.catch`

This patch does not yet implement promise jobs, leaving them for a follow-up patch.


nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -102,20 +174,34 @@ impl PromiseConstructor {
...
-    fn reject(
-        _agent: &mut Agent,
-        _this_value: Value,
-        _arguments: ArgumentsList,
-    ) -> JsResult<Value> {
-        todo!()
+
+    fn reject(agent: &mut Agent, _this_value: Value, arguments: ArgumentsList) -> JsResult<Value> {
+        // 1. Let C be the this value.
+        // 2. Let promiseCapability be ? NewPromiseCapability(C).
```
aapoalas, line 180: **suggestion:** Should we maybe intentionally crash on Promise subclassing?

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -102,20 +174,34 @@ impl PromiseConstructor {
...
+        let promise_capability = PromiseCapability::new(agent);
+        // 3. Perform ? Call(promiseCapability.[[Reject]], undefined, « r »).
+        promise_capability.reject(agent, arguments.get(0));
+        // 4. Return promiseCapability.[[Promise]].
```
aapoalas, line 184: **thought:** The spec does a lot of unnecessary things here in the usual case.

I presume in the case of actual `Promise.reject(new Error())` the only thing you _really_ need to do is create a new Promise with its internal state as rejected, right?

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_prototype.rs
```diff
@@ -34,16 +50,73 @@ impl Builtin for PromisePrototypeThen {
 }
 
 impl PromisePrototype {
-    fn catch(_agent: &mut Agent, _this_value: Value, _: ArgumentsList) -> JsResult<Value> {
-        todo!()
+    fn catch(agent: &mut Agent, this_value: Value, args: ArgumentsList) -> JsResult<Value> {
+        // 1. Let promise be the this value.
+        // 2. Return ? Invoke(promise, "then", « undefined, onRejected »).
+        let property_key = PropertyKey::from_static_str(agent, "then");
```
aapoalas, line 56: **nitpick:** Here too.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_prototype.rs
```diff
@@ -34,16 +50,73 @@ impl Builtin for PromisePrototypeThen {
...
+        let on_rejected = args.get(0);
+        if let Value::Promise(promise) = this_value {
+            let then_is_overriden = match agent[promise].object_index {
```
aapoalas: **issue:** It's also possible that then is overridden at the Promise prototype level, as stupid as that is :(

aapoalas (changes requested):
Found one issue but otherwise looks damn fine!

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_capability_records.rs
```diff
@@ -36,22 +43,181 @@ pub(crate) struct PromiseCapability {
...
+            PromiseState::Pending { is_resolved, .. }
+                if *is_resolved || !self.must_be_unresolved => {}
+            _ => {
+                // 5. If alreadyResolved.[[Value]] is true, return undefined.
+                return;
+            }
+        }
+
+        // 7. Perform RejectPromise(promise, reason).
+        self.reject(agent, reason);
```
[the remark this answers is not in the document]
aapoalas: Oh I see clippy already found this, nice <3

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
andreubotella: Now that you mention it, yeah, this could be simplified. In fact, *in every case* the only thing you'd have to do is create an already rejected promise. This will break down with promise subclasses, of course, but for now we're not caring at all about them.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_prototype.rs
andreubotella: Properly handling this turned out to be more complicated than anticipated, so we're instead making everything take the slow path in this PR. We'll revisit the fast path in a follow-up.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
andreubotella: Fixed the constructor, `reject` and `resolve` to do so.

aapoalas (approved):

PR #301, opened 2024-07-12
refactor(ecmascript): Return Option<Function> from IsCallable and IsConstructable

aapoalas:
The places where the result is not immediately used to call or construct are so rare as to not really exist.

This also fixed a bug where undefined get/set values in descriptors would cause a crash.


nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -53,37 +56,61 @@ pub(crate) fn is_array(_agent: &Agent, argument: Value) -> JsResult<bool> {
 /// The abstract operation IsCallable takes argument argument (an ECMAScript
 /// language value) and returns a Boolean. It determines if argument is a
 /// callable function with a [[Call]] internal method.
-pub(crate) fn is_callable(argument: impl IntoValue) -> bool {
+///
+/// > #### Note
+/// > Nova breaks with the specification to narrow the types automatically, and
+/// > returns an `Option<Function>`. Eventually this should become
+/// > `Option<Callable>` once callable proxies are supported.
+pub(crate) fn is_callable(argument: impl TryInto<Function>) -> Option<Function> {
```
andreubotella, line 64: I think the name should be `to_callable`, right?

aapoalas: I thought I'd keep the spec name.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -53,37 +56,61 @@ pub(crate) fn is_array(_agent: &Agent, argument: Value) -> JsResult<bool> {
...
+///
+/// The abstract operation IsConstructor takes argument argument (an ECMAScript
+/// language value) and returns a Boolean. It determines if argument is a
+/// function object with a [[Construct]] internal method.
+///
+/// > #### Note
+/// > Nova breaks with the specification to narrow the types automatically, and
+/// > returns an `Option<Function>`. Eventually this should become
+/// > `Option<Callable>` once callable proxies are supported.
+pub(crate) fn is_constructor(agent: &mut Agent, constructor: impl IntoValue) -> Option<Function> {
```
andreubotella, line 81: The spec concept of callable matches one to one to our concept of `Function`, but that is not the same for constructors. Would it make sense to split this function into `Function::is_constructor` (that takes a `Function` and returns a boolean), and `to_constructor`, that takes a value (or `IntoValue`) and returns an `Option<Function>`?

aapoalas: Yeah, that definitely sounds like a good idea, since it's essentially determining if a Function has a `[[Construct]]` internal method.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
aapoalas: The reasoning for keeping the name is that I don't want to have confusion over spec text saying `IsCallable(foo)` and our code then doing `to_callable(foo)` (or maybe rather `try_into_callable`? `to_X` usually suggests a conversion or a clone, I think). While the difference is small, it might end up confusing in the long run.

Now that is not to say that `is_callable` returning a non-bool isn't confusing... But the repeated `Function::try_from(foo).unwrap()` wasn't very nice either.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
aapoalas: I split the method so that the `is_constructor` now takes an `impl TryInto<Function>` like `is_callable`, and calls into `Function::is_constructor` which delegates to the different Function variants for the internal details of how constructorhood is defined.

andreubotella (approved):
As I said before, I would prefer `to_{callable,constructor}`, but I won't block on it. Otherwise, LGTM.

PR #302, opened 2024-07-12
feat(tests): Support async tests in the test262 runner

andreubotella:
For async test262 tests, both successful and unsuccessful tests must call the (engine-provided) `print()` function with strings of a particular kind. If it is not called (e.g. because a promise does not resolve when it should), the test runner is supposed to "select an appropriate length of time to wait before considering the test 'timed out' and failing."

This implies that if the process exits before `print()` is called, it should be considered a timeout. However, we are instead treating it as a different kind of failure, appearing in the expectations file as `"UNRESOLVED"`.

By enabling async tests, which were previously ignored, this patch adds thousands of lines to `expectations.json`. However, these were tests that we were not passing, they were just being ignored until now.


aapoalas (dismissed):
LGTM.

aapoalas (approved):

PR #305, opened 2024-07-13
fix(ecmascript): Fix bug with already-resolved promise determination

andreubotella:
Both the `resolve` and `reject` methods of `PromiseCapability` need to know when the promise capability should be considered as already resolved. This determination relies both on `PromiseCapability`'s `must_be_unresolved`, as well as on the `is_resolved` flag of pending promise states, and it was being done in both methods. However, this determination of the already resolved status had a bug in `reject` that was not in `resolve`.

This patch moves that determination in both of those methods into its own `is_already_resolved` method of `PromiseCapability`, fixing this bug. It also refactors that code to make it clearer.

----

Do not merge before #304.


aapoalas (dismissed):
Nice!

aapoalas (approved):

PR #307, opened 2024-07-13
fix(ecmascript): Array.prototype is an Array

aapoalas:
Built on top of #301 so should merge that first.

This mostly just makes `Array.prototype` an empty array exotic object itself. Stupid thing :D

But I also removed an unnecessary Vec allocation from element array grow code path, and updated the Rust toolchain to 1.79.0.


andreubotella:
**question:** Why are the changes to the heap needed? Are they relevant to `Array.prototype`, or could be split out into a separate PR?

aapoalas:
> **question:** Why are the changes to the heap needed? Are they relevant to `Array.prototype`, or could be split out into a separate PR?

Could definitely be split. Just what I happened to run into while I was looking at this.

aapoalas:
I split the PR into three separate ones, see #308 and #309

nova_vm/src/heap/element_array.rs
aapoalas, line 1: Hmm, this is refactoring that got left behind apparently

andreubotella (commented):
The `Array.prototype` changes LGTM.

andreubotella (approved):

PR #309, opened 2024-07-14
perf(vm): Remove an extra heap allocation from element arrays growing

aapoalas:


nova_vm/src/heap/element_array.rs
```diff
@@ -1411,6 +1316,299 @@ impl ElementArrays {
...
+                let remaining = elements.values.spare_capacity_mut();
+                let last = remaining.get_mut(0).unwrap();
+                debug_assert_eq!(
+                    std::mem::size_of::<Option<[Option<Value>; N]>>(),
+                    std::mem::size_of::<[Option<Value>; N]>()
+                );
+                // SAFETY: We can move MaybeUninit from outside of the array into individual items in it.
+                // Moving inside the Option<[_]> is less well defined; the size is asserted to be the same
+                // but it could theoretically be that we end up copying a bit that says "the array is None".
+                // Experimentally however, this works and we do not copy None but Some([_]).
```
andreubotella, line 1364: **issue:** Not sure about this. As you mention, there is some bit pattern for the array being None. Now, since you must be able to take a reference to the `&[Option<Value>; N]`, this means that every valid bit pattern for `[Option<Value>; N]` must correspond to `Some` **as long as no bits are padding bits in every bit pattern**. I'm not sure that we can rule that out.

I won't block on this, I guess, but I'd want to make sure that there's a performance benefit from doing this transmute and `copy_from_slice` over a simpler implementation that iterates through the elements and writes them into the target slice.

andreubotella (approved):

nova_vm/src/heap/element_array.rs
aapoalas: Yeah, it's honestly sort of clear that this is **not okay** in truth. But, this does not actually change the safety level from before:

The `push_with_key` code takes the starting address of the `Option<[Option<Value>; N]>` and transmutes it into an address of an `Option<Value>`. It then calls `copy_nonoverlapping` on to copy from the parameter vector into the target memory, and finally one-by-one writes all the rest of the elements in the target memory with `None`.

There is no real guarantee that we end up writing the `Some([_])` bit pattern at some point. And the most annoying thing is that we can't really help this. Pushing into the Vector means the data needs to pass through the stack, fully, and for a 4GB memory slice that's going to overflow the stack immediately when we call this function, regardless of how small a vector we're actually going to copy.

So anyway, we should be fully as unsafe as we already were. And a performance benefit is, well, likely going to be there since this does avoid double copying. The larger the elements vector, the larger the effect there will be.

I'll run MIRI on this to make sure this hasn't done any nasty things, though.

aapoalas: Dammit, can't run MIRI because scope analysis is doing dirty, dirty things on references that derive from the Agent borrow.

aapoalas: Got MIRI running; all green.

PR #310, opened 2024-07-14
feat(vm): for-of iteration

aapoalas:


nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -927,13 +992,62 @@ impl Vm {
...
+                        vm.result = Some(result);
+                    }
+                }
+            }
+            Instruction::IteratorValue => {
+                if matches!(
+                    vm.iterator_stack.last().as_ref().unwrap(),
+                    VmIterator::GenericIterator(_)
+                ) {
+                    // Generic iterators have to access the "value" property of the result object.
```
andreubotella, line 1030: **nitpick:** Wouldn't it be better to instead have a `VmIterator::next` method that abstracts away the differences between implementations and returns a `JsResult<Option<Value>>`? IIUC, this would simplify things, and would avoid the need for the `IteratorValue` instruction, right?

tests/expectations.json
```diff
@@ -32179,28 +32143,16 @@
...
   "language/statements/for-of/iterator-close-via-continue.js": "CRASH",
   "language/statements/for-of/iterator-close-via-return.js": "CRASH",
   "language/statements/for-of/iterator-close-via-throw.js": "CRASH",
-  "language/statements/for-of/iterator-next-error.js": "CRASH",
-  "language/statements/for-of/iterator-next-reference.js": "CRASH",
-  "language/statements/for-of/iterator-next-result-done-attr.js": "CRASH",
   "language/statements/for-of/iterator-next-result-type.js": "CRASH",
-  "language/statements/for-of/iterator-next-result-value-attr-error.js": "CRASH",
-  "language/statements/for-of/iterator-next-result-value-attr.js": "CRASH",
+  "language/statements/for-of/iterator-next-result-value-attr-error.js": "TIMEOUT",
```
andreubotella, line 32151: **nitpick:** Would be better to add this test to `skip.json`.

nova_vm/src/engine/bytecode/instructions.rs
```diff
@@ -244,10 +244,20 @@ pub enum Instruction {
...
     /// Call [[NextMethod]] of current iterator
     IteratorNext,
+    /// Check "done" boolean of current iterator result object
     IteratorComplete,
+    // Extract "value" of current iterator result object
     IteratorValue,
+    /// Perform CloseIterator on the current iterator
+    IteratorClose,
+    /// Perform AsyncCloseIterator on the current iterator
+    AsyncIteratorClose,
```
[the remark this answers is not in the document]
aapoalas: Yeah, that sounds reasonable. Currently the act of starting an iterator also only works for synchronous iterators, and an asynchronous iterator will need an intermediate await. This will definitely all need some instruction work and massaging in the future.

Let's leave it here for now as a marker of sorts, though.

Side note: That'd be a **suggestion** or **issue (nonblocking)** maybe? :)

tests/expectations.json
aapoalas: I thought I'd leave this in: The reason it's failing is because we don't support getters and setters yet, so an iterator never breaks.

That will be fixed very shortly, and it's also a properly valid "timeout" case instead of just being "copying 4GB of memory isn't fast enough" :)

andreubotella: Fair enough.

nova_vm/src/engine/bytecode/vm.rs
aapoalas: This was an excellent idea! I refactored to join these together into the `IteratorStepValue` spec algorithm, essentially.

It was already doing that, only that the `IteratorStepValue` also says to set the `done` value to `true` everywhere. We're still skipping that since the iterator record is about to be destroyed.

andreubotella (approved):

PR #315, opened 2024-07-15
fix(vm): Fix bugs with increment and decrement operators

andreubotella:
Prefix and postfix increment and decrement operators had the following two bugs:

- Prefix was returning the value before increment/decrement, and postfix the value before it; rather than the other way around.
- For postfix, the returned value should not be the reference's value, but the result of applying `ToNumeric()` on it.

This patch fixes both of them.


aapoalas (approved):
Nice

PR #316, opened 2024-07-15
fix(ecmascript): Fix prototypes for async and generator functions

andreubotella:
Currently async, generator, and async generator functions all have `Function.prototype` as their prototype. The reason for this is that all of those functions get instantiated as regular `ECMAScriptFunction` objects, which always have `Function.prototype` as their prototype.

`Value` has variants for `ECMAScriptAsyncFunction`, etc, but those are not being used, and they don't have any heap vector associated. Since the function instantiation already works, and is almost identical for async / generator / async generator functions, it is better to reuse `ECMAScriptFunction`. Furthermore, `Value` also has a similar variant for `ECMAScriptConstructorFunction`, even though `ECMAScriptFunction` already can represent constructors just fine.

This patch therefore deletes those types of value, and makes `ECMAScriptFunction::internal_prototype` depend on whether the function is async or a generator.


aapoalas:
Hmmm... My thinking has been that instead of `ECMAScriptFunction` being a collection of booleans, the Value variants + separate heap vectors would serve the purpose of splitting the different cases apart.

The idea then being both to enable some individual optimizations (like generators apparently have a `prototype` field that is not configurable and could thus be handled at the exotic heap data level fairly easily) but especially to possibly allow some optimizations at the algorithm level. eg. If we're calling a generator, we know we're going to get a generator response object. If we're calling an async function, we know we're going to get a Promise.

I've been hoping that those might end up being quite useful things to know "on the stack".

andreubotella:
> Hmmm... My thinking has been that instead of `ECMAScriptFunction` being a collection of booleans, the Value variants + separate heap vectors would serve the purpose of splitting the different cases apart.
> 
> The idea then being both to enable some individual optimizations (like generators apparently have a `prototype` field that is not configurable and could thus be handled at the exotic heap data level fairly easily) but especially to possibly allow some optimizations at the algorithm level. eg. If we're calling a generator, we know we're going to get a generator response object. If we're calling an async function, we know we're going to get a Promise.
> 
> I've been hoping that those might end up being quite useful things to know "on the stack".

Most of the implementation of everything is identical, at least for regular and async functions, until you get to `evaluate_function_body`, which would handle creating the promise or generator object. So if we go with that approach, it would be better to have functions generic over the function types.

aapoalas (dismissed):
Yeah, I guess this is fine. We can always revisit this at a later time as well. Do you plan on adding the steps like in https://tc39.es/ecma262/#sec-runtime-semantics-instantiategeneratorfunctionobject to prep the prototype property in this PR as well?

aapoalas (dismissed):

aapoalas (approved):

PR #318, opened 2024-07-15
feat(ecmascript): Implement the rest of `Math`

eliassjogreen:
This PR implements the remainder of the `Math` object, with the exception of [`f16round`](https://tc39.es/proposal-float16array/#sec-math.f16round).

cc @load1n9


nova_vm/src/ecmascript/builtins/numbers_and_dates/math_object.rs
```diff
@@ -377,87 +475,332 @@ impl MathObject {
...
-        let _x = to_number(agent, arguments.get(0))?;
-        todo!();
+        // 1. Let a be ℝ(? ToUint32(x)).
+        let a = to_uint32(agent, arguments.get(0))?;
+
+        // 2. Let b be ℝ(? ToUint32(y)).
+        let b = to_uint32(agent, arguments.get(1))?;
+        
+        // 3. Let product be (a × b) modulo 2**32.
+        let product = (a * b) % 2u32.pow(32);
```
aapoalas, line 709: **nitpick:** I think this is `wrapping_mul`.

nova_vm/src/ecmascript/builtins/numbers_and_dates/math_object.rs
```diff
@@ -377,87 +475,332 @@ impl MathObject {
...
+
+        // 4. If product ≥ 2**31, return 𝔽(product - 2**32); otherwise return 𝔽(product).
+        if product >= 2u32.pow(31) {
+            Ok(Value::from(product as i32 - 2i32.pow(32)))
```
aapoalas, line 713: **issue:** And this overflows. Actually, this 4. step is seemingly just equal to `product as i32`.

nova_vm/src/ecmascript/types/language/number.rs
```diff
@@ -287,6 +287,22 @@ impl Number {
         }
     }
 
+    pub fn is_pos_one(self, agent: &Agent) -> bool {
+        match self {
+            Number::Number(n) => 1.0 == agent[n],
```
aapoalas, line 292: **note:** Only `Number::Integer` should ever return positive to this. If any other others return true, that's a bug in Nova as it means that our variants are no longer "sound".

nova_vm/src/ecmascript/builtins/numbers_and_dates/math_object.rs
```diff
@@ -756,51 +756,126 @@ impl MathObject {
...
+        if n.is_sign_negative(agent) {
+            return Ok(Value::nan());
+        }
         // 6. Return an implementation-approximated Number value
+        Ok(Value::from_f64(agent, n.into_f64(agent).log2()))
     }
 
     fn max(agent: &mut Agent, _this_value: Value, arguments: ArgumentsList) -> JsResult<Value> {
         // 1. Let coerced be a new empty List.
+        let mut coerced = vec![];
```
aapoalas, line 815: **nitpick:** `Vec::with_capacity(arguments.len())`

aapoalas: Or you can actually do `let coerced = arguments.iter().map(...).collect::<Vec<_>>();`

nova_vm/src/ecmascript/builtins/numbers_and_dates/math_object.rs
```diff
@@ -342,122 +341,666 @@ impl MathObject {
...
+
+        // 2. If n is NaN, n > 1𝔽, or n < -1𝔽, return NaN.
+        if n.is_nan() || n > 1.0 || n < -1.0 {
+            return Ok(Value::nan());
+        }
+
+        // 3. If n is 1𝔽, return +0𝔽.
+        if n == 1.0 {
+            return Ok(Value::zero());
+        }
```
andreubotella, line 355: The Rust `acos` method says it returns `NaN` if the number is outside the range [-1, 1] (and presumably also if the input is `NaN` itself), and it presumably also returns positive zero if the input is 1. So this might not be needed.

These preludes are needed in the JS spec because it talks about the underlying mathematical operation, which is not defined outside of the [-1, 1] range. But I'm pretty sure that for all of these functions, the methods in Rust std (and in libc) would behave the same as JS in these cases.

nova_vm/src/ecmascript/builtins/numbers_and_dates/math_object.rs
```diff
@@ -342,122 +341,665 @@ impl MathObject {
...
+        if n.is_pos_zero(agent)
+            || n.is_neg_zero(agent)
+            || n.is_pos_infinity(agent)
+            || n.is_neg_infinity(agent)
+        {
+            return Ok(n.into_value());
+        }
+
+        // 4. Let n32 be the result of converting n to IEEE 754-2019 binary32 format using roundTiesToEven mode.
+        let n32 = n.into_f32(agent).round_ties_even();
```
andreubotella, line 759: This is wrong, `f32::round_ties_even()` rounds the number *to an integer*, using the round ties to even rounding mode. The spec asks to round ties to even *in the conversion to `f32`*, and in fact, that rounding mode is the one that Rust already uses in `n as f32`: https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast

nova_vm/src/ecmascript/types/language/number.rs
eliassjogreen: Should I change this to `unreachable!()` or keep as is?

aapoalas: unreachable sounds fine.

eliassjogreen: Looks like our number variants are not "sound" everywhere. I reverted changing it to `unreachable!()` with an added note.
```
	"built-ins/Math/log10/Log10-specialVals.js" -- Expected Pass, got Crash
	"built-ins/Math/log/S15.8.2.10_A3.js" -- Expected Pass, got Crash
	"built-ins/Math/log2/log2-basicTests.js" -- Expected Pass, got Crash
	"built-ins/Math/acosh/nan-returns.js" -- Expected Pass, got Crash
	"built-ins/Math/log/S15.8.2.10_A2.js" -- Expected Pass, got Crash
```

load1n9 (approved):
LGTM

nova_vm/src/ecmascript/builtins/numbers_and_dates/math_object.rs
```diff
@@ -342,122 +341,687 @@ impl MathObject {
...
+
+        // 6. Return the ECMAScript Number value corresponding to n64.
+        Ok(Value::from_f64(agent, n64))
     }
 
-    fn hypot(_agent: &mut Agent, _this_value: Value, _arguments: ArgumentsList) -> JsResult<Value> {
-        todo!();
+    fn hypot(agent: &mut Agent, _this_value: Value, arguments: ArgumentsList) -> JsResult<Value> {
+        // 1. Let coerced be a new empty List.
+        let mut coerced = vec![];
```
aapoalas, line 784: **nitpick:** You can prep `Vec::with_capacity(arguments.len())` or do eg. `arguments.iter().map(...).collect::<Vec<_>>()`.

nova_vm/src/ecmascript/builtins/numbers_and_dates/math_object.rs
```diff
@@ -342,122 +341,687 @@ impl MathObject {
...
+            return Ok(Value::nan());
+        }
+        // 6. Return an implementation-approximated Number value
+        Ok(Value::from_f64(agent, n.into_f64(agent).log2()))
     }
 
     fn max(agent: &mut Agent, _this_value: Value, arguments: ArgumentsList) -> JsResult<Value> {
-        let _x = to_number(agent, arguments.get(0))?;
-        todo!();
+        // 1. Let coerced be a new empty List.
```
aapoalas, line 948: **suggestion:** Add a fast path where you check if all arguments are integers (or stupid other simple things that coerce to eg. 0 or 1 in `ToNumber`) and go for an integer-max algorithm if the fast-path check succeeds.

nova_vm/src/ecmascript/builtins/numbers_and_dates/math_object.rs
```diff
@@ -342,122 +341,687 @@ impl MathObject {
...
+
+        // 5. Return highest.
+        Ok(highest.into_value())
     }
 
     fn min(agent: &mut Agent, _this_value: Value, arguments: ArgumentsList) -> JsResult<Value> {
-        let _x = to_number(agent, arguments.get(0))?;
-        todo!();
+        // 1. Let coerced be a new empty List.
+        let mut coerced = Vec::with_capacity(arguments.len());
```
aapoalas, line 989: **suggestion:** Same thing here, integer-fast-path.

aapoalas (approved):
Looks good to me. I did have some suggestions and there's one place(s) where you could optimize by using / adding PI "constants".

nova_vm/src/ecmascript/types/language/number.rs
```diff
@@ -287,6 +287,22 @@ impl Number {
...
+            Number::Integer(n) => 1i64 == n.into(),
+            Number::Float(n) => 1.0 == n,
+        }
+    }
+
+    pub fn is_neg_one(self, agent: &Agent) -> bool {
+        match self {
+            Number::Number(n) => -1.0 == agent[n],
+            Number::Integer(n) => -1i64 == n.into(),
+            Number::Float(n) => -1.0 == n,
```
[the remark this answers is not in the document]
eliassjogreen: Why is that? Wouldn't it be possible to also get `-1` from integer too?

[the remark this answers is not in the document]
aapoalas: Ah sorry oh yeah of course, only Integer should return `-1`. I was thinking about `-0.0`. Sorry for the noise :D

nova_vm/src/ecmascript/builtins/numbers_and_dates/math_object.rs
```diff
@@ -342,122 +341,687 @@ impl MathObject {
...
+        let n = to_number(agent, arguments.get(0))?;
+
+        // 2. If n is one of NaN, +0𝔽, or -0𝔽, return n.
+        if n.is_nan(agent) || n.is_pos_zero(agent) || n.is_neg_zero(agent) {
+            return Ok(n.into_value());
+        }
+
+        // 3. If n is +∞𝔽, return an implementation-approximated Number value representing π / 2.
+        if n.is_pos_infinity(agent) {
+            return Ok(Value::from_f64(agent, consts::FRAC_PI_2));
```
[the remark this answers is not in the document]
eliassjogreen: Something like the `BUILTIN_STRINGS` but for numbers? It would make sense but feels like it is probably a bit out of scope for this pr. Should I create an issue for it instead?

nova_vm/src/ecmascript/builtins/numbers_and_dates/math_object.rs
```diff
@@ -342,122 +341,687 @@ impl MathObject {
...
+        }
+
+        // 6. If ny is +0𝔽, then
+        if ny.is_pos_zero(agent) {
+            // a. If nx > +0𝔽 or nx is +0𝔽, return +0𝔽.
+            if nx.is_sign_positive(agent) || nx.is_pos_zero(agent) {
+                return Ok(Value::pos_zero());
+            }
+            // b. Return an implementation-approximated Number value representing π.
+            return Ok(Value::from_f64(agent, consts::PI));
```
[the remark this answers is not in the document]
eliassjogreen: See https://github.com/trynova/nova/pull/318#discussion_r1683265386

aapoalas (approved):

PR #319, opened 2024-07-16
feat(ecmascript): Implement strict vs. non-strict code distinction

andreubotella:
Whenever the spec said to have different behaviors for strict and non-strict code, we previously sometimes went with the strict and sometimes with the loose behavior, because we weren't tracking whether the current execution was strict or not. This patch tracks that.

Ideally, oxc would store whether a function or not is strict code (i.e. whether it has a `"use strict"` declaration or its lexical parent was strict code), since it needs to have that information in order to properly parse it. However, it only exposes whether it has the `"use strict"` declaration. So we track whether the current code is strict mode in `ECMAScriptCodeEvaluationState`. This also lets us use that information for direct `eval`.

We don't currently implement mapped `arguments` objects, which are needed for non-strict mode function. In order to avoid blocking this patch on implementing that, or having most non-strict functions crash (since we currently create an `arguments` object for every function), this patch uses unmapped `arguments` objects always. Mapped `arguments` objects are left for later.

This patch also changes `OrdinaryFunctionCreateParams`'s `this_mode` parameter, which used to be the `ThisMode` enum, into a `lexical_this` bool, which better reflects the spec's distinction between the "lexical this" and "non lexical this" states. We now have the information to translate that into the `ThisMode` enum inside the `ordinary_function_create` function.

When running code with nova_cli, by default it will be run as strict mode, unless the `--no-strict` flag is passed. As for test262, most tests must be run twice, once in strict and once in loose mode, to verify that the behavior does not change, and so we run both modes, failing as soon as one of them fails.


nova_vm/src/ecmascript/scripts_and_modules/script.rs
```diff
@@ -181,9 +181,11 @@ pub fn parse_script(
     source_text: Box<str>,
     realm: RealmIdentifier,
     host_defined: Option<HostDefined>,
+    strict_mode: bool,
```
aapoalas, line 184: **nitpick:** IMO HostDefined should be last argument.

tests/test262_runner.rs
aapoalas, line 1: **suggestion:** For the "strict or loose, either goes" tests, how about we just run the test in both modes and make sure it passes in both?

aapoalas (approved):
LGTM <3

nova_vm/src/ecmascript/scripts_and_modules/script.rs
andreubotella: Done.

tests/test262_runner.rs
andreubotella: Done.

PR #322, opened 2024-07-17
fix(vm): Missing `GetValue` for property key expressions in object literals

andreubotella:


nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -971,6 +971,11 @@ impl CompileEvaluation for ast::ObjectExpression<'_> {
                         | ast::PropertyKey::TSNonNullExpression(_)
                         | ast::PropertyKey::TSInstantiationExpression(_) => unreachable!(),
                     }
+                    if let Some(prop_key_expression) = prop.key.as_expression() {
+                        if is_reference(prop_key_expression) {
```
aapoalas, line 975: **issue:** `PropertyKey::Super` above includes a `GetValue` in it directly.

This `as_expression` is something I didn't know about and is quite a bit cleaner (although maybe a tad bit less efficient). Elsewhere I've manually added the `GetValue` instruction additions to the match paths.

So, either the Super case needs to have the GetValue "call" removed so it relies on only this, or alternatively you can manually inline this into every reference-like PropertyKey match path :)

andreubotella: As far as I can tell it's not valid syntax to have `super` as a property key, oxc just doesn't distinguish that in their AST types. If you're okay with that, I'll change the super match to `unreachable!()`.

aapoalas: Ah, the `PropertyKey::Super` is inherited from member expression, ie. calculated property key. So it's eg. `this[super]`. So it's dumb but reachable.

andreubotella: > Ah, the `PropertyKey::Super` is inherited from member expression, ie. calculated property key. So it's eg. `this[super]`. So it's dumb but reachable.

If I'm reading the spec correctly, `super` by itself is never allowed, only `super[expr]`, `super.ident` or `super()`.

aapoalas: Oh, very true, it was I who was dumb and it should indeed be unreachable :)

PR #323, opened 2024-07-17
feat: repl command

load1n9:


nova_cli/src/main.rs
```diff
@@ -158,72 +165,53 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-
-    // `print` function
-    fn print(agent: &mut Agent, _this: Value, args: ArgumentsList) -> JsResult<Value> {
-        if args.len() == 0 {
-            println!();
-        } else {
-            println!("{}", args[0].to_string(agent)?.as_str(agent));
+            loop {
+                let mut input = String::new();
+                std::io::stdin().read_line(&mut input)?;
```
andreubotella, line 192: Lines should have a prompt, right? Maybe:
```suggestion
                print!("> ");
                std::io::stdin().read_line(&mut input)?;
```

andreubotella (approved):
LGTM. This wouldn't properly handle `let` or `const` variable definitions, because each line runs in its own script. But this is good for a start and can be tweaked later.

nova_cli/src/main.rs
load1n9: I was testing that and it was double printing after errors so I'll look into that after

PR #324, opened 2024-07-18
refactor(vm): Implement a SmallF64 variant to replace Float

aapoalas:


nova_vm/src/engine/small_double.rs
```diff
@@ -0,0 +1,134 @@
...
+        };
+
+        Self { data }
+    }
+}
+
+impl TryFrom<f64> for SmallF64 {
+    type Error = ();
+    fn try_from(value: f64) -> Result<Self, Self::Error> {
+        if SmallF64::can_convert(value) {
```
andreubotella, line 56: **question:** `f64::NAN` has its trailing byte as zero, but not all possible f64 NaN values do. If we don't care about preserving the bits of NaN values, we should special-case NaN in this conversion.

nova_vm/src/engine/small_double.rs
```diff
@@ -0,0 +1,134 @@
...
+        // The enum repr zero is the first byte
+        if cfg!(target_endian = "little") {
+            // In little endian that is the least significant byte, which is
+            // exactly where we want it.
+            f64::from_bits(number)
+        } else {
+            // In big endian that is the most significant byte. We want to shift
+            // up one byte to "pop off" the zero and bring a zero byte in to
+            // the least significant byte.
+            f64::from_bits(number << 8)
```
andreubotella, line 93: We should be running tests for these endian-dependent conversions [in Miri using cross-interpretation](https://github.com/rust-lang/miri?tab=readme-ov-file#cross-interpretation-running-for-different-targets).

andreubotella (dismissed):

nova_vm/src/engine/small_double.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    #[inline]
+    pub fn into_f64(self) -> f64 {
+        self.into()
+    }
+
+    #[inline(always)]
+    fn can_convert(value: f64) -> bool {
+        value.to_bits().trailing_zeros() >= 8
```
andreubotella, line 31: Should this also check that `value` is not a valid `SmallInteger`?

nova_vm/src/engine/small_double.rs
aapoalas: Yeah, it's... probably a good idea. I've been kind of thinking that we might allow passing magic data in NaNs but it's honestly a stupid thing to do, so probably best to not support it.

nova_vm/src/engine/small_double.rs
aapoalas: Oooh, nice!

nova_vm/src/engine/small_double.rs
aapoalas: Yeah, sounds like a good idea.

nova_vm/src/engine/small_double.rs
aapoalas: Created #342

PR #327, opened 2024-07-19
feat: treat typescript as comments

load1n9:


aapoalas (changes requested):
This should definitely be a build flag.

load1n9:
cli flag?

aapoalas:
A cargo feature.

eliassjogreen:
Once https://github.com/tc39/proposal-type-annotations is finalized this will basically be the implementation but until then I agree that it should be behind a flag.

PR #329, opened 2024-07-19
feat(vm): Semantic analysis

aapoalas:
So we had been living under a misunderstanding of sorts. The OXC parser doesn't really yet guarantee much about the validity of the AST. We need to run `oxc_semantic` on it to check it.

This PR just adds in the semantic checking. It does not use any of that data. Probably a lot of that would be quite helpful in eg. script instantiation etc.


aapoalas:
Oh yeah and this doesn't actually really change parsing performance in a very measurable way. Before this, running test262 took 17 seconds on my machine. After this it between 17 and 20 seconds.

andreubotella (approved):

PR #331, opened 2024-07-20
feat: Support `toLowerCase` & `toUpperCase` in StringPrototype

marc2332:


aapoalas (changes requested):
Code looks good to me. You'll need to update the test262 expectations though.

You can to do that by first running `cargo build --profile release` to get an optimised build artifact, then run `cargo run --profile release test262 -- -u` if memory serves. That will run the tests and update expectations.

Thank you and welcome to the contributors list! <3

aapoalas:
Oh, that should probably be `--bin test262` maybe.

But yeah, the CI run finished so you can see how the CI machinery tests that we pass the tests that are not mentioned in the expectations file, and fail/crash/timeout those mentioned in the file in the expected way. (Though actually we consider crash and failure to be equivalent, currently.)

That's our main regression testing method at present. We also have `cargo test` which runs some simple tests but those are not very comprehensive or good tests.

aapoalas (dismissed):

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -781,12 +781,26 @@ impl StringPrototype {
...
 
-    fn to_lower_case(_agent: &mut Agent, _this_value: Value, _: ArgumentsList) -> JsResult<Value> {
-        todo!()
+    fn to_lower_case(agent: &mut Agent, this_value: Value, _: ArgumentsList) -> JsResult<Value> {
+        // 1. Let O be ? RequireObjectCoercible(this value).
+        let o = require_object_coercible(agent, this_value)?;
+        // 2. Let S be ? ToString(O).
+        let s = to_string(agent, o)?;
+
+        // 3. Return the lower case version of the string
```
andreubotella, line 790:
```suggestion
        // 3. Let sText be [StringToCodePoints](https://tc39.es/ecma262/#sec-stringtocodepoints)(S).
        // 4. Let lowerText be toLowercase(sText), according to the Unicode Default Case Conversion algorithm.
        // 5. Let L be [CodePointsToString](https://tc39.es/ecma262/#sec-codepointstostring)(lowerText).
        // 6. Return L.
```

The comments should reflect the spec: https://tc39.es/ecma262/#sec-string.prototype.tolowercase

Now, I'm not sure whether `str::to_lowercase` in Rust reflects the Unicode Default Case Conversion. The Rust docs say "‘Lowercase’ is defined according to the terms of the Unicode Derived Core Property `Lowercase`", which might not be equivalent, since there might be different case conversion algorithm with different behavior for some writing scripts. So I would also add a note in the comments saying that this might not reflect the spec.

tests/expectations.json
```diff
@@ -31183,6 +31149,7 @@
   "language/white-space/string-nbsp.js": "FAIL",
   "language/white-space/string-space.js": "FAIL",
   "language/white-space/string-vertical-tab.js": "FAIL",
+  "language\\expressions\\object\\11.1.5-0-1.js": "FAIL",
```
andreubotella, line 31152:
```suggestion
  "language/expressions/object/11.1.5-0-1.js": "FAIL",
```

Ugh, Windows file paths. I'll have to fix the test262 runner to not do this.

aapoalas (approved):
Thank you <3 and great work

PR #333, opened 2024-07-20
fix: Canonicalize the base path of tests in test262_runner

marc2332:
This caused some issues on Windows as [`PathBuf::canonicalize`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.canonicalize) adds a `\\\\` in Windows Paths, causing `starts_with` to fail if the other path isn't canonicalized as well.

Solution? Just canonicalize the base tests path.


aapoalas (approved):
LGTM!

PR #339, opened 2024-07-21
feat(ecmascript): `Date.prototype[Symbol.toPrimitive]`

aapoalas:
Extracted from #335


tests/expectations.json
```diff
@@ -20741,6 +20727,7 @@
   "language/expressions/new/spread-sngl-iter.js": "CRASH",
   "language/expressions/new/spread-sngl-literal.js": "CRASH",
   "language/expressions/new/spread-sngl-obj-ident.js": "CRASH",
+  "language/expressions/object/11.1.5-0-1.js": "PASS",
```
andreubotella, line 20730: **nitpick:** Maybe remove? This expectation was added in #331, but it was added in the wrong place (not in alphabetical order), and `-u` reordered it here. But it's a pass expectation, so if this PR is already changing it, it'd be better to remove it.

andreubotella (dismissed):

andreubotella (approved):

PR #343, opened 2024-07-21
fix(heap): Keep parsed source code in the heap

### LINKED ISSUE #334: bug: Unable to resolve variables
The following code fails sometimes, not always!:

```js
var s1 = "In getter";
var s2 = "In setter";
var s3 = "Modified by setter";
var o;
eval("o = { get foo(){ return s1;}, set foo(arg){return s2 = s3} };");

function eatFoo(a, b){
  if (a === b) {
    // Handle +/-0 vs. -/+0
    return a !== 0 || 1 / a === 1 / b;
  }
}

eatFoo(o.foo, s1);

o.foo=10;

print(o.foo)
```

platform: Windows

aapoalas:
Fix #334 and moves Allocator creation to be a heap-internal matter.


nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -158,6 +159,24 @@ pub fn perform_eval(
...
+            // Add 7 whitespace bytes to the end of the eval string. This should
+            // guarantee that the string gets heap-allocated, no question.
+            let data = format!("{}       ", x.as_str());
+            let heap_string_x = String::from_string(agent, data);
+            let String::String(x) = heap_string_x else {
+                unreachable!()
+            };
+            x
+        }
+    };
```
andreubotella, line 178: **issue:** Off-by-one-error: It's not actually true that adding 7 whitespace bytes will guarantee the string is heap-allocated, you're forgetting the empty string.

**question:** Wouldn't it be better to transmute the string to `&'static str` in this step, and then slice it for small strings so the whitespace isn't parsed?

nova_vm/src/heap.rs
```diff
@@ -182,6 +182,11 @@ impl Heap {
             weak_maps: Vec::with_capacity(0),
             weak_refs: Vec::with_capacity(0),
             weak_sets: Vec::with_capacity(0),
+            // Drop scripts, modules, and eval sources last to ensure that all
+            // objects referring to them have dropped first.
+            eval_sources: Vec::with_capacity(0),
+            modules: Vec::with_capacity(0),
+            scripts: Vec::with_capacity(1),
```
andreubotella, line 189: **issue:** Changing the order of fields when constructing `Heap` wouldn't do anything, right? I think you want to change them in the struct definition.

nova_vm/src/ecmascript/scripts_and_modules/eval_source.rs
```diff
@@ -0,0 +1,113 @@
...
+    ecmascript::{execution::Agent, types::HeapString},
+    heap::{
+        indexes::BaseIndex, CompactionLists, CreateHeapData, Heap, HeapMarkAndSweep, WorkQueues,
+    },
+};
+
+type EvalSourceIndex = BaseIndex<EvalSourceHeapData>;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct EvalSource(EvalSourceIndex);
```
andreubotella, line 24: **question:** Is the plan to eventually extend `EvalSource` to scripts as well, so scripts can be GC'd if they're not referenced by anything? In that case, we should probably name this `ScriptSource` or something like that.

tests/expectations.json
```diff
@@ -20722,7 +20722,6 @@
   "language/expressions/new/spread-sngl-literal.js": "CRASH",
   "language/expressions/new/spread-sngl-obj-ident.js": "CRASH",
   "language/expressions/object/11.1.5-0-1.js": "PASS",
```
andreubotella, line 20724:
```suggestion
```

Somehow this passing test got added to the expectations again. Since the original failure was caused by this eval bug, it makes sense to remove it here as well.

nova_vm/src/ecmascript/builtins/global_object.rs
aapoalas: Haha, true :D

Hmm... Probably the extra whitespace shouldn't really matter. But yeah, micro-opt wise could definite slice it.

nova_vm/src/heap.rs
aapoalas: Oops, yeah.

nova_vm/src/ecmascript/scripts_and_modules/eval_source.rs
aapoalas: Scripts already can be GC'd if they're not referenced if no functions reference them.

Scripts have quite a bit more stuff going on with them. I _think_ the way forward with this sort of setup for scripts would be to just add these equivalent fields to ScriptHeapData.

So in effect: This would stay as eval-source only.

tests/expectations.json
aapoalas: Fuuck :D

nova_vm/src/ecmascript/scripts_and_modules/eval_source.rs
andreubotella: In test262 there's `$262.evalScript()`, which IIUC should create a script record from a JS string source. In HTML you can also create a script from a JS string. How does that play into all of this?

aapoalas: Oh, ugh, oof.

I mean, I guess it mostly wouldn't be any different from calling `parse_script` and all that stuff, except that we'd move creating the `Allocator` inside that method entirely and keep it in the `ScriptHeapData` similarly to how we keep it for eval'd data in `EvalSource`. The `parse_script` would also probably change from taking a `Box<str>` to taking a `std::string::String` or an alternative API for taking `&'static str`, or maybe just going broke and taking a `nova_vm::String` directly? And then we'd keep a reference to that inside the `ScriptHeapData`, replacing the current `Box<str>` there.

aapoalas: Yes, that's what I'm talking about in the latter half. The allocator would move inside the heap similarly to how it does here for eval calls. It would also take a JS String as the parameter.

If you want, I can use this PR to change the script running API as well so that it handles the Script Allocator creation inside of the heap.

andreubotella: That doesn't need to be in this PR.

aapoalas: Alright, I'll open a new PR for that.

eliassjogreen (approved):

PR #344, opened 2024-07-21
refactor(cli): Refactor execution & microtasks into `CliRunner` struct

andreubotella:
In order to add a repl to the cli, #323 moved the `initialize_global_object` and `exit_with_parse_errors` functions into a new `helper` module. However, this still left duplicate code between the `eval` and `repl` commands.

This patch moves the isolate and realm initialization, as well as the script execution and microtask checkpoint, to a new `CliRunner` struct that lives in the `helper` module.

Aside from just refactoring, this patch makes two visible changes related to promise jobs:
  - When passing multiple files to `eval`, now promise jobs run after each script execution, rather than at the end.
  - In `repl` mode it used to be that promise jobs were never run. Now, all promise jobs enqueued when running a line of JS are run before the result is printed.

### DIFF
--- a/nova_cli/Cargo.toml
+++ b/nova_cli/Cargo.toml
@@ -10,6 +10,7 @@ cliclack = { workspace = true }
 console = { workspace = true }
 miette = { workspace = true }
 nova_vm = { path = "../nova_vm" }
+oxc_allocator = { workspace = true }
 oxc_ast = { workspace = true }
 oxc_parser = { workspace = true }
 oxc_semantic = { workspace = true }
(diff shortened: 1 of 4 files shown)


aapoalas:
I'd kind of want to land #268 before this. Making GC work would be kind of nice, even if the API is pretty ugly.

aapoalas (approved):
LGTM: We could merge this if you have the time to solve the merge conflicts?

andreubotella:
> LGTM: We could merge this if you have the time to solve the merge conflicts?

Can't right now, but maybe in a couple hours.

aapoalas:
> > LGTM: We could merge this if you have the time to solve the merge conflicts?
> 
> Can't right now, but maybe in a couple hours.

Absolutely no rush <3

PR #345, opened 2024-07-22
feat(vm): Handle script data on the heap

aapoalas:
Built on top of #343.

Note: MIRI gives this the green light with (a good number of) `cargo test` tests cases. I didn't run all of them due to lack of time.


aapoalas:
@andreubotella Main question: Looking at this, would you prefer that scripts actually refer to an `EvalSource` instead of holding the equivalent data internally?

andreubotella:
> @andreubotella Main question: Looking at this, would you prefer that scripts actually refer to an `EvalSource` (then renamed to eg. `SourceCode`) instead of holding the equivalent data internally?

I haven't had the chance to look at all of this PR and #343, but yes, that seems like better code organization.

aapoalas:
> I haven't had the chance to look at all of this PR and #343, but yes, that seems like better code organization.

The only thing that makes me sad about that is that then each script always creates two "heap objects", when it only really needs to create one. So there's a bit more tracing and sweeping to be done.

But, scripts are rare things in general so you're right: The code organization is way better.

aapoalas:
Subsumed by the original #343

PR #346, opened 2024-07-22
feat(ecmascript): implement `iterator_step_value`

eliassjogreen:


tests/expectations.json
```diff
@@ -29487,4 +29460,4 @@
   "staging/set-methods/set-intersect-other-is-set-like.js": "CRASH",
   "staging/set-methods/set-intersection-other-is-map.js": "CRASH",
   "staging/set-methods/set-intersection-other-is-set.js": "CRASH"
-}
\ No newline at end of file
+}
```
eliassjogreen, line 29463:
```suggestion
}```

File: nova_vm/src/ecmascript/builtins/keyed_collections/set_objects/set_constructor.rs [hunk 2]
```diff
@@ -107,11 +107,11 @@ impl SetConstructor {
             }
         }
         // 7. Let iteratorRecord be ? GetIterator(iterable, SYNC).
-        let iterator_record = get_iterator(agent, iterable, false)?;
+        let mut iterator_record = get_iterator(agent, iterable, false)?;
```
aapoalas, line 110: **thought:** We should probably split sync and async iterator creator functions from one another. There's some code duplication but the late bool check inside the functions hardly makes sense.

aapoalas (approved):
Looks good to me! Very nice as well, thank you <3

PR #347, opened 2024-07-22
feat: Built-in functions with Storage

marc2332:
Notes:
- I had to manually mark the storage via unsafe as `Send` & `Sync` because the garbage collector will iterate over the builtin functions in a separate thread. Not sure if there is a better way. I could maybe find a map similar to AnyMap that is also thread-safe. Or maybe just build one, like Deno has done.
- this feature is going to be useful for functions that might want to storage certain values (aka resources) between calls, e.g opening a fs File).
- Because the storage cannot be `Copy` I had to remove the Copy derive from `Behaviour` and adapt two places accordingly.


marc2332:
I also thught about maybe just storing an index key in `Behavior::RegularWithStorage`, this would probably remove the need for the unsafe Send and Sync, and also make `Behavior` copyable again. But of course, where do you save the actual storages then? The heap maybe? Thoughts?

aapoalas (changes requested):
Unfortunately I am not in favour of this. This likely increases the size of the Behaviour enum massively which then makes all built-in strings less cache efficient even though they do not carry any storage.

If this was truly a must have feature, the way it would be done is to add a HashMap on the side of the built-in functions vector that acts as a "sparse vector": It only contains values for those functions that actually carry storage.

However, the way this should be done, I believe, is through embedder objects. An embedder object will be free to hold a pointer to some embedder data (together with a vtable pointer). These builtins would then be methods on that embedder object, of alternatively the embedder object itself could be the function.

marc2332:
> Unfortunately I am not in favour of this. This likely increases the size of the Behaviour enum massively which then makes all built-in strings less cache efficient even though they do not carry any storage.

It certainly increases the size, perhaps not massively, but it really does jump the size from 16bytes to 24bytes, which a large scale obviously hurts more. But in any case, I do agree, I must find a better way.

> If this was truly a must have feature, the way it would be done is to add a HashMap on the side of the built-in functions vector that acts as a "sparse vector": It only contains values for those functions that actually carry storage.

That sounds interesting. Seems pretty similar to what I had on my mind in https://github.com/trynova/nova/pull/347#issuecomment-2243856219, although it will still require Send and Sync now that I think about it.

> However, the way this should be done, I believe, is through embedder objects. An embedder object will be free to hold a pointer to some embedder data (together with a vtable pointer). These builtins would then be methods on that embedder object, of alternatively the embedder object itself could be the function.

Not so sure about this one, I don't know anything about these embedder objects thing yet.

marc2332:
I could also just add a new variant of `Behavior` to support boxed closures instead of `fn` pointers

this is some WIP code but shows my idea:

```rs
#[derive(Clone)]
struct RegularBoxedFn(Arc<Box<dyn Fn(&mut Agent, Value, ArgumentsList<'_>) -> JsResult<Value> + Send + Sync>>);

impl std::fmt::Debug for RegularBoxedFn {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.write_str("fn") // todo
    }
}

impl PartialEq for RegularBoxedFn {
    fn eq(&self, other: &Self) -> bool {
        Arc::ptr_eq(&self.0, &other.0)
    }
}

impl Eq for RegularBoxed {}

#[derive(Debug, Clone, PartialEq, Eq)]
pub enum Behaviour {
    Regular(RegularFn),
    RegularBoxed(RegularBoxedFn),
    Constructor(ConstructorFn),
}
```

benefits:
- no need to do any storage implementation in the VM at all as the closures can simply capture the state
- the `Behavior` enum size doesn't increase at all

what do you think about this @aapoalas?

aapoalas:
I'm... ambivalent on that.

It definitely does not increase the enum size, so that's good. This would not be pessimising all builtin functions. But I am not sure if this is a great approach either.

Looking at V8, they have a few different ways to handle something like this:
1. Objects can be given "embedder slots", which grows the size of the object. The number of these slots is, I believe, decided before building the object and tested at runtime by reading the object's header data. Your builtin function would then either be a method on this object and would test its `this` parameter to contain this embedder slot and extra the data from there, or it would take the object as a parameter and test that.
2. Builtin functions have a `data` pointer that can have a pointer embedded into it on a per-function basis.

Embedders (runtimes like Deno) have a third way as well:
3. They pass a plain number through the JS engine, and keep an outside table/hashmap of items. You could call this a "handle" approach. (In V8 it's also possible to hide the handle value inside an object in a way that it cannot be observed or mutated from within JS.)

For V8, avoiding a closure is fairly important: The closure call API would be pretty nasty to implement in JIT-land. That's not directly an issue for us, we do not have a JIT nor will have one in the near future. But it's one of those things that does make me think: If this is done and we decide we want a JIT somewhere down the line, this will basically block it. (It's possible to sidestep this manually with some `Arc<BuiltinFunctionClosure>` that can be "called" with a statically known function but that has to be done outside of normal Rust closures.)

Another issue with the closure approach is that the closure cannot capture any JS data. There is no way (that I am aware of) of adding a `HeapMarkAndSweep` trait impl on a closure. eg. Consider a builtin function that holds a closure that starts a Future and resolves a Promise after it is done: This cannot be done in a way where the closure would both keep the Promise alive (stop it from being garbage collected) _and_ would allow the Promise to die (be garbage collected) if the function itself gets garbage collected. The Promise could be kept alive by "rooting" it (once that's enabled, probably today :D ) but then it would leak if the function itself is dropped.

The "data pointer" approach pessimises the common case (no data pointer) so I'm actively opposed to that. Objects with embedder slots I'm... not actively against but it's not a high priority either. Embedder objects I am fairly happy to implement (they're already in the heap as well, they just lack any contents) or see implemented.

Finally: For the time being, I am kind of most in favour of the dumb-but-simple "handle" approach. Right now the only problem of course is that you don't really have a way to get from the `&mut Agent` to your `HostHooks` stuff which would probably contain your table / hashmap of handle-to-data stuff. But that is fairly easy to fix: In `Agent` add a method like `get_host_data` which returns an `Any`, and add a method to HostHooks that gets that `Any` from the embedder's HostHooks.

With this you can now set up data within your HostHooks that you can access in a BuiltinFunction via the `&mut Agent` reference. This also means that you don't have to create a new closure and a new BuiltinFunction and whatnot on every call to eg. `readFile`. This should generally be much more performant.

nova_vm/src/ecmascript/execution/agent.rs
```diff
@@ -189,6 +193,10 @@ impl Agent {
     pub(crate) fn running_execution_context_mut(&mut self) -> &mut ExecutionContext {
         self.execution_context_stack.last_mut().unwrap()
     }
+
+    pub fn get_hoost_hooks(&self) -> &'static dyn HostHooks {
```
aapoalas, line 197: **nitpick (blocking):** It's probably not a good idea to expose the entire host hooks through the Agent. eg. It probably shouldn't be possible to call the uncaught rejection handler.

So maybe just do a `get_host_storage` or something?

nova_vm/src/ecmascript/execution/default_host_hooks.rs
```diff
@@ -29,4 +29,8 @@ impl HostHooks for DefaultHostHooks {
     fn enqueue_promise_job(&self, _job: Job) {
         // No-op
     }
+
+    fn get_storage_handle(&self) -> std::cell::RefMut<anymap::AnyMap> {
```
aapoalas, line 33: **nitpick:** Is this actually a "handle"? Wouldn't that depend on the host? Also I'd suggest making this just an `Any` instead of `AnyMap`.

Likely the embedder knows exactly what sort of stuff they're storing, so they'll likely have just some `struct MyHostStorage` for efficient access of sub-parts.

**suggestion:** I guess maybe it might be a good option to make this return an `Option` and then make the default implementation just always return `None`.

marc2332: > suggestion: I guess maybe it might be a good option to make this return an Option and then make the default implementation just always return None.

I actually thought about doing this, but I thought it would just make every function that  reads the handler unnecessarily more verbose for little benefit

marc2332: > Wouldn't that depend on the host

I might as well just call it `get_host_data` to avoid assuming it is just for storage then, wdyt?

aapoalas: Yeah, that's true. Not sure which one I prefer, really.

aapoalas: > I might as well just call it `get_host_data` to avoid assuming it is just for storage then, wdyt?

Yeah. Please add a doc comment (`///`) in the method that it panics if the host has not overridden the default hook.

nova_cli/Cargo.toml
```diff
@@ -14,4 +14,4 @@ oxc_ast = { workspace = true }
 oxc_parser = { workspace = true }
 oxc_semantic = { workspace = true }
 oxc_span = { workspace = true }
-oxc_diagnostics = { workspace = true }
+oxc_diagnostics = { workspace = true }
```
aapoalas, line 17: **nitpick:** You lost the trailing newline.

aapoalas (dismissed):
Yes, now this I can get behind of :)

aapoalas (approved):

marc2332:
> Yes, now this I can get behind of :)

Nice! Really appreciate all the help and reviews. We got a working version that is also very flexible, with no performance regression for existing code and that is very simple.

As a sidenote, I was able to implement the storage thing in Andromeda by using this PR new API. (Still working on the API, but this is the idea more or less)

```rs
#[derive(Default)]
    pub struct DemoExt;

    impl Ext for DemoExt {
        fn load(&self, mut loader: ExtLoader) {
            loader.init_storage(45.0f32);
            loader.load_op("more", Self::more, 0);
            loader.load_op("less", Self::less, 0);
        }
    }

    impl DemoExt {
        fn more(agent: &mut Agent, _this: Value, _args: ArgumentsList) -> JsResult<Value> {
            let host_data: &RefCell<AnyMap> = agent.get_host_data().downcast_ref().unwrap();
            let mut storage = host_data.borrow_mut();
            let state = storage.get_mut::<f32>().unwrap();
            *state += 1.0;
            Ok(Value::SmallF64(SmallF64::from(*state)))
        }
        fn less(agent: &mut Agent, _this: Value, _args: ArgumentsList) -> JsResult<Value> {
            let host_data: &RefCell<AnyMap> = agent.get_host_data().downcast_ref().unwrap();
            let mut storage = host_data.borrow_mut();
            let state = storage.get_mut::<f32>().unwrap();
            *state += 1.0;
            Ok(Value::SmallF64(SmallF64::from(*state)))
        }
    }
```

PR #348, opened 2024-07-24
feat(vm): Implement `Global` to enable rooting of Values

marc2332:
⚠️  Depends on https://github.com/trynova/nova/pull/347

- New `GlobalValue` struct that allows to hold the index of a heap value even across multiple GC calls. This allows things like return a promise in a built-in function and creating a `GlobalValue` based on it, so later (potentially a lot later) it can be used to resolve the actual value and thus resolve the promise. This is thanks to the fact that global values are not moved after CG, so this makes it possible to hold a position index of this vector across time.
- I also made public some types related to Promises for third-party crates to use, instead of having them as `pub(crate)`, this is completely unrelated to the main change of this PR (`GlobalValue`) but it was needed for my testing of it. I would understand if you didn't want it in this PR, and if that was the case I would happily revert those changes in this PR and open a new one just making public the types I need.
- I needed the changes I made https://github.com/trynova/nova/pull/347 in order to fully make  and test this PR, hence it includes those changes as well, because the #347 branch is not pushed in this repo I could not make GitHub use it as a target branch so it could show only the actual changes I made today, but in any case, you can see the changes from the 295837e02763a142b20376650429ca2fe27f69ae commit forward

### DIFF
--- a/nova_vm/src/ecmascript/builtins.rs
+++ b/nova_vm/src/ecmascript/builtins.rs
@@ -54,4 +54,5 @@ pub use builtin_function::{
     BuiltinFunctionArgs, BuiltinGetter, ConstructorFn, RegularFn as JsFunction, RegularFn,
 };
 pub(crate) use builtin_function::{BuiltinIntrinsic, BuiltinIntrinsicConstructor};
+pub use control_abstraction_objects::*;
 pub(crate) use ecmascript_function::*;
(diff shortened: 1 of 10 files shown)


marc2332:
Probably worth adding a test to make sure the global values are released correctly from the heap and that slots are reused

load1n9 (approved):

aapoalas (approved):

PR #349, opened 2024-07-25
feat: Let runtimes handle imports

marc2332:
DEPENDS ON https://github.com/trynova/nova/pull/348 (sorry 💀 )

You can see the changes from https://github.com/trynova/nova/pull/349/commits/36d282662f9c703ec55eab8efc6511baab14b35d forward


nova_vm/src/ecmascript/execution/agent.rs
```diff
@@ -122,6 +124,20 @@ pub trait HostHooks: std::fmt::Debug {
     ) {
         // The default implementation of HostPromiseRejectionTracker is to return unused.
     }
+
+    /// Handle import declarations.
+    ///
+    /// Note: This will panic if not implemented manually.
+    fn import_module(&self, _import: &ast::ImportDeclaration<'_>, _agent: &mut Agent) {
+        unimplemented!()
```
marc2332, line 132: This should probably receive an &mut Executable or return something so the imported values can get injected.

PR #351, opened 2024-07-26
feat(vm): Migrate to `ahash`

marc2332:
Switches to https://crates.io/crates/ahash for faster performance but still resistant against DOS attacks:

code:

```js
for(let i = 0; i < 100000; i++){
    //console.log(i)
}
```

Before (std):
image

After (ahash):
image

results in 40-20ms less to run the same code.


marc2332:
open question: Could having a non-cryptographic algorithm be exploited in nova? It is used by the Rust compiler and Oxc too.

aapoalas:
> open question: Could having a non-cryptographic algorithm be exploited in nova? It is used by the Rust compiler and Oxc too.

Probably yes. That does go for rustc and oxc likewise, but their threat model is mainly "running on developer machines and/or CI with mostly trusted code".

Nova's threat model is not very well defined but if eg. you write an HTTP server in Nova and store some client-controlled data about requests in a Map or Set (say the UserAgent?), now you're susceptible to a DDOS attack caused by a targeted attack on the hashing algorithm.

eliassjogreen:
> Nova's threat model is not very well defined but if eg. you write an HTTP server in Nova and store some client-controlled data about requests in a Map or Set (say the UserAgent?), now you're susceptible to a DDOS attack caused by a targeted attack on the hashing algorithm.

It seems like (after a bit of googling) V8 (and perhaps other engines?) use a seed generated at startup to salt their hashes making Hash-DOS attacks harder/impossible, although apparently still allowing for certain timing attacks. For client side execution (e.g. browsers) I would think it is ok to just use the fastest hashing algo as it's not really a security issue, server side execution though should provide protection against Hash-DOS attacks. So my suggestion to make it secure (_in the future?_) is to salt with a random seed, and if it worsens performance I suggest having it behind a build flag: `secure` which could be enabled for server side environment while being disabled for client side environments where we care more about performance than security.

eliassjogreen (approved):

.gitignore
```diff
@@ -1,3 +1,4 @@
 target
 Cargo.lock
-/.vscode
\ No newline at end of file
+/.vscode
+nova_vm/src/heap/heap_bits.rs
```
eliassjogreen, line 4:
```suggestion
```

aapoalas (changes requested):
Definitely can't gitignore heap bits file.

.gitignore
marc2332: Wooooooot? I did not put this in here 💀 I swear...

aapoalas:
And then just gotta resolve conflict

marc2332:
I think I might have dragged the file accidentally with the touchpad (I know, touchpad 💀 )

.gitignore
eliassjogreen: :suspect:

aapoalas (approved):

eliassjogreen:
@marc2332 Lets also remember to change the PR title to `ahash` and not `rustc-hash`.

marc2332:
> @marc2332 Lets also remember to change the PR title to `ahash` and not `rustc-hash`.

Done, also updated the description. Only thing still referring to rustc-hash is the branch name which I can't do much about it without reopening the PR again 😢

PR #352, opened 2024-07-26
feat(ecmascript): String prototype `padStart` and `padEnd`

eliassjogreen:
Got some issues with surrogate strings in `built-ins/String/prototype/padEnd/normal-operation.js` and `built-ins/String/prototype/padStart/normal-operation.js`:
```js
print('abc'.padEnd(6, '\uD83D\uDCA9') === 'abc\uD83D\uDCA9\uD83D');
print('abc'.padStart(6, '\uD83D\uDCA9') === '\uD83D\uDCA9\uD83Dabc');
```

**Update:** We will need to deal with UTF-16 surrogates in lots of places: https://github.com/trynova/nova/blob/main/small_string/lib.rs#L40 so I am leaving this PR as is with a `TODO` comment in the `string_pad` function.


nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -243,6 +245,113 @@ impl Builtin for StringPrototypeIterator {
     const BEHAVIOUR: Behaviour = Behaviour::Regular(StringPrototype::iterator);
 }
 
+/// ### [22.1.3.17.1 StringPaddingBuiltinsImpl ( O, maxLength, fillString, placement )](https://tc39.es/ecma262/#sec-stringpaddingbuiltinsimpl)
```
aapoalas, line 248: **nitpick:** I've usually put these sorts of "helper" functions to the bottom of the file, after the impls and so on.

aapoalas (dismissed):
LGTM

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -243,6 +245,113 @@ impl Builtin for StringPrototypeIterator {
...
+        vec
+    } else if fill_len % fill_string_length == 0 {
+        let fill_count = (fill_len / fill_string_length) as usize;
+        let mut vec = VecDeque::with_capacity(fill_count + 1);
+        vec.extend(repeat(fill_string).take(fill_count));
+        vec
+    } else if fill_len < fill_string_length {
+        let mut vec = VecDeque::with_capacity(2);
+        // TODO: Deal with surrogates.
+        let sub_string = &fill_string.as_str(agent)[..fill_len as usize];
```
aapoalas: **thought:** This is actually probably not quite correct. The length values are UTF-16 indexes but then we're indexing into a UTF-8 string with those. These should probably go through the index mapping.

eliassjogreen: Pushed what I think is a fix for this, doesn't really change any expectations but now the crashes point to https://github.com/trynova/nova/blob/main/small_string/lib.rs#L40 which is more correct.

Pretty messy code which might be cleaned up by some iterator magic but I couldn't figure it out.

nova_vm/src/builtin_strings
```diff
@@ -444,4 +445,4 @@ WeakSet
 with
 withResolvers
 writable
-xor
\ No newline at end of file
+xor
```
aapoalas, line 448: **praise:** Nice <3

aapoalas (approved):
Thank you, good job implementing quite a complicated algorithm (due to UTF-16/8 fighting) <3

aapoalas (changes requested):
Requires new updates on expectations unfortunately :(

eliassjogreen:
@aapoalas Fixed the merge and formatted correctly + regenerated the expectations. Should be able to merge it after CI runs :)

aapoalas:
> @aapoalas Fixed the merge and formatted correctly + regenerated the expectations. Should be able to merge it after CI runs :)

Oops, we both did it 1 minute apart :D

eliassjogreen:
Yeah I realized that haha. I pushed but then when I went to the web to write the message I see that you pushed literally seconds after me. Great minds think alike

aapoalas (approved):

PR #354, opened 2024-07-27
feat(ecmascript): String prototype `repeat`

eliassjogreen:


tests/skip.json
```diff
@@ -19,6 +19,7 @@
     "built-ins/Array/S15.4_A1.1_T10.js",
     "built-ins/Array/S15.4.5.2_A1_T1.js",
     "built-ins/Array/S15.4.5.2_A3_T3.js",
+    "built-ins/String/prototype/repeat/repeat-string-n-times.js",
```
aapoalas, line 22: **question:** Why is this skipped? If it's simply because the engine's performance is not good enough to do the work within 10 seconds then I'd say leave it unskipped. It's then a valid reminder to us that this is too slow :)

aapoalas (dismissed):
LGTM!

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -561,8 +561,50 @@ impl StringPrototype {
...
+        }
+
+        if n == 1 {
+            return Ok(s.into_value());
+        }
+
+        // 6. Return the String value that is made from n copies of S appended together.
+        Ok(Value::from_string(
+            agent,
+            s.as_str(agent).repeat(n as usize),
```
aapoalas, line 606: **thought:** We might want to create a repeat-copy the index mapping as well... Hmm. For another day.

tests/skip.json
eliassjogreen: Yeah, it's because the while loop or comparison in the test takes too long. The operation itself is fast though.

aapoalas: I'll create a perf issue about that. Please remove the skip, let's let it annoy us so that we're forced to fix it :)

PR #356, opened 2024-07-28
feat(tests): Introduce `--update-metrics` flag

eliassjogreen:
Introduces a new `--update-metrics` flag which outputs the metrics of the test run. It also introduces a shorthand flag `--update` (`-u`) which sets both `--update-metrics` and `--update-expectations` true.

### DIFF
--- a/tests/metrics.json
+++ b/tests/metrics.json
@@ -1,0 +1,11 @@
+{
+  "results": {
+    "crash": 23248,
+    "fail": 6021,
+    "pass": 15752,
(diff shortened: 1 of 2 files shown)


aapoalas (approved):
LGTM!

andreubotella:
This seems to always be updating both the expectations and the metrics, even without passing any flags.

eliassjogreen:
Oh, I will create a fix for it today then

PR #357, opened 2024-07-28
feat(vm): Support `ArrayLike` argument in `Function.prototype.apply`

marc2332:
-> https://tc39.es/ecma262/#sec-function.prototype.apply

It seems like this was giving errors in the V8 benchmark suite.


aapoalas (approved):
LGTM! Thank you <3

PR #358, opened 2024-07-28
feat(cli): Allow replacing `print`

marc2332:


nova_cli/src/helper.rs
```diff
@@ -28,6 +28,7 @@ pub fn initialize_global_object(agent: &mut Agent, global: Object) {
             property_key,
             PropertyDescriptor {
                 value: Some(function.into_value()),
+                writable: Some(true),
```
aapoalas, line 31: **suggestion:** Also add `enumerable: Some(false)` and `configurable: Some(true)`: That makes it a fully defined descriptor (less chance of confusion in the future) and matches what eg. `parseInt` has.

aapoalas (approved):

PR #361, opened 2024-07-28
feat(ecmascript): Basic support for async functions

andreubotella:
This patch builds basic support for async functions on top of #316. It does not add support for `await`, but async functions that don't have `await` will always return a promise.

This patch also causes a number of regressions in test262 related to generators and async generators. This is because async, generator, and async generator functions could be created from function expressions, and they were being treated as regular functions. With this patch, calling sync or async generators will panic, since those are not yet implemented, which causes 12 test262 tests to crash when they were previously passing.

-----

This PR builds on top of #316, and shouldn't be merged before it. It might also be rebased on top of an alternative to that PR that adds heap vectors for each kind of ECMAScript function.


aapoalas (approved):
Looks damn good to me!

PR #362, opened 2024-07-29
refactor: Refactor promise reactions in preparation for `await`

andreubotella:
In the spec, there are only two types of promise reaction handlers: callback handlers and empty handlers. The internal behavior of `await` and of other spec-internal machinery that depends on promise reactions simply creates JS functions backed by a spec-internal abstract closure and uses them as callback handlers.

These JS functions, however, should not be observable from JS code, and in fact they should only ever be handled by the engine code that deals with promise reactions. Therefore, our implementation of `await` will instead use a new variant of the `PromiseReactionHandler` enum.

To support this, however, there should be a way to invoke PerformPromiseThen with the `on_fulfilled` and `on_rejected` arguments being `PromiseReactionHandler`. To this end, `perform_promise_then` is made closer to the spec, taking those arguments as `Value`s, and a new `inner_promise_then` function is added.

This patch also refactors `PromiseReactionRecord` to store the reaction type (whether it is a reaction to a promise fulfillment or rejection), rather than storing it in `PromiseReactionHandler` only in the empty case. This was not implemented this way initially, even though it matches the spec better, because in the spec only empty reaction handlers depend on the reaction type. However, this will no longer be true if we add `await` continuations as a reaction handler.

### DIFF
--- a/nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_prototype.rs
+++ b/nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_prototype.rs
@@ -86,13 +86,11 @@ impl PromisePrototype {
         let result_capability = PromiseCapability::new(agent);
 
         // 5. Return PerformPromiseThen(promise, onFulfilled, onRejected, resultCapability).
-        let on_fulfilled = Function::try_from(args.get(0)).ok();
-        let on_rejected = Function::try_from(args.get(1)).ok();
         perform_promise_then(
             agent,
             promise,
-            on_fulfilled,
-            on_rejected,
+            args.get(0),
+            args.get(1),
             Some(result_capability),
         );
         Ok(result_capability.promise().into_value())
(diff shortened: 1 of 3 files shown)


nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_prototype.rs
```diff
@@ -127,36 +125,58 @@ impl PromisePrototype {
...
-    on_rejected: Option<Function>,
+    on_fulfilled: Value,
+    on_rejected: Value,
     result_capability: Option<PromiseCapability>,
 ) {
     // 3. If IsCallable(onFulfilled) is false, then
     //     a. Let onFulfilledJobCallback be empty.
     // 4. Else,
     //     a. Let onFulfilledJobCallback be HostMakeJobCallback(onFulfilled).
+    // TODO: Add the HostMakeJobCallback host hook. Leaving it for later, since in implementations
```
aapoalas, line 136: **thought:** I guess this could eventually be a build flag, so only browsers pay this price.

aapoalas (approved):
LGTM!

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_prototype.rs
andreubotella: Are we planning of implementing wasm enabled by default, without a build flag? Because it seems like the wasm JS API depends on the incumbent realm, which is an HTML spec concept which needs Host{Make,Call}JobCallback to set up. V8 implements those host hooks according to the HTML spec by itself, and it doesn't give the embedder even the decision to not implement them.

andreubotella: Although I'm not sure if this is actually observable in any way in Node.js and Deno...

PR #363, opened 2024-07-29
refactor: Allow storing VM and execution state in GC'd types

andreubotella:
In order to allow `await` and `yield` to work, the execution state of a function at the time of evaluating an `await` or `yield` expression must be stored as part of one of the heap vectors, so it can later be resumed.

This means that `Vm`, `Executable` and `ExecutionContext` must implement `HeapMarkAndSweep`. This patch adds those implementations to those types, as well as many other internal types that did not already support GC tracing.

All such types must also implement `Send` and `Sync`, since the GC is meant to become multi-threaded in the future. This was not the case for `Executable`, since it includes static references into the oxc allocator, which is single-threaded. However, it would be fine to unsafely send those references across threads as long as they're only dereferenced in the main thread. So we add a `SendableRef` type to make this possible.

### DIFF
--- a/nova_vm/src/ecmascript/execution/execution_context.rs
+++ b/nova_vm/src/ecmascript/execution/execution_context.rs
@@ -3,9 +3,12 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use super::{Agent, EnvironmentIndex, PrivateEnvironmentIndex, RealmIdentifier};
-use crate::ecmascript::{
-    scripts_and_modules::{source_code::SourceCode, ScriptOrModule},
-    types::*,
+use crate::{
+    ecmascript::{
+        scripts_and_modules::{source_code::SourceCode, ScriptOrModule},
+        types::*,
+    },
+    heap::{CompactionLists, HeapMarkAndSweep, WorkQueues},
 };
 
 // TODO: Remove this.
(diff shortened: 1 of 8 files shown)


nova_vm/src/ecmascript/execution/execution_context.rs
```diff
@@ -93,6 +96,36 @@ impl ExecutionContext {
...
+        self.private_environment.sweep_values(compactions);
+    }
+}
+
+impl HeapMarkAndSweep for ExecutionContext {
+    fn mark_values(&self, queues: &mut WorkQueues) {
+        self.ecmascript_code.mark_values(queues);
+        self.function.mark_values(queues);
+        self.realm.mark_values(queues);
+        self.script_or_module.mark_values(queues);
```
aapoalas, line 118: **question:** Is this missing the source code marking? IIRC I needed to add that to this struct.

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -69,16 +69,43 @@ impl CompileContext<'_> {
     }
 }
 
+#[derive(Debug)]
+/// A `Send` and `Sync` wrapper over a `&'static T` where `T` might not itself
+/// be `Sync`. This is safe because the reference can only be obtained from the
+/// same thread in which the `SendableRef` was created.
+pub(crate) struct SendableRef<T: 'static> {
```
aapoalas, line 76: **thought:** I definitely would've just gone with "sure sure we don't deref this outside the main thread so it's all safe" :D but this is much better

aapoalas (dismissed):
LGTM! <3

nova_vm/src/ecmascript/execution/execution_context.rs
andreubotella: The implementation for `ECMAScriptCodeEvaluationState` is missing it indeed. I think this was an issue with rebasing.

aapoalas (approved):

PR #364, opened 2024-07-29
feat(ecmascript): Implement `await`

andreubotella:
The `await` operator pauses the execution of an async function and resumes it in a promise job when the awaited promise resolves. The spec describes this in terms of suspending an execution context and pushing or popping it from the stack, since in the spec an execution context contains all state needed to evaluate some code. In Nova, however, this state is also stored in `Executable` as well as in `Vm`. (See also #363, which made these types GC-able.)

Before this patch, there was no way to create a `Vm` struct, it was only created and dropped in `Vm::execute`. However, to allow suspending and resuming functions, the `Vm` must be preserved until the function is resumed. So we change the return value of `execute` from `JsResult<Option<Value>>` representing a return or throw completion, to an `ExecutionResult` enum which also allows representing `await` (and in the future, `yield`) suspensions.

The `ExecutionResult::Await` variant contains the `Vm` struct representing the current state of the VM at suspension time; as well as the value which was awaited, so the caller can perform `.then` on it. When the function is resumed, it will set be with either a normal completion containing a value (for fulfilled promises) or a throw completion (for rejected promises), so the `resume` and `resume_throw` methods are added to `Vm` to do so. They take `self` by value, and return `ExecutionResult`, so the previous suspended state of the VM cannot be used again.

The final piece of the `await` puzzle is the promise reactions that resume the execution. In the spec, JS functions are created to be the reaction handler callbacks for an `await`. However, these functions are never exposed to JS code. So instead we add `AwaitReaction`, a new heap type storing the state needed to resume an async function (including a `PromiseCapability` to resolve the returned promise after a return or throw), and we add a new `Await` variant to the `PromiseReactionHandler` enum to store it. (See also #362, which refactors `perform_promise_then` and promise reactions to allow for engine-internal reaction types.)


nova_vm/src/ecmascript/syntax_directed_operations/function_definitions.rs
```diff
@@ -198,8 +211,24 @@ pub(crate) fn evaluate_async_function_body(
...
+                    promise
+                } else {
+                    let new_capability = PromiseCapability::new(agent);
+                    new_capability.resolve(agent, awaited_value);
+                    new_capability.promise()
+                };
+                let handler = PromiseReactionHandler::Await(agent.heap.create(AwaitReaction {
+                    vm: Some(vm),
+                    executable: Some(exe),
+                    execution_context: Some(agent.running_execution_context().clone()),
```
aapoalas, line 227: **question:** Is it not possible to pop it here? Though I guess the computational complexity is actually exactly equal now that I think about it.

aapoalas (approved):
Awesome!

nova_vm/src/ecmascript/syntax_directed_operations/function_definitions.rs
andreubotella: Well, the context is popped in `ECMAScriptFunction::internal_call` after this function returns, so popping it here as well would mess up the context stack.

aapoalas: Yeah, I was wondering if we could somehow intimate to `internal_call` that we don't want it popping... But since this is actually trivially copyable data (IIRC) it shouldn't matter if we clone here and pop-to-nowhere there, or pop-to-heap here and do nothing there.

PR #366, opened 2024-07-29
feat(ecmascript): Unmapped arguments objects

aapoalas:
CreateUnmappedArgumentsObject now creates an actual `Object::Arguments(OrdinaryObject)` value.

I also cleaned up a few minor annoyances.


nova_vm/src/heap/heap_bits.rs
```diff
@@ -371,33 +311,96 @@ impl WorkQueues {
...
-            && self.dates.is_empty()
-            && self.bound_functions.is_empty()
-            && self.builtin_functions.is_empty()
-            && self.ecmascript_functions.is_empty()
-            && self.numbers.is_empty()
-            && self.objects.is_empty()
-            && self.regexps.is_empty()
-            && self.strings.is_empty()
-            && self.symbols.is_empty()
+        let Self {
```
aapoalas, line 314: **note:** This was an unreliable function as it was not being updated when more stuff was being added to WorkQueues.

Destructuring the borrow fixes that problem.

nova_vm/src/heap/heap_bits.rs
```diff
@@ -287,66 +287,6 @@ impl WorkQueues {
         }
     }
 
-    pub fn push_value(&mut self, value: Value) {
```
aapoalas, line 290: **note:** This function was mostly unused.

PR #370, opened 2024-07-30
feat(vm): basic chain expressions

littledivy:
Generate byte code for basic chain expressions like:
```js
print?.a // undefined
```

Couldn't get this to work tho, seems like a bug elsewhere (`operations_on_objects`?)
```js
const x = {}
x.a = 1;

x?.a; // Uncaught exception: TypeError: Not a callable object
```


nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -1313,12 +1313,30 @@ impl CompileEvaluation for ast::StaticMemberExpression<'_> {
             ctx.exe.add_instruction(Instruction::GetValue);
         }
 
+        ctx.exe.add_instruction(Instruction::LoadCopy);
```
aapoalas, line 1316: **issue:** Let's avoid the extra load-store work if we don't have optional. So move the instruction adding inside the if and return a Some(JumpIndex) or None if we don't have conditional access.

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -1313,12 +1313,30 @@ impl CompileEvaluation for ast::StaticMemberExpression<'_> {
...
+
+        ctx.exe.add_instruction(Instruction::Store);
         // 4. Return EvaluatePropertyAccessWithIdentifierKey(baseValue, IdentifierName, strict).
         let identifier = String::from_str(ctx.agent, self.property.name.as_str());
         ctx.exe.add_instruction_with_identifier(
             Instruction::EvaluatePropertyAccessWithIdentifierKey,
             identifier,
         );
+
+        let jump_to_end = ctx.exe.add_instruction_with_jump_slot(Instruction::Jump);
```
aapoalas, line 1333: **issue:** Here you can then conditionally produce the jump handling based on the Some(JumpIndex)/None from above. That way we don't need to add any jumps for when we do not have a conditional in the access.

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -1313,12 +1313,30 @@ impl CompileEvaluation for ast::StaticMemberExpression<'_> {
...
+        ctx.exe.set_jump_target_here(jump_to_access);
+
```
aapoalas, line 1335: **issue:** I think here you need to pop the value from the stack that LoadCopy put there.

aapoalas (changes requested):
Neat stuff! We can produce a little more efficient bytecode for non-optional cases, so left a few issues regarding that.

You'll need to update the expectations as well. First build release and then run the test262 binary with the --update flag. That should take care of that.

Thank you <3

nova_vm/src/engine/bytecode/executable.rs
littledivy: Done

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -1313,12 +1313,33 @@ impl CompileEvaluation for ast::StaticMemberExpression<'_> {
             ctx.exe.add_instruction(Instruction::GetValue);
         }
 
+        let maybe_jmp = if self.optional {
+            ctx.exe.add_instruction(Instruction::IsNullOrUndefined);
```
aapoalas, line 1317: **issue:** You shouldn't have removed the LoadCopy. That was necessary.

You can think of it like this: Before this `IsNullOrUndefined` we're at a state where stack has a depth of N which we can decide for our purposes to be 0. The result register has our `foo?.bar` expressions `foo` in it. `IsNullOrUndefined` takes the result register value and replaces it with a true if it was null or undefined, otherwise it puts in false. So, we lose `foo` entirely if we do nothing.

So, the `LoadCopy` you had here originally was perfectly correct. With `LoadCopy` and then `IsNullOrUndefined` done we're at a state where stack has depth 1 (we decided it was 0 before), containing `foo` and result register contains true or false. Now if the result is nullish, we should jump to returning undefined. Otherwise we need to move pop `foo` from the stack back into the result register.

So, I think the jump here should actually be `JumpIfTrue` (IIRC I added that a few weeks ago), and after the jump (so if we did not jump from the nullish check) you should add back in the `Store` instruction you had there originally. (Note: Both the LoadCopy and Store should still be done inside the if of course.)

Then, if we from `maybe_jmp` then we arrive at its "receiving end" with a stack depth of 1 (containing `foo`) and an empty result. We always want to leave the function with a stack depth of 0 (otherwise we're leaking stack memory) and in the was-nullish case with `Value::Undefined` in the result register.

So at the receiving end, after `jump_to_end` (which is there to jump over the was-nullish case handling) you need to add `Store` to pop `foo` from the stack before you do the `StoreConstant`. That way we empty the stack before exiting.

aapoalas:
@littledivy Sorry I ended up kinda trying this for myself. This is mostly working but I couldn't solve an issue I brought upon myself. This sort of code fails:
```js
(async () => {
    print(undefined?.[
        await Promise.reject(new Error('unreachable'))
    ]);
})()
```

The reason basically is that I added ChainExpression as a possible valid "reference" in the `is_reference` call. This ... is sort of right but sort of wrong. It's right in that there are places where ChainExpression can return a reference and then we'll want to `GetValueKeepReference`. Or I think so, anyway.

But it's wrong in the sense that it's not actually okay to do `GetValue` unconditionally after ChainExpression. The state of the stack and everything after a ChainExpression as written right now can be either "nullish on value stack" or "reference on reference stack". This is obviously no good as we cannot statically know where we end up in.

I added this `optional_chains` Vec into the compiling context. I think that should be moved a step up and it would then act as our statically knowable differentiator of where we end up in. If we jump from `optional_chains` then we're in "nullish on value stack" otherwise we're in "reference on reference stack".

Another option would be to have some instruction that turns "nullish on value stack" into "undefined value reference on reference stack" but that's kinda dumb.

aapoalas (approved):

aapoalas:
Thank you @littledivy for starting this work, and sorry for stealing the PR a bit :)

I _think_ the chain expressions are now fully supported and should be pretty much correct. It required some mangling and this-that but nothing too-too bad.

PR #377, opened 2024-08-01
feat(ecamscript): Basic generator implementation

andreubotella:
This patch implements generator objects, but it leaves the `yield` expression for a follow-up.


nova_vm/src/ecmascript/syntax_directed_operations/function_definitions.rs
```diff
@@ -251,3 +269,50 @@ pub(crate) fn evaluate_async_function_body(
...
+
+    // 4. Perform GeneratorStart(G, FunctionBody).
+    // SAFETY: We're alive so SourceCode must be too.
+    let body = unsafe {
+        agent[function_object]
+            .ecmascript_function
+            .ecmascript_code
+            .as_ref()
+    };
+    let is_concise_arrow_function = agent[function_object]
```
aapoalas, line 307: **nitpick:** IIRC it's not possible for a concise arrow function to be a generator?

aapoalas (approved):
LGTM!

nova_vm/src/ecmascript/syntax_directed_operations/function_definitions.rs
andreubotella: True.

PR #379, opened 2024-08-02
feat(ecmascript): Implement `JSON.parse`

eliassjogreen:
This PR implements `JSON.parse` using https://github.com/cloudwego/sonic-rs but it should be pretty simple to replace it if we ever have the need to. At first I followed the spec but soon realized it was just as simple to just convert it directly from the json value type of sonic-rs so thats what I did, seems to pass the same number of tests either way.

The only issue I found with sonic rs was that it doesn't parse `-0` as a negative float and therefor we fail one of the tests. That will need to be fixed in sonic-rs so I will create a ticket for that tomorrow.

Also serde_json/sonic-rs messes with the impl for `From<i64>` for some reason which is why I had to replace a bunch of `.into()` with `.into_i64()`.

## Performance

This is fun, we are faster than node, deno and bun when it comes to JSON parsing! At least for my super dumb benchmark:

```
for (let i = 0; i < 1e3; i++) {
  JSON.parse('{"a":1,"b":2,"c":3,"d":4,"e":5}');
}
```

```
❯ hyperfine --warmup 10 "./target/release/nova_cli eval scripts/json.js" "node scripts/json.js" "deno run scri
pts/json.js" "bun run scripts/json.js"
Benchmark 1: ./target/release/nova_cli eval scripts/json.js
  Time (mean ± σ):      13.7 ms ±   2.0 ms    [User: 7.8 ms, System: 3.4 ms]
  Range (min … max):    10.2 ms …  19.5 ms    156 runs

Benchmark 2: node scripts/json.js
  Time (mean ± σ):      74.5 ms ±   4.3 ms    [User: 56.1 ms, System: 15.2 ms]
  Range (min … max):    69.2 ms …  87.2 ms    37 runs

Benchmark 3: deno run scripts/json.js
  Time (mean ± σ):      61.9 ms ±   3.5 ms    [User: 40.9 ms, System: 22.9 ms]
  Range (min … max):    56.3 ms …  70.5 ms    47 runs

Benchmark 4: bun run scripts/json.js
  Time (mean ± σ):      34.1 ms ±   2.6 ms    [User: 18.5 ms, System: 15.7 ms]
  Range (min … max):    29.3 ms …  39.7 ms    63 runs

Summary
  ./target/release/nova_cli eval scripts/json.js ran
    2.49 ± 0.41 times faster than bun run scripts/json.js
    4.51 ± 0.71 times faster than deno run scripts/json.js
    5.43 ± 0.85 times faster than node scripts/json.js
```

This benchmark is definitely flawed, if for no other reason than the increased startup times of the runtimes but it's fun none the less.

## Performance the sequel

Running https://github.com/GoogleChromeLabs/json-parse-benchmark I realized that what I said above is actually false. Performance sucks.

At least when parsing objects where there are lots of strings. This is not because of the JSON parsing or object creation but because of the heap method `find_equal_string`. Commenting it out, even though I know that is incorrect, makes the `JSON.parse` function really fast again. I am guessing the `find_equal_string` method could be heavily optimized using a cache of some sort and some smart algo, or perhaps just a simple hash table?


tests/expectations.json
```diff
@@ -3031,6 +2966,8 @@
   "built-ins/JSON/stringify/replacer-array-wrong-type.js": "CRASH",
   "built-ins/JSON/stringify/replacer-function-abrupt.js": "FAIL",
   "built-ins/JSON/stringify/replacer-function-arguments.js": "FAIL",
+  "built-ins/JSON/stringify/replacer-function-array-circular.js": "CRASH",
+  "built-ins/JSON/stringify/replacer-function-object-circular.js": "CRASH",
```
andreubotella, line 2970: Have you looked into why these crashes are happening?

andreubotella (approved):
LGTM

tests/expectations.json
eliassjogreen: It's because I added the `stringify` builtin string and corrected the function name I think so it's to be expected.

andreubotella: So we're talking tests that expect a `TypeError`, and were seeing one because the `stringify` function didn't exist, but now that it exists and they're calling it, they crash instead?

load1n9 (approved):
LGTM

nova_vm/src/ecmascript/builtins/structured_data/json_object.rs
```diff
@@ -68,3 +167,119 @@ impl JSONObject {
...
+                let prop = PropertyKey::from(SmallInteger::try_from(i as i64).unwrap());
+                let js_value = value_from_json(agent, value)?;
+                create_data_property(agent, array_obj, prop, js_value)?;
+            }
+            Ok(array_obj.into())
+        }
+        sonic_rs::JsonType::Object => {
+            let json_object = json.as_object().unwrap();
+            let object =
+                ordinary_object_create_with_intrinsics(agent, Some(ProtoIntrinsics::Object), None);
```
aapoalas, line 276: **thought:** We need some way to reserve the backing store size before we go into the loop. That way this won't eg. first fill a backing store of 8 property values, then copy them into a larger backing store and continue filling that.

aapoalas (approved):
Niiice! <3

PR #381, opened 2024-08-03
feat: `Number.prototype.toPrecision`

marc2332:
As I said in the discord a few days ago, I initially tried to implement this from scratch by myself, but I couldn't get it to work fully. So instead, I copied the implementation of Boa and adjusted it to fit with the Nova APIs.


nova_vm/src/ecmascript/builtins/numbers_and_dates/number_objects/number_prototype.rs
```diff
@@ -160,8 +160,213 @@ impl NumberPrototype {
...
+                ExceptionType::RangeError,
+                "Precision out of range",
+            ));
+        }
+
+        // 6.
+        let mut x_f64 = x.into_f64(agent);
+
+        // 7.
+        let mut s = std::string::String::new();
```
aapoalas, line 202: **suggestion:** This probably deserves some `with_capacity` call.

aapoalas (approved):

PR #382, opened 2024-08-06
feat(array): Implement `from_slice` to array

load1n9:


nova_vm/src/ecmascript/builtins/array.rs
```diff
@@ -44,6 +44,21 @@ impl Array {
         Self(ArrayIndex::from_u32_index(0))
     }
 
+    /// Creates a new array with the given elements that are converted to values.
+    pub fn from_slice(agent: &mut Agent, elements: &[impl IntoValue]) -> Self {
+        let array = agent.heap.create(ArrayHeapData::default());
```
andreubotella, line 49: It'd be better to reserve the left upfront:

```suggestion
        let mut heap_data = ArrayHeapData::default();
        heap_data.elements.reserve(
            &mut agent.heap.elements,
            u32::try_from(elements.len()).unwrap(),
        );
        let array = agent.heap.create(heap_data);
```

andreubotella (approved):

andreubotella (approved):

nova_vm/src/ecmascript/builtins/array.rs
```diff
@@ -44,6 +44,26 @@ impl Array {
...
 
+    /// Creates a new array with the given elements that are converted to values.
+    pub fn from_slice(agent: &mut Agent, elements: &[impl IntoValue]) -> Self {
+        let mut heap_data = ArrayHeapData::default();
+        heap_data.elements.reserve(
+            &mut agent.heap.elements,
+            u32::try_from(elements.len()).unwrap(),
+        );
+        let array = agent.heap.create(heap_data);
+        for (index, value) in elements.iter().enumerate() {
```
aapoalas, line 55: **issue:** We don't need to use `[[DefineOwnProperty]]` as we know exactly which steps it'll take in this case.

We can skip right to the chase and just do something like this:
```rs
heap_data.elements.as_mut_slice().iter_mut().zip(elements).for_each(|(elements_slot, element) {
  *elements_slot = Some(element.into_value());
});
```
I believe this should compile down to just a memcpy.

aapoalas (changes requested):
Let's go a tad bit lower :)

aapoalas (approved):
@andreubotella is absolutely right <3 I switched this to call `create_array_from_list`.

load1n9:
LGTM

PR #385, opened 2024-08-10
feat(ecmascript): Array iterator objects

andreubotella:


nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_iterator_objects/array_iterator_prototype.rs
```diff
@@ -0,0 +1,142 @@
...
+            }
+            // 3. If kind is value, then
+            ArrayIteratorKind::Value => {
+                // 1. Let elementKey be ! ToString(indexNumber).
+                // 2. Let elementValue be ? Get(array, elementKey).
+                // a. Let result be elementValue.
+                match array {
+                    Object::Array(array) if agent[array].object_index.is_none() => {
+                        let idx = usize::try_from(index.into_i64()).unwrap();
+                        array.as_slice(agent)[idx].unwrap_or(Value::Undefined)
```
aapoalas, line 97: **issue:** If the slot contains None or is out of bounds, then we need to go into the prototype chain or call possibly call a getter.

Checking the slot being Some can also be done even if we have a backing object.

The internal_get method of Array could check that for indexed properties directly actually.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -573,8 +575,18 @@ impl ArrayPrototype {
         Ok(o.into_value())
     }
 
-    fn entries(_agent: &mut Agent, _this_value: Value, _: ArgumentsList) -> JsResult<Value> {
-        todo!()
+    fn entries(agent: &mut Agent, this_value: Value, _: ArgumentsList) -> JsResult<Value> {
+        // 1. Let O be ? ToObject(this value).
+        let Value::Array(o) = this_value else {
+            return Err(
+                agent.throw_exception_with_static_message(ExceptionType::TypeError, "blargh")
```
aapoalas, line 582: **praise:** blargh indeed

nova_vm/src/heap.rs
```diff
@@ -153,6 +155,7 @@ impl Heap {
         let mut heap = Heap {
             array_buffers: Vec::with_capacity(1024),
             arrays: Vec::with_capacity(1024),
+            array_iterators: Vec::with_capacity(1024),
```
aapoalas, line 158: **nitpick:** Probably unnecessarily large default size.

aapoalas (changes requested):
Wrong logic for array indexed property access. Otherwise looks really good, thank you! <3

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
andreubotella: I also didn't realize that I changed the implementation of `next` to allow non-array this values but didn't change the array prototype methods to allow arbitrary objects.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_iterator_objects/array_iterator.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// This is only for Value discriminant creation.
+    pub(crate) const fn _def() -> Self {
+        Self(ArrayIteratorIndex::from_u32_index(0))
+    }
+
+    pub(crate) fn get_index(self) -> usize {
+        self.0.into_index()
+    }
+
+    pub(crate) fn from_array(agent: &mut Agent, array: Object, kind: ArrayIteratorKind) -> Self {
```
aapoalas, line 35: **nitpick:** Should probably be "from_object".

aapoalas (approved):

PR #388, opened 2024-08-13
feat(vm): Support the rest pattern in array destructuring

andreubotella:


aapoalas (approved):
**praise:** That turned out way simpler than I anticipated! Great job <3

PR #389, opened 2024-08-13
feat(ecmascript): Add initial version of `Object.groupBy`

### LINKED ISSUE #252: Implement `Object.groupBy`
[Object.groupBy](https://tc39.es/ecma262/#sec-object.groupby) is the only builtin function still missing from Object constructor. The main work needed here is the [GroupBy](https://tc39.es/ecma262/#sec-groupby) abstract operation.

thegreatercurve:
Closes #252


thegreatercurve:
@aapoalas Is there a way of running individual test262 tests to see the runtime errors from the tests which are still crashing?

eliassjogreen:
> @aapoalas Is there a way of running individual test262 tests to see the runtime errors from the tests which are still crashing?

Running `cargo run --bin test262 -- eval-test path/to/the/test.js` should do the trick. Remember to rebuild the nova cli before running too as the test runner does not do that.

aapoalas:
> @aapoalas Is there a way of running individual test262 tests to see the runtime errors from the tests which are still crashing?

What I usually do is that I run the `eval-test` command once, copy its output and replace `.../target/debug/nova_cli` with `cargo run` to get automatic cargo builds of `nova_cli` in subsequent runs.

nova_vm/src/ecmascript/abstract_operations/operations_on_objects.rs
```diff
@@ -756,3 +763,143 @@ pub(crate) fn get_function_realm(
...
+
+    // 3. Append group to groups.
+    groups.push(group);
+
+    // 4. Return UNUSED.
+    Ok(())
+}
+
+pub(crate) struct GroupByRecord {
+    pub(crate) key: Value,
```
aapoalas, line 801: **issue:** This should rather be `PropertyKey`.

aapoalas (changes requested):
Some issues made their way in, but overall looks good to me! Thank you very much <3

thegreatercurve:
@aapoalas Made the requested changes. The remaining string test which is not passing appears to be an unrelated issue with parsing emoji Unicode chars.

aapoalas:
> @aapoalas Made the requested changes. I was using the wrong method to pass in the value to the callback.
> 
> The remaining string test which is not passing appears to be an unrelated issue with parsing emoji Unicode chars.

Looks good to me! I had made a mistake when I said that the type in the Group should be PropertyKey: It should only be so in the Property case. The GroupBy function is essentially generic with two variants. I manually split it apart while making the internal helper functions actually generic.

Now we get nice types :)

thegreatercurve:
@aapoalas Thanks for the additional context and adding the extra variant.

aapoalas (approved):

aapoalas:
Again, thank you very much and good job! <3

PR #394, opened 2024-08-21
feat(vm): Support default values in array destructuring

andreubotella:
The way array and object destructuring are currently implemented by jumping into a separate function that consumes destructuring-related bytecode instructions, and doesn't return to the regular VM execution until the destructuring is done.

This works fine for simple cases, but when the destructuring includes initializers for default values, this no longer works. This is because the initializers could contain arbitrary expressions, and so their bytecode instructions would need to be run in the regular VM execution.

As it turns out, array binding patterns were already separated into simple and complex patterns, where the presence of initializers would make the pattern complex. Complex patterns had not been implemented yet, but they would have also jumped into a separate function in the VM execution. This patch instead implements complex array binding patterns with regular bytecode instructions dealing with iterators.


nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -1264,6 +1251,34 @@ impl Vm {
...
+            }
+            Instruction::IteratorRestIntoArray => {
+                let mut iterator = vm.iterator_stack.pop().unwrap();
+                let capacity = iterator.remaining_length_estimate(agent).unwrap_or(0);
+                let array = array_create(agent, 0, capacity, None)?;
+
+                let mut idx: u32 = 0;
+                while let Some(value) = iterator.step_value(agent)? {
+                    let key = PropertyKey::Integer(idx.into());
+                    create_data_property_or_throw(agent, array, key, value)?;
```
aapoalas, line 1277: **thought:** I'm pretty sure this cannot throw under any circumstances, right? We would maybe benefit from an `Array::push` method that could be used directly in these cases.

aapoalas (dismissed):
LGTM, thank you <3 I had some minor comments

nova_vm/src/engine/bytecode/vm.rs
andreubotella: You're right that this can't throw, I somehow ended up copying this line from the `ArrayPush` instruction and failed to notice that the spec explicitly marks this operation with `!`.

aapoalas (approved):

Issue #183, opened 2024-05-14
PropertyKey does not accurately check conversions from Strings


aapoalas:
We rely on an assumption that number-like `PropertyKey`s show up as `Integer` and that property keys can be compared by-value so that eg. `"0" == 0` but the `From` and `TryFrom` implementations of `PropertyKey` do not actually check if strings are number-like.

As a result, `ProperyKey::from("0") != PropertyKey::from(0)` which is of course not correct.

This sort of points to the direction that the `From<Value>` and `From<String>` type of conversions for `PropertyKey` are invalid: We cannot freely and cheaply convert a string into a `PropertyKey` without first checking its contents. It would thus be better to remove these `String` and `Value` conversions and instead implement functions that do the same but also take an `&Agent` parameter so as to be capable of checking the string data.

nair-sreerag:
Hi @aapoalas, is this bug fixed? I tried it in js and it worked as expected. If its not, how can i reproduce it?

Below is the code I ran it on - 
``` test.js
const obj = {};

obj["0"] = "string zero";
console.log("obj[0]:", obj[0]); // Should print "string zero"
console.log("obj['0']:", obj["0"]); // Should print "string zero"
console.log("obj[0] === obj['0']:", obj[0] === obj["0"]);  // should return true
```

aapoalas:
This is _mostly_ fixed; I think it will might still appear with numbers larger than 7 characters in length, like `obj[12345678]`.

nair-sreerag:
Will try to reproduce it then. 
Is the approach in the sample code correct though?

aapoalas:
Yeah, that seems correct. There is a further complication that the PropertyKey <-> Integer <-> String <-> Value paths are somewhat split here-and-there in the engine code, a lot of code will eg. just use `PropertyKey::from(some_string)` which can perform the correct conversion for up to 7 character integer strings but not for any others. So, it's possible that your test code will actually function correctly but in some more obscure case it will fail, eg. if the `"12345678"` string is first assigned to a variable:

```js
const obj = {};
const str = "12345678"
const int = 12345678;

obj[str] = "string zero";
console.log("obj[int]:", obj[int]); // Should print "string zero"
console.log("obj[str]:", obj[str]); // Should print "string zero"
console.log("obj[int] === obj[str]:", obj[int] === obj[str]);  // should return true
```

nair-sreerag:
This seems to be working too. I am getting true in the last console.log()
[...]

PR #398, opened 2024-08-24
refactor(vm): Refactor simple object patterns

andreubotella:
This patch prepares object patterns to be split into simple and complex similarly to array patterns (see #394), where complex patterns (having either initializers or computed property keys) would be transpiled into regular bytecode.

Since only a subset of simple object patterns were supported so far, this patch refines the compilation to only use two binding properties, `BindingPatternBind` and `BindingPatternBindNamed`. In the future, rest bindings will also be supported.

Despite being a refactoring, this patch makes certain tests pass due to bugs in the original implementation of object patterns.


nova_vm/src/ecmascript/types/language/object/property_key.rs
```diff
@@ -52,6 +52,15 @@ impl PropertyKey {
         self.into()
     }
 
+    pub fn from_value(agent: &Agent, value: Value) -> Option<Self> {
```
aapoalas, line 55: **praise:** This is good. I think I have an issue somewhere (#183) that basically says we should remove the TryFrom and Into impls to convert between PropertyKey and Value/String so freely, since they're effectively unsafe from consistency point of view.

An unsafe method to turn a String into a PropertyKey without checking for integerness is probably worth a consideration, though.

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -2044,278 +2046,68 @@ impl CompileEvaluation for ast::ObjectPattern<'_> {
...
+                );
+            } else {
+                let key_string = match &ele.key {
+                    ast::PropertyKey::StaticIdentifier(identifier) => {
+                        String::from_str(ctx.agent, &identifier.name).into_value()
+                    }
+                    ast::PropertyKey::NumericLiteral(literal) => {
+                        let numeric_value = Number::from_f64(ctx.agent, literal.value);
+                        if let Number::Integer(_) = numeric_value {
+                            numeric_value.into()
```
aapoalas, line 2072: **nitpick:** `into_value()` to be explicit about the conversion target and fit with the rest of the code.

nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -1390,50 +1390,40 @@ impl Vm {
...
-                    this_value: None,
-                };
+            match instr.kind {
+                Instruction::BindingPatternBind | Instruction::BindingPatternBindNamed => {
+                    let binding_id =
+                        vm.fetch_identifier(executable, instr.args[0].unwrap() as usize);
+                    let property_key = if instr.kind == Instruction::BindingPatternBind {
+                        binding_id.into()
+                    } else {
+                        PropertyKey::from_value(
```
aapoalas, line 1409: **thought:** It's a bit sad to check the integerness of the value every time, instead of doing the conversion once at compile time and storing a String or Number based on the result, and then doing essentially `PropertyKey::from_value_unchecked` here.

aapoalas (approved):
Looks very good to me! Thank you, great to see my idiocy get fixed :) <3

aapoalas (approved):

PR #400, opened 2024-08-24
feat(vm): Implement complex object binding patterns

andreubotella:
This patch implements support for default values, nested destructuring and computed rather than static property keys in object destructuring bindings. It does so by building a slow path, separate from the previous fast path of simple object binding, which implements the entire destructuring in regular bytecode. This is parallel to the distinction between simple and complex array binding introduced in #394.

----

This PR builds on top of #398 and #399, do not merge before those.


nova_vm/src/engine/small_integer.rs
```diff
@@ -3,7 +3,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 /// 56-bit signed integer.
-#[derive(Clone, Copy, PartialEq)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
```
aapoalas, line 6: **nitpick:** A manual Hash impl that takes out the i64 and hashes that would probably be a good idea.

nova_vm/src/ecmascript/abstract_operations/operations_on_objects.rs
```diff
@@ -763,6 +766,55 @@ pub(crate) fn get_function_realm(
...
+            if dest.enumerable.unwrap() {
+                // 1. Let propValue be ? Get(from, nextKey).
+                let prop_value = get(agent, from, next_key)?;
+                // 2. Perform ! CreateDataPropertyOrThrow(target, nextKey, propValue).
+                entries.push(ObjectEntry::new_data_entry(next_key, prop_value));
+            }
+        }
+    }
+
+    Ok(agent.heap.create_object_with_prototype(
```
aapoalas, line 808: **praise:** Niiiice! <3

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -2040,74 +2040,224 @@ fn complex_array_pattern(
...
-                match &ele.value.kind {
-                    ast::BindingPatternKind::BindingIdentifier(identifier) => {
-                        let value_identifier_string = ctx.create_identifier(&identifier.name);
-                        ctx.exe.add_instruction_with_identifier_and_constant(
-                            Instruction::BindingPatternBindNamed,
-                            value_identifier_string,
-                            key_string,
-                        )
+            }
+            _ => todo!("Complex object patterns"),
```
aapoalas, line 2130: **nitpick:** unreachable.

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -2040,74 +2040,224 @@ fn complex_array_pattern(
...
+    // NOTE: RequireObjectCoercible throws in the same cases as ToObject, and other operations
+    // later on (such as GetV) also perform ToObject, so we convert to an object early.
+    ctx.exe.add_instruction(Instruction::Store);
+    ctx.exe.add_instruction(Instruction::ToObject);
+    ctx.exe.add_instruction(Instruction::Load);
+
+    for property in &object_pattern.properties {
+        match &property.key {
+            ast::PropertyKey::StaticIdentifier(identifier) => {
+                ctx.exe.add_instruction(Instruction::Store);
```
aapoalas, line 2154: **thought:** Do we need a `StoreCopy` as well? :)

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -2040,74 +2040,224 @@ fn complex_array_pattern(
...
+            ast::BindingPatternKind::AssignmentPattern(_) => unreachable!(),
+        }
+    }
+
+    if let Some(rest) = &object_pattern.rest {
+        let ast::BindingPatternKind::BindingIdentifier(identifier) = &rest.argument.kind else {
+            unreachable!()
+        };
+
+        // We have kept the references for all of the properties read in the reference stack, so we
```
aapoalas, line 2242: **praise:** This is brilliant. Nice work!

nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -581,6 +584,26 @@ impl Vm {
                 );
                 vm.stack.push(object.into())
             }
+            Instruction::CopyDataPropertiesIntoObject => {
+                // 7.3.25 CopyDataProperties ( target, source, excludedItems )
+                // 1. If source is either undefined or null, return unused.
+                // 2. Let from be ! ToObject(source).
+                let Ok(from) = Object::try_from(vm.result.unwrap()) else {
+                    return Ok(ContinuationKind::Normal);
```
aapoalas, line 592: **question:** Shouldn't we still clear the reference stack (though I suppose we don't _actually_ ever enter here with a non-object on the stack)?

aapoalas (approved):
Absolutely beautiful work! <3

nova_vm/src/engine/small_integer.rs
andreubotella: Done.

nova_vm/src/engine/bytecode/executable.rs
andreubotella: Fixed.

nova_vm/src/engine/bytecode/executable.rs
andreubotella: Maybe at some point.

nova_vm/src/engine/bytecode/vm.rs
andreubotella: I changed this to instead panic if the result value isn't an object.

aapoalas (approved):

PR #402, opened 2024-08-25
feat(vm): Reimplement FunctionDeclarationInstantiation as part of function compilation

andreubotella:
FunctionDeclarationInstantiation is an abstract operation in the spec which runs at the start of a function call and sets up the parameters, lexical environments and variables for the function. Until now, we implemented it as a Rust function called before the function body was executed.

Since this operation is not implemented using bytecode, and it binds the function arguments, it couldn't support complex bindings (such as default values or object/array bindings) without reimplementing them.

This patch therefore instead implements this operation as bytecode instructions emitted before the function body. That way the binding of arguments can be compiled into a regular array binding pattern.

Doing this requires some way to pass the arguments into the VM, and have them stored in its state until they are evaluated. We do this by having `Vm::execute` take an optional arguments slice, which it stores on the iterator stack.

For generators, currently FunctionDeclarationInstantiation happens when the generator function is called, and this patch moves it to when the generator object is first resumed, which is correct. However, this means that the arguments must be stored until that first time the generator is resumed. To do this, the `Option<Vm>` field in `GeneratorState::Suspended` is made into an enum which can store either a `Vm` or a boxed slice of arguments.

------

This PR builds on top of #398, #399, #400 and #401, do not merge before those.


nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -63,6 +65,26 @@ pub(crate) struct CompileContext<'agent> {
 }
 
 impl CompileContext<'_> {
+    fn new(agent: &'_ mut Agent) -> CompileContext<'_> {
```
aapoalas, line 68: **praise:** Good idea.

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -80,28 +102,31 @@ impl CompileContext<'_> {
...
-impl<T> SendableRef<T> {
+impl<T: ?Sized> SendableRef<T> {
     pub(crate) fn new(reference: &'static T) -> Self {
         Self {
             reference,
             thread_id: std::thread::current().id(),
         }
     }
-    pub(crate) fn get(&self) -> &T {
+    pub(crate) unsafe fn new_as_static(reference: &T) -> Self {
```
aapoalas, line 117: **suggestion:** A comment about the safety conditions for calling this function would not go amiss.

nova_vm/src/engine/bytecode/executable/function_declaration_instantiation.rs
```diff
@@ -0,0 +1,334 @@
+use ahash::{AHashMap, AHashSet};
```
aapoalas, line 1: **nitpick:** Missing MPL header.

nova_vm/src/engine/bytecode/executable/function_declaration_instantiation.rs
```diff
@@ -0,0 +1,334 @@
...
+pub(crate) fn instantiation(
+    ctx: &mut CompileContext,
+    formals: &FormalParameters,
+    body: &FunctionBody<'_>,
+    strict: bool,
+    is_lexical: bool,
+) {
+    // 5. Let parameterNames be the BoundNames of formals.
+    // 6. If parameterNames has any duplicate entries, let hasDuplicates be true. Otherwise, let hasDuplicates be false.
+    let mut parameter_names = AHashSet::with_capacity(formals.parameters_count());
```
aapoalas, line 31: **praise:** Nice.

nova_vm/src/engine/bytecode/executable/function_declaration_instantiation.rs
```diff
@@ -0,0 +1,334 @@
...
+    //   e. Set the LexicalEnvironment of calleeContext to env.
+    if !strict && has_parameter_expressions {
+        ctx.exe
+            .add_instruction(Instruction::EnterDeclarativeEnvironment);
+    }
+
+    // 21. For each String paramName of parameterNames, do
+    // NOTE: The behavior should not depend on the order in which the parameter
+    // names are iterated, so it's fine for `parameter_names` to be a set.
+    for param_name in &parameter_names {
```
aapoalas, line 98: **thought:** At some point it might be a good idea to create a `InitializeDeclarativeEnvironment` kind of instruction that would point to a "declarative environment template" which has (mostly?) uninitialized bindings for all the names that will be needed. Performing that instruction would check that the current declarative environment is empty, and then replace its Bindings Map with a clone of a Bindings Map that comes from the Vm context (or that place where we save bindings etc. anyhow).

Essentially, a quicker way to prep the declarative environment since we know beforehand which names we're going to define in it.

nova_vm/src/engine/bytecode/executable/function_declaration_instantiation.rs
```diff
@@ -0,0 +1,334 @@
...
+        parameter_names.insert("arguments".into());
+    }
+
+    // 24. Let iteratorRecord be CreateListIteratorRecord(argumentsList).
+    // 25. If hasDuplicates is true, then
+    //   a. Perform ? IteratorBindingInitialization of formals with arguments iteratorRecord and undefined.
+    // 26. Else,
+    //   a. Perform ? IteratorBindingInitialization of formals with arguments iteratorRecord and env.
+    if !formals.has_parameter() {
+        ctx.exe.add_instruction(Instruction::IteratorClose)
```
aapoalas, line 168: **suggestion:** A comment about how this is here to remove the arguments iterator from the iterator stack would not be out of place perhaps.

nova_vm/src/engine/bytecode/executable/function_declaration_instantiation.rs
```diff
@@ -0,0 +1,334 @@
...
+    } else if has_parameter_expressions {
+        complex_array_pattern(
+            ctx,
+            formals.items.iter().map(|param| Some(&param.pattern)),
+            formals.rest.as_deref(),
+            !has_duplicates,
+        );
+    } else {
+        simple_array_pattern(
```
aapoalas, line 177: **praise:** This stuff is really cool seeing.

nova_vm/src/engine/bytecode/executable/function_declaration_instantiation.rs
```diff
@@ -0,0 +1,334 @@
...
+        for n in function_body_var_declared_names(body) {
+            // i. If instantiatedVarNames does not contain n, then
+            if instantiated_var_names.contains(&n) || parameter_names.contains(&n) {
+                continue;
+            }
+            // 1. Append n to instantiatedVarNames.
+            let n_string = String::from_str(&mut ctx.agent, &n);
+            instantiated_var_names.insert(n);
+            // 2. Perform ! env.CreateMutableBinding(n, false).
+            ctx.exe
```
aapoalas, line 201: **question:** Isn't this the `CreateAndInitializeVariableBinding` instruction you added?

nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -1003,6 +1016,56 @@ impl Vm {
                 let w = vm.result.take().unwrap();
                 initialize_referenced_binding(agent, v, w)?;
             }
+            Instruction::InitializeVariableEnvironment => {
+                let num_variables = instr.args[0].unwrap();
+                assert!(instr.args[1].unwrap() <= 1);
```
aapoalas, line 1021: **thought:** I was surprised this couldn't be done with `bool::try_from(u16_value).unwrap()` but apparently that's not a thing.

tests/expectations.json
aapoalas, line 1: **praise:** That's a nice amount of new tests passing! <3

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -461,6 +461,51 @@ fn is_chain_expression(expression: &ast::Expression) -> bool {
     }
 }
 
+trait ContainsExpression {
```
aapoalas, line 481: **praise:** This is a nice impl to have <3

**suggestion:** This should maybe go in the syntax_directed_operations folder (or was it a module?), same place where `VarDeclaredNames` and friends live in.

nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -1341,55 +1306,73 @@ impl Vm {
...
-            if instr.kind == Instruction::BindingPatternBindRest {
-                let capacity = iterator.remaining_length_estimate(agent).unwrap_or(0);
-                let rest = array_create(agent, 0, capacity, None).unwrap();
-                let mut idx = 0u32;
-                while let Some(result) = iterator.step_value(agent)? {
-                    create_data_property_or_throw(agent, rest, PropertyKey::from(idx), result)
-                        .unwrap();
-                    idx += 1;
-                }
+            let value = match instr.kind {
```
aapoalas, line 1414: **praise:** This stuff is a work of art. Beautiful <3

aapoalas (approved):
One word: Awesome!

nova_vm/src/engine/bytecode/executable.rs
andreubotella: This was added in #401, not in the changes specific to this PR. I'll make a follow-up PR to move it.

nova_vm/src/engine/bytecode/executable/function_declaration_instantiation.rs
andreubotella: If `!has_parameter_expressions`, then the variables are initialized in `env`, which is the *lexical* environment; otherwise, they are initialized in `varEnv`, which is the *variable* environment (and might also be set to the lexical environment or not afterwards depending on `strict`).

I added that instruction in order to do the `varEnv` initialization if `has_parameter_expressions`, but then I decided to add `InitializeVariableEnvironment` instead, and I forgot to remove `CreateAndInitializeVariableBinding`.

aapoalas (approved):

aapoalas (approved):

nova_vm/src/engine/bytecode/executable.rs
andreubotella: Done.

nova_vm/src/engine/bytecode/executable/function_declaration_instantiation.rs
andreubotella: Done.

nova_vm/src/engine/bytecode/executable/function_declaration_instantiation.rs
andreubotella: Done.

nova_vm/src/engine/bytecode/executable/function_declaration_instantiation.rs
andreubotella: I just removed the `CreateAndInitializeVariableBinding` instruction.

PR #403, opened 2024-08-25
feat(vm): Support the spread operator on array and object literals

andreubotella:
The spread operator on array literals is implemented as a for-of loop that pushes the elements in bytecode. For object literals, however, there is a difference between for-in and spread, in that for-in will not yield symbol property keys. So we instead implement it with a new `CopyDataProperties` instruction.


nova_vm/src/ecmascript/abstract_operations/operations_on_objects.rs
```diff
@@ -766,6 +766,45 @@ pub(crate) fn get_function_realm(
...
+    // 1. If source is either undefined or null, return unused.
+    if source.is_undefined() || source.is_null() {
+        return Ok(());
+    }
+    // 2. Let from be ! ToObject(source).
+    let from = to_object(agent, source).unwrap();
+
+    // 3. Let keys be ? from.[[OwnPropertyKeys]]().
+    // 4. For each element nextKey of keys, do
+    for next_key in from.internal_own_property_keys(agent)? {
```
aapoalas, line 791: **suggestion:** Reserve room in the object for these items using the `reserve` method on the ElementVectors.

aapoalas (approved):
LGTM <3

nova_vm/src/ecmascript/abstract_operations/operations_on_objects.rs
andreubotella: This would mean changing the type of `target` to be `OrdinaryObject`. Will do that in a follow-up PR if you don't mind.

PR #407, opened 2024-08-26
feat(vm): Support the spread operator in function arguments

andreubotella:
The call instructions (`DirectEvalCall`, `EvaluateCall` and `EvaluateNew`) take their arguments from the VM stack, and get the number of arguments from an immediate argument in the instruction. This requires knowing the number of arguments at compile-time, which does not work for the spread operator.

In order to support this operator while using the same instructions, rather than adding new ones, the value `u16::MAX` is treated as a magic constant indicating that the result value should be used as the number of arguments.


nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -1171,84 +1171,76 @@ impl CompileEvaluation for ast::ArrayExpression<'_> {
...
+                ctx.exe.add_instruction(Instruction::GetValue);
+            }
+            if let Some(num_arguments) = known_num_arguments.as_mut() {
+                *num_arguments = num_arguments
+                    .checked_add(1)
+                    .expect("Too many function arguments");
                 ctx.exe.add_instruction(Instruction::Load);
+            } else {
+                // result: value; stack: [num, ...args]
+                ctx.exe.add_instruction(Instruction::LoadStoreSwap);
```
aapoalas, line 1225: **praise:** Nifty trick here!

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -1171,84 +1171,76 @@ impl CompileEvaluation for ast::ArrayExpression<'_> {
...
+    // If the arguments don't contain the spread operator, then we can know the
+    // number of arguments at compile-time and we can pass it as an argument to
+    // the call instruction.
+    // Otherwise, the first time we find a spread operator, we need to start
+    // tracking the number of arguments in the compiled bytecode. We store this
+    // number in the result value, and we pass u16::MAX to the call instruction.
+    let mut known_num_arguments = Some(0 as IndexType);
+
+    for argument in arguments {
+        // If known_num_arguments is Some, the stack contains the number of
```
aapoalas, line 1184: **nitpick:** Isn't this the other way around: If it's `None` then we've already loaded the number of arguments on the stack, if it's Some then we've yet to do that.

nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -1171,84 +1171,76 @@ impl CompileEvaluation for ast::ArrayExpression<'_> {
...
+                .add_jump_instruction_to_index(Instruction::Jump, iteration_start);
+            ctx.exe.set_jump_target_here(iteration_end);
+        } else {
+            let expression = argument.to_expression();
+            expression.compile(ctx);
+            if is_reference(expression) {
+                ctx.exe.add_instruction(Instruction::GetValue);
+            }
+            if let Some(num_arguments) = known_num_arguments.as_mut() {
+                *num_arguments = num_arguments
```
aapoalas, line 1219: **issue (nitpicky):** If this reaches the max value, then we should already move to tracking the value on the stack as we can't tell the difference.

aapoalas (approved):
Looks good to me! There's a corner case with `u16::MAX` parameters but that's kinda dumb anyway :D

nova_vm/src/engine/bytecode/executable.rs
andreubotella: Fixed.

nova_vm/src/engine/bytecode/executable.rs
andreubotella: Fixed.

aapoalas (approved):
LGTM, and thank you <3

PR #415, opened 2024-09-14
build: add `justfile`

DonIsaac:
## What This PR Does

Adds a [justfile](https://github.com/casey/just?tab=readme-ov-file#just) with some shortcuts of common tasks. I've only added a few for now.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,17 @@
+_default: 
+    @just --list -u
+
+# Run the Test262 test suite. Run `just test262 --help` for more information.
+test262 *ARGS:
+    cargo run --bin test262 -- {{ARGS}}
+
+# Start a REPL session
+repl:
+    cargo run --bin nova_cli -- repl
+
+# Fix Clippy, rustfmt, and typos issues
+fmt:
+    cargo clippy --no-deps --all-targets --fix --allow-staged --allow-dirty
+    cargo fmt --all
+    typos -w
+    git status


aapoalas:
Thank you for the PR. Documenting these commands is _very good_ and I'm sorry you (probably) had to figure them out on your own.

That being said, what do you think about moving these to the `CONTRIBUTING.md` file instead of adding a `justfile`? I've personally never heard of `just` and I'm not sure any other active contributors are using it either; this would basically be a file by you, for you, and you alone at least for now.

eliassjogreen:
I have to agree with @aapoalas that documenting these commands are very good, but should probably not be done with a `justfile`. I would probably not be in favor of these kinds of shortcuts even if they were done with `make` or something modern like `task` or `just`. But adding a section about common commands to the `CONTRIBUTING.md` would be very useful, I remember the pain of digging through the code before I figured out how to run (and update) the tests. 😅

DonIsaac:
Documentation would certainly be useful. Do you guys not get tired of writing all those words all the time? I personally find task runners quite helpful.

aapoalas:
> Do you guys not get tired of writing all those words all the time? I personally find task runners quite helpful.

I personally use shell Ctrl+R search and Up key to jump to the latest command that starts with whatever I've filled in.

So eg. `c` + Up a few times usually gives me what I want :)

DonIsaac:
Makes sense. This behavior is quite annoying when using zsh with  multiple terminal windows open. This is, however, a me problem and if nobody else would find this useful we should close this PR.

PR #416, opened 2024-09-14
feat(ecmascript): `String.fromCharCode`

DonIsaac:


nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -112,12 +116,35 @@ impl StringConstructor {
...
+        //   a. Let nextCU be the code unit whose numeric value is ℝ(? ToUint16(next)).
+        //   b. Set result to the string-concatenation of result and nextCU.
+        for next in code_units.iter() {
+            let code_unit = next.to_uint16(agent)?;
+            result.extend_from_slice(&code_unit.to_ne_bytes());
+        }
+
+        // 3. Return result.
+        let result_str = str::from_utf8(result.as_slice()).unwrap();
+        let result = SmallString::try_from(result_str)
```
DonIsaac, line 144: needs (nonblocking): `String::from_str` always allocates strings on the heap. It should attempt to inline strings that are small enough.

DonIsaac: Additionally, `SmallString` should expose a `MAX_LEN` constant. We could then remove all the `8` magic numbers floating around the codebase.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -112,12 +116,35 @@ impl StringConstructor {
...
+        // 2. For each element next of codeUnits, do
+        //   a. Let nextCU be the code unit whose numeric value is ℝ(? ToUint16(next)).
+        //   b. Set result to the string-concatenation of result and nextCU.
+        for next in code_units.iter() {
+            let code_unit = next.to_uint16(agent)?;
+            result.extend_from_slice(&code_unit.to_ne_bytes());
+        }
+
+        // 3. Return result.
+        let result_str = str::from_utf8(result.as_slice()).unwrap();
```
DonIsaac, line 143: issue: what is the correct behavior here? Do we have a way of casting structs implementing `std::error::Error` to `JsError`?

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
aapoalas: It's a bit convoluted but `from_str` does indeed check and return a `SmallString` if it is possible.

A `MAX_LEN` constant would definitely be a good thing, although we actually have only _one_ place where we still check for `< 8` outside of the `small_string.rs` itself. Having the constant be abstracted would make it possible (in an easier way anyway) to generalise `SmallString` to eg. 3 bytes.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
aapoalas: We indeed do not have that, and at the engine level we also shouldn't really need one. Any errors that the engine produces should be just plain JS `TypeError`, `SyntaxError` etc. A runtime would definitely want to have something like that, though.

The correct thing here would probably be to create a `Wtf8Buf`, though that may be a bit trickier.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -112,12 +116,35 @@ impl StringConstructor {
...
+        if code_units.is_empty() {
+            return Ok(String::EMPTY_STRING.into_value());
+        }
+        let mut result: Vec<u8> = Vec::with_capacity(code_units.len() * 2);
+
+        // 2. For each element next of codeUnits, do
+        //   a. Let nextCU be the code unit whose numeric value is ℝ(? ToUint16(next)).
+        //   b. Set result to the string-concatenation of result and nextCU.
+        for next in code_units.iter() {
+            let code_unit = next.to_uint16(agent)?;
```
aapoalas, line 138: **issue:** You've effectively only performed the `R(? ToUint16(next))` part but are missing the `be the code unit` step. This is referring to the Unicode code units, so effectively a piece of text. This is probably the API you really want to be using here: https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf16_lossy

DonIsaac: I spent an hour looking for that API... thank you!

nova_vm/src/ecmascript/types/language/value.rs
```diff
@@ -443,6 +443,14 @@ impl Value {
         to_uint32(agent, self)
     }
 
+    pub fn to_int16(self, agent: &mut Agent) -> JsResult<i16> {
```
aapoalas, line 446: **issue(nonblocking):** I don't think these functions are really worth exposing through the `Value`'s public API, or on the `Value` API at all.

The spec defines `ToUint16` as an abstract operation that takes a `Value`, not as a method of `Value`. Hence, I'd prefer matching spec call logic so when the spec says `ToUint16(value)` we do `to_uint16(agent, value)`

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -112,12 +116,35 @@ impl StringConstructor {
...
         _this_value: Value,
-        _arguments: ArgumentsList,
+        code_units: ArgumentsList,
     ) -> JsResult<Value> {
-        todo!();
+        // 1. Let result be the empty String.
+        if code_units.is_empty() {
+            return Ok(String::EMPTY_STRING.into_value());
+        }
+        let mut result: Vec<u8> = Vec::with_capacity(code_units.len() * 2);
```
aapoalas: **suggestion:** Add a `code_units.len() == 1 && let Value::Integer(code_unit) = args.get(0)` fast path ... This could probably actually use `std::char::from_u32(code_unit.into_i64() as u16 as u32)` to get a single char... And something from there? :D

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -112,12 +116,35 @@ impl StringConstructor {
         Ok(s.into_value())
     }
 
+    /// ### [22.1.2.1 String.fromCharCode ( ...`codeUnits` )](https://262.ecma-international.org/15.0/index.html#sec-string.fromcharcode)
+    ///
+    /// > This function may be called with any number of arguments which form
```
aapoalas, line 120: **nitpick:** I've used "quotes" `>` for "NOTE"s that appear in the spec. The text here is just description of the function, and should just appear normally without the "quoting".

aapoalas (changes requested):
Thank you very much and welcome to Nova! This looks like a great start though some of the details regarding string creation are currently off. That aside, this is a very nice function to have done! <3

nova_vm/src/ecmascript/types/language/value.rs
DonIsaac: Would this also then apply to `to_uint32` and `to_int32`?

aapoalas: Yeah.

aapoalas:
Assuming tests pass, LGTM! Though I _would_ love to see the single-integer-char-code fast path, possibly with just for ASCII char codes to be even more trivial to implement. That would be doable with just `i8::try_from(integer.into_i64() as u16)`.

DonIsaac:
> Assuming tests pass, LGTM! Though I _would_ love to see the single-integer-char-code fast path, possibly with just for ASCII char codes to be even more trivial to implement. That would be doable with just `i8::try_from(integer.into_i64() as u16)`.

Just added it! Mind queueing another CI run please?

aapoalas (approved):
LGTM! Thank you very much <3

aapoalas (approved):

PR #417, opened 2024-09-17
feat(ecmascript): `String.prototype.split()`

oscarotero:
Please, bear with me. I'm learning Rust and I don't have much experience with the borrow checker.


oscarotero:
@aapoalas I think it's ready to review.

btw, what is the recommended way to test the function? I'd like to have a file to execute (like `cargo run -- eval testing.js`) to manually test the function, but there isn't a `console.log()` that I can use to see the result.

aapoalas:
`nova_cli` has a `print` function you can use.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -732,8 +733,73 @@ impl StringPrototype {
...
+        let s = to_string(agent, o)?;
+
+        // 3. Let isRegExp be ? IsRegExp(searchString).
+        // 4. If isRegExp is true, throw a TypeError exception.
+        // TODO
+
+        // 5. Let separatorStr be ? ToString(separatorString).
+        let separator_value = args.get(0);
+
+        // 6. If separator is undefined, return an array with this.
```
aapoalas, line 749: **issue (nonblocking):** This should be spec step 7, and this is also not the spec text.

I won't block this PR on this, but in the future please leave the spec text in the code mostly as-is. Commenting on the spec text with eg. "Note: ..." is fine. Moving spec text and steps around is also fine as long as it doesn't change any observable features of the function. Oftentimes leaving a note about these sort of reorderings is a good idea.

But rewriting the spec text in your own words, and then in the process having it turn into just your custom algorithm with steps outlined is not a good thing. If someone else comes to look at this function and tries to figure out how this corresponds to the spec algorithm, they'll have very little idea what is going on. Maybe this is written based on some old spec text that is no longer available online? And in that case, what has changed? Presumably the algorithm has been changed in meaningful ways and then this old code written to the old spec is no longer valid; but how?

It's better to abide by the spec text as much as possible, and only make additions (fast paths) and minor reorderings where it makes sense. That way a future reader (which may be you yourself) has an easier time understanding what is being done and why, and if the spec changes they'll also know what they need to change to fix the code.

aapoalas:
Oh, oops. I only now realized this is actually your first PR here! Sorry, I wasn't paying attention and thought you were behind #416 as well!

Welcome, and great work getting yourself up from the ground with the borrow checker! And sorry about the nitpicking on the spec text: It's just a good idea to stick fairly close to it. I don't mean it personally <3

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: Sorry, I didn't realize these comments were spec comments. I though they were simple comments to follow easier what the code does.
I'll take a look to that.

oscarotero:
Thanks for your review @aapoalas 
Writing a JS engine is harder than I thought :D

I'm learning the codebase of the repo, how to implement and test things, and now I just learned that the numbers in the comments corresponds to the steps in the ecma262 specs.

I'll continue working on this.

aapoalas:
> I'm learning the codebase of the repo, how to implement and test things, and now I just learned that the numbers in the comments corresponds to the steps in the ecma262 specs.

I'm sorry! This is definitely my failing! I should add some more about this to the CONTRIBUTING.md and maybe elsewhere as well so that new contributors have an easier way in.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -733,73 +734,91 @@ impl StringPrototype {
...
+        // 3. Let S be ? ToString(O).
+        let s = to_string(agent, o)?;
 
-            return Ok(results.into_value());
-        }
+        // 4. If limit is undefined, lim is u32::MAX - 1.
+        let limit = args.get(0);
+        let lim = match limit {
+            Value::Undefined => u32::MAX,
+            _ => to_uint32(agent, limit)?,
```
aapoalas, line 756: **suggestion:** Add a `Value::Integer(value) => value.into_i64() as u32`

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -733,73 +734,91 @@ impl StringPrototype {
...
-            let slice: [Value; 1] = [this_value];
-            let results = Array::from_slice(agent, &slice);
+        // 3. Let S be ? ToString(O).
+        let s = to_string(agent, o)?;
 
-            return Ok(results.into_value());
-        }
+        // 4. If limit is undefined, lim is u32::MAX - 1.
+        let limit = args.get(0);
+        let lim = match limit {
```
aapoalas, line 754: **praise:** I've not actually thought about combining spec `if undefined { X } else { Y }` into matches like this. Makes for fairly pretty code and makes it possible to add fast paths pretty trivially. Nice! <3

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -733,73 +734,91 @@ impl StringPrototype {
...
+            Value::Undefined => u32::MAX,
```
aapoalas, line 755: **issue:** Should be `u32::MAX - 1`

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -733,73 +734,91 @@ impl StringPrototype {
...
-        // 6. If separator is undefined, return an array with this.
-        if let Value::Undefined = separator_value {
-            let slice: [Value; 1] = [this_value];
-            let results = Array::from_slice(agent, &slice);
+        // 3. Let S be ? ToString(O).
+        let s = to_string(agent, o)?;
 
-            return Ok(results.into_value());
-        }
+        // 4. If limit is undefined, lim is u32::MAX - 1.
```
aapoalas, line 752: **nitpick:** You're still paraphrasing the spec. I'd just copy it verbatim:

```rs
let lim = match limit {
  // 4. If limit is undefined, let lim be 2**32 - 1;
  Value::Undefined => u32::MAX - 1,
  // else let lim be ℝ(? ToUint32(limit)).
  // Note: Fast path for integer parameter.
  Value::Integer(limit) => limit.into_i64() as u32,
  _ => to_uint32(agent, limit)?,
};
```

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: The spec says `2**32 - 1`. I don't know if `u32::MAX` is that number or it should substract 1.
(I know the comment is wrong)

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -733,73 +734,91 @@ impl StringPrototype {
...
+            let list: [Value; 0] = [];
+            return Ok(create_array_from_list(agent, &list).into_value());
+        }
 
-        // 8. If the limit is zero, returns an empty array
-        if limit.is_pos_zero(agent) || limit.is_neg_zero(agent) {
-            let a = array_create(agent, 0, 0, None)?;
-            return Ok(a.into_value());
+        // 7. If separator is undefined, return an array with the whole string
+        if let Value::Undefined = separator {
```
aapoalas, line 769: **nitpick:** Could just use `separator.is_undefined()` here.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -733,73 +734,91 @@ impl StringPrototype {
...
+            let list: [Value; 1] = [s.into_value()];
```
aapoalas, line 773: **nitpick:** You can just do `return Ok(create_array_from_list(agent, &[s.into_value()]);` to avoid having to define the array type.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: okay, fixed!

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: happy to know at least I did something well! :D

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: fixed

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -733,73 +734,91 @@ impl StringPrototype {
...
 
-        // 10. The limit must be positive integer or 0
-        let limit = if limit.is_sign_positive(agent) {
-            limit.into_i64(agent) as usize
-        } else {
-            0
-        };
+        // 9. If separatorLength = 0, the split by characters
+        if separator_length == 0 {
+            let subject = s.as_str(agent).to_owned();
```
aapoalas, line 782: **nitpick:** I would leave in the spec steps a through e, BUT also **praise:** Explaining that this is doing a split by characters is very nice!

**suggestion:** I _think_ we're guaranteed that all of the characters are necessarily expressible within 7 bytes as UTF-8. I think so anyway. As such, I think you can actually avoid the `to_owned()` here by just asserting that each split part will pass `SmallString::try_from(part).unwrap()`.

There would also be a way to avoid the Vec altogether and just shuffle data within the heap directly, but we're missing a `ArrayHeapMutIndexable` struct for now. `map_constructor.rs:172` and the `ArrayHeapIndexable` struct may serve as hints as to what I'm thinking about here. Just add in the knowledge that any UTF-16 character (which is what we should be splitting the string into) should fit into 7 bytes when converted into UTF-8, I think, and all should be "clear" :D

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -733,73 +734,91 @@ impl StringPrototype {
...
+            let head = subject.split("");
```
aapoalas, line 783: **though:** I wonder if this does exactly the correct thing: The spec deals with UTF-16, but we're in UTF-8 world here. That makes things complicated sometimes. @andreubotella might understand this more deeply.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -733,73 +734,91 @@ impl StringPrototype {
...
+                results.push(Value::from_str(agent, part));
             }
-            results.push(Value::from_str(agent, part));
-        }
 
-        // 12. If separator is empty, Rust's split inserts an empty string in the beginning and end of the array
-        if separator.is_empty() {
+            // Note: Rust's split inserts an empty string in the beginning and end of the array. We remove them
             if results.len() > 1 {
                 results.remove(0);
```
aapoalas, line 795: **suggestion:** It would be better to do this "filtering" out as part of the enumeration itself.

You can probably do something like: `let results: Vec<Value> = head.enumerate().skip(1).take_while(lim).map(|part| Value::from_str(agent, part)).collect();`. Now when creating this from slice, just drop the last item if the length of the Vec is longer than 0.

aapoalas (approved):
Found a bunch to comment on, but nothing terribly out of whack. Thank you very much, and welcome again to Nova!

Absolutely feel free to keep on trying to solve any remaining cases: The `eval-test` command is very useful for testing individual test262 case. eg.
```sh
cargo build && cargo run --bin test262 eval-test built-ins/String/prototype/split/argument-is-new-reg-exp-and-instance-is-string-hello.js
```

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: done

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: I've refactored this part.

oscarotero:
@aapoalas I have fixed a lot of tests due a stupid bug: I was getting the limit argument with `args.get(0)` instead of `args.get(1)`. The only failing tests are those related with regexp, `Symbol.split` and other advanced edge cases.

I've also applied your suggestions. The only unresolved comment is the one related with UTF-16.

oscarotero:
No problem! I saw the pictures and completely understand your priorities! 🍇

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -732,8 +734,87 @@ impl StringPrototype {
...
+    /// ### [22.1.3.23 String.prototype.split ( separator, limit )](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.split)
+    fn split(agent: &mut Agent, this_value: Value, args: ArgumentsList) -> JsResult<Value> {
+        // 1. Let O be ? RequireObjectCoercible(this value).
+        let o = require_object_coercible(agent, this_value)?;
+
+        // 2. If separator is neither undefined nor null, then
+        let separator = args.get(0);
+
+        if matches!(separator, Value::Undefined | Value::Null) {
+            // TODO: get_method(separator, Symbol.split)
```
aapoalas, line 746: **issue:** This shouldn't be too hard: `get_method` function already exists, and then you just perform a `call_function(splitter, separator, Some(&[o, limit]))`

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -732,8 +734,87 @@ impl StringPrototype {
...
+        // 9. If separatorLength = 0, the split by characters
+        if separator_length == 0 {
+            let subject = s.as_str(agent);
+            let head = subject.split("");
+
+            let mut results: Vec<Value> = head
+                .enumerate()
+                .skip(1) // Rust's split inserts an empty string in the beginning.
+                .take_while(|(i, _)| *i < lim as usize)
+                .map(|(_, part)| Value::from(SmallString::try_from(part).unwrap()))
```
aapoalas, line 787: **nitpick:** Prefer `.into_value()` over `Value::from()`.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -732,8 +734,87 @@ impl StringPrototype {
...
+                .collect();
+
+            // Remove the latest empty string if exists
+            results.pop();
```
aapoalas, line 791: **question:** Are you sure this works currectly? What happens if limit is less than the length of the iterator? Won't this then drop one string that was supposed to be received?

aapoalas (approved):
Minor nitpicks / thoughts / questions.

But looks good to me! Thank you for this, great work!

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: This case only happens when the separator is an empty string. You can see the documentation of `split()`:

```rs
/// When the empty string is used as a separator, it separates
/// every character in the string, along with the beginning
/// and end of the string.

let f: Vec<_> = "rust".split("").collect();
assert_eq!(f, &["", "r", "u", "s", "t", ""]);
```

In JavaScript `"rust".split("")` must return `["r", "u", "s", "t"]`, so that's why I remove the first and last elements. The first is removed in  `skip(1)`. The last is removed with `results.pop()`, that, according to the docs:
```
Removes the last element from a vector and returns it, or [`None`] if it is empty.
```

I think this code should work fine (all related tests are passed), but I'm open to change it if you want to implement it differently.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: I didn't implement this because I don't know how to create the `Symbol.split` value in Rust.
But if you help me with that, I can implement this.

aapoalas: Ah, these "well-known symbols" can be "created" as simply as `WellKnownSymbolIndexes::Split.into()` :)

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
aapoalas: Yeah, but now what happens if we do : `"rust".split("", 2)`. This should return `["r", "u"]` but I _think_ it will currently return just `["r"]` because we drop the last one regardless of the limit.

oscarotero: That's a good point. And, indeed, it doesn't work. I'm fixing it.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: Done in the last commit.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: Ok, I think it's now fixed:
- The `take_while` was wrong because it didn't take into account the skip() done before.
- Only removes the lastest string if the limit is bigger than the result length.

It's weird that there isn't any test for this use case. I would have detected earlier.

aapoalas (approved):

aapoalas:
Thank you again and great work! <3

PR #419, opened 2024-09-21
feat(vm): Store compiled bytecode in ECMAScriptFunctionHeapData

aapoalas:
Entry step towards supporting class instance field initializers.


nova_vm/src/engine/bytecode/executable.rs
```diff
@@ -72,7 +70,6 @@ impl CompileContext<'_> {
                 instructions: Vec::new(),
                 constants: Vec::new(),
                 identifiers: Vec::new(),
-                references: Vec::new(),
```
aapoalas, line 75: **note:** I noticed this isn't being used at all.

PR #424, opened 2024-09-28
`String.prototype.replace`

oscarotero:
Here I'm again!
This time, I hope you don't have to spend so much time doing reviewing :P


nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -645,8 +645,85 @@ impl StringPrototype {
...
+        // 1. Let O be ? RequireObjectCoercible(this value).
+        let o = require_object_coercible(agent, this_value)?;
+
+        let search_value = args.get(0);
+        let replace_value = args.get(1);
+
+        // 2. If searchValue is neither undefined nor null, then
+        if !search_value.is_null() && !search_value.is_undefined() {
+            let symbol = WellKnownSymbolIndexes::Replace.into();
+            // If replacer is not undefined, Return ? Call(replacer, searchValue, « O, replaceValue »).
```
aapoalas, line 659: **issue:** We're missing the comment that explains how `replacer` is to be found. (I'm quite interested in that because I hope to god that we don't actually have to look up the replacer function for strings as well :cry: )

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -645,8 +645,85 @@ impl StringPrototype {
...
+                    agent,
+                    replacer,
+                    search_value,
+                    Some(ArgumentsList(&[o, replace_value])),
+                );
+            }
+        }
+
+        // 3. Let s be ? ToString(O).
+        let s = to_string(agent, o)?;
```
aapoalas, line 671: **suggestion:** Here we could do an "expected" early path:

```rs
if let (Ok(s), Ok(search_string), Ok(replace_string)) = (String::try_from(o), String::try_from(search_value), String::try_from(replace_value)) { ... } else { ... }
```
The expected path this function gets called is something like `"asdbarbazfoobot".replace("bar", "foo")`. So, we can skip the is_callable etc. tests if we find ourselves on this "normal" path.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -645,8 +645,85 @@ impl StringPrototype {
...
+        // 8. Let position be StringIndexOf(s, searchString, 0).
+        // 9. If position is not-found, return s.
+        let position = if let Some(position) = search_string.as_str(agent).find(s.as_str(agent)) {
+            position
+        } else {
+            return Ok(s.into_value());
+        };
+
+        // Use rust's replace if the replace is a string, not a function
+        if let Some(replace_value) = replace_value {
```
aapoalas, line 697: **isse:** Move the `position` calculation from above to below this if check.

aapoalas: And actually, that means that you can move this entirely into the None path of the `match functional_replace` statement.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -645,8 +645,85 @@ impl StringPrototype {
...
+        // 10. Let preceding be the substring of s from 0 to position.
+        // 11. Let following be the substring of s from position + searchLength.
+        // 12. If functionalReplace is true,
+        let preceding = &s.as_str(agent)[0..position].to_owned();
+        let following = &s.as_str(agent)[position..search_length].to_owned();
+
+        // Let replacement be ? ToString(? Call(replaceValue, undefined, « searchString, 𝔽(position), string »)).
+        let result = call(
+            agent,
+            functional_replace.unwrap().into_value(),
```
aapoalas, line 713: **note:** This unwrap here kind of points to this function probably benefiting from a bit of a refactor: It would be nice if the functional_replace and/or replace_value path could be done entirely within a `match` branch or perhaps a `if let Some(..)/Ok(..) = .. {}` branch.

aapoalas (changes requested):
Welcome back! Looking really good, but the gloves are now off so now I'm marking **issue** more eagerly :) The code seems to be entirely correct and functional, just some (very minor) issues that I'd like to see changed.

But absolutely looks good!

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: I've added the comment with the instructions from the spec. The replacer is found with the `get_method` function.

oscarotero:
@aapoalas Okay, I've refactored the function, and now the logic is inside the match of step 5.

I've also moved the "exected" early path to the top, but this logic is also duplicated in  the point 6. So I'm not sure if this is a good thing.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -645,8 +645,99 @@ impl StringPrototype {
...
+            String::try_from(replace_value),
+        ) {
+            let result = s
+                .as_str(agent)
+                .replace(search_string.as_str(agent), replace_string.as_str(agent));
+            return Ok(String::from_string(agent, result).into_value());
+        }
+
+        // 2. If searchValue is neither undefined nor null, then
+        if !search_value.is_null() && !search_value.is_undefined() {
```
aapoalas, line 669: **issue:** We can't actually move the fast path to be before this step :( 

```
String.prototype[Symbol.replace] = () => "test";
"foobar".replace("") // "test"
```

Though optionally we can do this check on the fast-path with a `get_method` operating on `agent.current_realm().intrinsics().string_prototype()` instead of on `search_string`. That's still a somewhat valuable skip :)

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -645,8 +645,99 @@ impl StringPrototype {
...
+        }
+
+        // 3. Let s be ? ToString(O).
+        let s = to_string(agent, o)?;
+
+        // 4. Let searchString be ? ToString(searchValue).
+        let search_string = to_string(agent, search_value)?;
+
+        // 5. Let functionalReplace be IsCallable(replaceValue).
+        match is_callable(replace_value) {
```
aapoalas, line 692: **nitpick:** I'd do this as
```rs
if let Some(functional_replace) = is_callable(replace_value) {
 // Some branch here, step 7 etc. Ends with a return.
}
// None branch but now with less None matching visible
let replace_string = ...;
```
But this way is of course _absolutely_ fine. No need to change, this is just me nitpicking for the sake of it I guess :D

aapoalas (approved):
LGTM! Nice and clean code, thank you very much! <3

oscarotero:
I made the following changes:

- Removed the fast path before step 2. It doesn't make sense to move it because the same situation is checked after that (if the step 5 is false).
- The step 5 now follows your suggestion. I'm agree it's more elegant and, in fact, I usually to do the same (an `if` with an early return).

BTW, most of the tests use regular expressions (`"foo".replace(/o/, "a")`), but the spec doesn't mention anything about how to handle them. Since regexps are not supported right now in Nova, I think the current implementation is okay, but maybe we should include a `todo` macro somewhere.

aapoalas:
The RegExp support comes from the `Symbol.replace` path. That's (unfortunately) how RegExp replacing is implemented.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -645,8 +645,83 @@ impl StringPrototype {
...
+                position
+            } else {
+                // 9. If position is not-found, return s.
+                return Ok(s.into_value());
+            };
+
+            // 10. Let preceding be the substring of s from 0 to position.
+            // 11. Let following be the substring of s from position + searchLength.
+            // 12. If functionalReplace is true,
+            let preceding = &s.as_str(agent)[0..position].to_owned();
```
aapoalas, line 696: **nitpick:** You can move these after the `call`; that avoids doing any allocations if the `functional_replace` throws an error.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -645,8 +645,83 @@ impl StringPrototype {
...
+            let following = &s.as_str(agent)[position..search_length].to_owned();
+
+            // Let replacement be ? ToString(? Call(replaceValue, undefined, « searchString, 𝔽(position), string »)).
+            let result = call(
```
aapoalas, line 700: **issue:** We have a `call_function` helper that skips some doing `is_callable` again.

aapoalas (changes requested):
Some issues I realized on a yet another review, sorry :( But the format of the function is now _really_ neat. Really nice! <3

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: done!

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: good point!

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -645,8 +645,83 @@ impl StringPrototype {
...
+        let o = require_object_coercible(agent, this_value)?;
+
+        let search_value = args.get(0);
+        let replace_value = args.get(1);
+
+        // 2. If searchValue is neither undefined nor null, then
+        if !search_value.is_null() && !search_value.is_undefined() {
+            // a. Let replacer be ? GetMethod(searchValue, %Symbol.replace%).
+            let symbol = WellKnownSymbolIndexes::Replace.into();
+            let replacer = get_method(agent, search_value, symbol);
```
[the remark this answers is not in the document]
oscarotero: good catch.

aapoalas (approved):
You need to redo the expectations.

Aside from that, this looks really good to me! Thank you and good job! <3

oscarotero:
Test updated. Sorry, for not updating them before, my computer is a bit old, it takes a long time to complete all the tests and sometimes the results doesn't match with GitHub result (mainly timeouts).
I'd love to have a way to run and update the test metrics of a specific feature.
Something like:

```
cargo run --bin test262 built-ins/String/prototype/split/ -- --update`
```

aapoalas:
> I'd love to have a way to run and update the test metrics of a specific feature. Something like:
> 
> ```
> cargo run --bin test262 built-ins/String/prototype/split/ -- --update`
> ```

That sounds like it would be quite useful sometimes.

aapoalas:
Thank you again and great work! <3

PR #428, opened 2024-10-06
String/replaceall

oscarotero:
Hi again 👋 

In this PR I've implemented the `String.prototype.replaceAll` function.

During the implementation I've discovered some bugs in `String.prototype.replace` that are also fixed, specifically:
- The Rust's `replace` function works like `replaceAll` in javascript (replacing all occurrences). I've changed to `replacen(search, replace, 1)` to replace only the first occurence.
- The position calculation was inversed (`search.find(subject)` instead of `subject.find(search)`).


nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -718,16 +718,140 @@ impl StringPrototype {
...
+                        Number::from(position as u32).into_value(),
+                        s.into_value(),
+                    ])),
+                )?;
+
+                // d. Set result to the string-concatenation of result, preserved, and replacement.
+                let replacement_str = replacement.to_string(agent)?;
+                let replacement_str = replacement_str.as_str(agent);
+
+                let concat = format!("{}{}", &preserved, &replacement_str);
```
aapoalas, line 831: **note:** Once you move the `preserved` to be calculated only after the functional_replace call (and the following `to_string` call) it then probably becomes better to first call `result.reserve(preserved.len() + replacement_str.len()` and then individually do two `push_str` calls instead of using `format`.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -718,16 +718,140 @@ impl StringPrototype {
...
+            // 12. Let endOfLastMatch be 0.
+            let mut end_of_last_match = 0;
+
+            // 13. Let result be the empty String.
+            let mut result = "".to_string();
+
+            // 14. For each element p of matchPositions, do
+            for p in match_positions {
+                // a. Let preserved be the substring of string from endOfLastMatch to p.
+                let preserved = subject[end_of_last_match..p].to_owned();
```
aapoalas, line 813: **issue:** We can move this after the functional_replace call without the change being observable by callers, I believe. That allows us to also avoid the `to_owned` call.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -718,16 +718,140 @@ impl StringPrototype {
...
+                } else {
+                    break;
+                };
+            }
+
+            // 12. Let endOfLastMatch be 0.
+            let mut end_of_last_match = 0;
+
+            // 13. Let result be the empty String.
+            let mut result = "".to_string();
```
aapoalas, line 808: **suggestion:** It's probably better to do here `std::string::String::with_capacity(subject.len())`

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -718,16 +718,140 @@ impl StringPrototype {
...
+                let replacement_str = replacement_str.as_str(agent);
+
+                let concat = format!("{}{}", &preserved, &replacement_str);
+                result.push_str(&concat);
+                end_of_last_match = p + search_length;
+            }
+
+            // 15. If endOfLastMatch < the length of string, set result to the string-concatenation of result and the substring of string from endOfLastMatch.
+            if end_of_last_match < subject.len() {
+                let preserved = subject[end_of_last_match..].to_owned();
```
aapoalas, line 838: **issue:** `to_owned` here should be unnecessary as far as I understand.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -718,16 +718,140 @@ impl StringPrototype {
...
+                result.push_str(&preserved);
+            }
+
+            // 16. Return result.
+            return Ok(String::from_str(agent, &result).into_value());
```
aapoalas, line 843: **issue:** This makes an unnecessary copy, as you pass a borrow of an owned string to Nova's heap; the heap has to make a copy of the string data. You then drop that data.

You can pass ownership of an owned string to the heap using the `from_string` function instead. That way the string doesn't get dropped at all and the heap doesn't need to make that copy.

aapoalas (changes requested):
Some optimisations that can be made but looks absolutely good to me aside from a correctness standpoint. Let's make those changes and get this merged!

Thank you and great work! <3

load1n9 (approved):
LGTM

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -718,16 +718,135 @@ impl StringPrototype {
...
+            let search_length = search_string.len(agent);
+
+            // 8. Let advanceBy be max(1, searchLength).
+            let advance_by = max(1, search_length);
+
+            // 9. Let matchPositions be a new empty List.
+            let mut match_positions: Vec<usize> = vec![];
+
+            // 10. Let position be StringIndexOf(s, searchString, 0).
+            let search_str = search_string.as_str(agent).to_owned();
```
aapoalas, line 781: **nitpick:** Best as I can see, the `to_owned` is not actually needed here at all. `search_str` is not used after the while loop over the positions which also doesn't do any mutations on `agent`, so `search_str` could just be a borrowed string from the heap.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -718,16 +718,135 @@ impl StringPrototype {
...
+            let subject = s.as_str(agent).to_owned();
```
aapoalas, line 782: **nitpick:** It would actually also be possible to avoid this copy by just reborrowing `s.as_str(agent)` after each `functional_replace` call.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -718,16 +718,135 @@ impl StringPrototype {
...
+
+            let mut position = if let Some(position) = subject.find(&search_str) {
```
aapoalas, line 784: **nitpick:** You could combine this and the below `while` loop and return `Ok(s.into_value())` if `match_positions.is_empty()`.

oscarotero:
@aapoalas thanks! Don't merge it yet. I'll make the changes, just need a bit more time (I'm busy right now)

aapoalas:
> @aapoalas thanks! Don't merge it yet. I'll make the changes, just need a bit more time (I'm busy right now)

Ah, alright! Don't worry about it <3

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -718,16 +718,135 @@ impl StringPrototype {
...
+        // 1. Let O be ? RequireObjectCoercible(this value).
+        let o = require_object_coercible(agent, this_value)?;
+
+        let search_value = args.get(0);
+        let replace_value = args.get(1);
+
+        // 2. If searchValue is neither undefined nor null, then
+        if !search_value.is_null() && !search_value.is_undefined() {
+            // a. Let isRegExp be ? IsRegExp(searchValue).
+            let is_reg_exp = false;
```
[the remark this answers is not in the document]
oscarotero: I think this goes beyong my knowledge.
Maybe [I can try to implement this](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-isregexp) a different PR but I would need some guidance.

I did in this way because doing a quick search or `todo!()` macro, you can find all remaining things to implement. So it's not lost.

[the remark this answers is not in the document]
aapoalas: Yeah, it's not completely lost so this is fine :)

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: Yep. I don't remember why I had to add it, but it's deleted now and it works.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: Good point. Due I try to follow step-by-step the specs, sometimes I miss these small improvements.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: I'm not able to remove the `to_owned()` here and not having errors in other places. I think this is the typical borrow checker madness that I cannot understand.

imaxe

aapoalas: So the borrow checker is (sort of) absolutely correct here: When we do `s.as_str(agent)` it means we borrow a `&str` from the agent's heap. When we then call `call_function(agent,  ...)` below we give exclusive access to `agent` and its heap into the `call_function` call. For all that the borrow checker knows, this could remove the `&str` that we're borrowing and hence it would be a use-after-free to allow us to use that borrow later. (We actually kind of have guarantees that the string data won't move so it _would_ theoretically be possible to make this sort of borrow safe, but I don't know how that would be nicely explainable to the borrow checker.)

Hence the "reborrow" part; we currently have a guarantee that the `s` "JS String" never changes identity, so calling `s.as_str(agent)` will always give us the same `&str` borrow (unless we have a bug in the system; we might, but right now it's unlikely here). So, we could replace each `&subject` with `s.as_str(agent)` and each `&search_str` with `search_string.as_str(agent)` to effectively "reborrow" the `&str` each time it is used. With this the borrow checker won't have anything to complain about, though it means that we exchange a heap allocation for more pointer loads. The pointer load targets should be in L1 cache though, so this should be a fairly good exchange.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
aapoalas: Yeah, that's definitely the downside of going "by the book".

I actually updated the [CONTRIBUTING.md](https://github.com/trynova/nova/blob/main/CONTRIBUTING.md) yesterday and added text about following the spec, but I forgot to add in anything about when it is okay to step away from the spec (mainly, when it cannot be detected by a user).

aapoalas (approved):

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
oscarotero: Okay, thanks for the detailed explanation. It makes sense now!
As I see you already merged this PR, I assume the current code is also good, and your suggestion is just a nice to have improvement.

aapoalas: > Okay, thanks for the detailed explanation. It makes sense now! As I see you already merged this PR, I assume the current code is also good, and your suggestion is just a nice to have improvement.

Ah yeah, it was all mostly unimportant nitpicking and minor improvements in the first place. If you want to, you can open a PR with the changes. But the code is absolutely good as it already is, thank you very much! <3

PR #433, opened 2024-10-10
feat: array buffer feature

load1n9:


aapoalas (dismissed):
We do need to make DataView and TypedArrays require this feature.

aapoalas (approved):

PR #436, opened 2024-10-11
feat: shared arraybuffer as feature

load1n9:


aapoalas (approved):
Thank you for this and the others @load1n9 <3 This is great stuff to have

PR #441, opened 2024-10-13
feat(string): add normalize method

ivandevp:
Hey!

This is my attempt to implemente the `String.prototype.normalize` method - any feedback is really appreciated.


nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -567,8 +568,42 @@ impl StringPrototype {
...
+        // 2. Let S be ? ToString(O).
+        let s = to_string(agent, o)?;
+
+        // 3. If form is undefined, let f be "NFC".
+        let form = arguments.get(0);
+        let f = if form.is_undefined() {
+            NormalizeForm::NFC
+        } else {
+            // 4. Else, let f be ? ToString(form).
+            let form_result = NormalizeForm::from_str(form.to_string(agent).unwrap().as_str(agent));
```
aapoalas, line 589: **issue:** The `unwrap()` here is not according to spec: Spec says `? ToString(form)` so we should do `to_string(agent, form)?`.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -1478,3 +1513,44 @@ enum TrimWhere {
...
+            "NFD" => Ok(NormalizeForm::NFD),
+            "NFKC" => Ok(NormalizeForm::NFKC),
+            "NFKD" => Ok(NormalizeForm::NFKD),
+            _ => Err(()),
+        }
+    }
+}
+
+fn unicode_normalize(s: &str, f: NormalizeForm) -> std::string::String {
+    match f {
```
aapoalas, line 1550: **suggestion:** Let's use the unicode-normalization crate's `is_nfX_quick` functions to check (without allocations) if `s` is already normalized, and only if that returns `No` or `Maybe` do the `s.nfX().collect()` work. That way we can avoid allocating in the (probably?) common case of "this string is already normalized".

I don't know if ASCII is guaranteed to be already normalized, but if it is then we have another data point for heap-allocated strings that we can use to check the need to normalize, which is the UTF-8 => UTF-16 mapping table that heap strings carry.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -1478,3 +1513,44 @@ enum TrimWhere {
     End,
     StartAndEnd,
 }
+
+enum NormalizeForm {
```
aapoalas, line 1517: **praise:** This is an excellent idea, nice work!

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -1478,3 +1513,44 @@ enum TrimWhere {
...
+    NFC,
+    NFD,
+    NFKC,
+    NFKD,
+}
+
+impl NormalizeForm {
+    fn as_str(&self) -> &'static str {
```
aapoalas, line 1525: **nitpick:** Is this function actually ever used?

nova_vm/Cargo.toml
```diff
@@ -20,6 +20,7 @@ rand = { workspace = true }
 ryu-js = { workspace = true }
 small_string = { path = "../small_string" }
 sonic-rs = { workspace = true, optional = true }
+unicode-normalization = "0.1.24"
```
aapoalas, line 23: **issue:** We've been keeping all our dependencies at the top level, so please add this dependency to the main `Cargo.toml` and then make the dependency `{ workspace = true }` here.

aapoalas (changes requested):
Hello and welcome! First off, great work and thank you very much: Aside from one issue with unwrapping when the spec says to rethrow an error, and the dependency definition location,  this looks fully mergeable as-is which is absolutely great!

You'll need to update the test262 expectations & metrics: The magic spells for doing this are mentioned in the contributing doc under https://github.com/trynova/nova/blob/main/CONTRIBUTING.md#tests-in-prs

Again, great work, the code is really pretty and I have no real significant complaints. Stellar work! <3

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -1478,3 +1513,44 @@ enum TrimWhere {
     End,
     StartAndEnd,
 }
+
+enum NormalizeForm {
+    NFC,
```
aapoalas, line 1518: **issue:** Apparently clippy doesn't like these names and wants these to be `Nfc`, `Nfd` etc. Let's give unto clippy what is theirs and do it that way.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
ivandevp: Oh, yeah. That's right! Thanks for catching this!

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
ivandevp: Goot catch, I was using it in a previous implementation and forgot to remove it after refactoring it :)

nova_vm/Cargo.toml
ivandevp: Got it, it's now moved at the top level.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
ivandevp: That's right. Just renamed them ;)

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
aapoalas: **issue:** You're now correctly checking for already-normalized strings but you're still making a clone of the string data. We don't want to do that: Please instead return eg. `None` from this function to signify that the original `s` parameter in the `normalize` function already contains a normalized function.

That way we actually avoid making a clone of the string data entirely.

aapoalas (changes requested):
Still has unnecessary heap data clone.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
ivandevp: ohhh... got it, yes... that makes sense! I refactored the helper function, I think we could avoid the extra clone now

aapoalas (approved):
Looks excellent to me now! Thank you again, welcome, and great work! <3

PR #442, opened 2024-10-13
feat: WeakMap, WeakSet, WeakRef behind the `weak-refs`  feature

load1n9:


aapoalas (commented):
**suggestion::** How about combining WeakMap, WeakSet, WeakRef, and FinalizationRegistry under a combined "weak-refs" feature? Disabling that feature would mean that the engine does not have any ephemerons and thus garbage collection can be done with a very simple algorithm (the one we use right now; I haven't actually implemented ephemeron collection yet).

load1n9:
sounds good!

aapoalas (approved):
LGTM, thank you <3

PR #444, opened 2024-10-15
feat: atomics behind atomics feature to complete atomic things 

load1n9:


eliassjogreen:
**Question**: This should maybe have a dependency on SABs and or ABs being enabled or even the same flag as them?

load1n9:
> **Question**: This should maybe have a dependency on SABs and or ABs being enabled or even the same flag as them?

made it depend on both of those features

aapoalas (approved):
LGTM, thank you again <3

PR #445, opened 2024-10-15
feat(array): add reverse prototype method

ivandevp:
Hey!

This PR adds `Array.prototype.reverse` method implementation. 

There are still 3 tests failing but as far as I could check they're failing because some built-in constructors are still not implemented (e.g. `TypedArrays`), a bit of guidance on how to implement them would be highly apprecciated :)


nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -2088,8 +2088,68 @@ impl ArrayPrototype {
         Ok(accumulator)
     }
 
-    fn reverse(_agent: &mut Agent, _this_value: Value, _: ArgumentsList) -> JsResult<Value> {
-        todo!()
+    fn reverse(agent: &mut Agent, this_value: Value, _: ArgumentsList) -> JsResult<Value> {
+        // 1. Let O be ? ToObject(this value).
+        let o = to_object(agent, this_value)?;
```
aapoalas, line 2093: **suggestion:** Here it would be a good idea to add a fast path for "trivial" Arrays, see `is_trivial` check usage in this file. I haven't checked but I assume Rust has a reverse-in-place kind of function for mut slices, which we can safely use with trivial Arrays.

aapoalas (approved):

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
ivandevp: Hey @aapoalas, thanks for your suggestion. I just added the fast path :)

ivandevp:
> **praise:** Oh that's a beautiful fast path, absolutely splendid! Great work, and thank you! This looks good to merge from my point of view!

Thanks for the fast follow up on this PR @aapoalas  ❤️  - Just curious if we need to address the failing tests on this PR:

```
  "built-ins/Array/prototype/reverse/get_if_present_with_delete.js": "CRASH",
  "built-ins/Array/prototype/reverse/length-exceeding-integer-limit-with-proxy.js": "CRASH",
  "built-ins/Array/prototype/reverse/resizable-buffer.js": "CRASH",
```

If so, do you have any advice on how to fix them?

aapoalas:
> > **praise:** Oh that's a beautiful fast path, absolutely splendid! Great work, and thank you! This looks good to merge from my point of view!
> 
> Thanks for the fast follow up on this PR @aapoalas ❤️ - Just curious if we need to address the failing tests on this PR:
> 
> ```
>   "built-ins/Array/prototype/reverse/get_if_present_with_delete.js": "CRASH",
>   "built-ins/Array/prototype/reverse/length-exceeding-integer-limit-with-proxy.js": "CRASH",
>   "built-ins/Array/prototype/reverse/resizable-buffer.js": "CRASH",
> ```
> 
> If so, do you have any advice on how to fix them?

Ah yeah, sorry I'm traveling and your question kind of slipped my mind :)

No need to address them: They're failing for an entirely acceptable reason, at least the resizable buffer and proxy case. If the `get_if_present_with_delete.js` also is failing from eg. Proxy not being implemented then it's absolutely fine as is.

ivandevp:
Great - thanks for taking the time to check it @aapoalas  - then yes, I think we can merge it as is for now 🚀

aapoalas:
Thank you again <3 Excellent code!

PR #446, opened 2024-10-16
feat(array): add to_reversed prototype method

ivandevp:
Hey!

This PR adds `Array.prototype.to_reversed` method.

I tried to add a fast path similar to `Array.prototype.reverse` method but I wasn't able to figure out how to clone the array in order to not mutate the original array order. At least by doing `array.clone()` didn't work (it mutated the original array as well) and then the converting it to slice and back to an array triggers some issues between `&[Option<Value>]` and `&[Value]` that I couldn't find an easy fix - if we should add the fast path anyway, ideas are welcomed on how to implement it :pray:


nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -2624,8 +2624,31 @@ impl ArrayPrototype {
...
+            let from = PropertyKey::Integer((len - k - 1).try_into().unwrap());
+            //    b. Let Pk be ! ToString(𝔽(k)).
+            let pk = PropertyKey::try_from(k).unwrap();
+            //    c. Let fromValue be ? Get(O, from).
+            let from_value = get(agent, o, from)?;
+            //    d. Perform ! CreateDataPropertyOrThrow(A, Pk, fromValue).
+            create_data_property_or_throw(agent, a, pk, from_value)?;
+            //    e. Set k to k + 1.
+            k += 1;
+            eprintln!("k: {}", k);
```
aapoalas, line 2648: **nitpick (blocking):** Leftover debug printing.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -2624,8 +2624,31 @@ impl ArrayPrototype {
...
+        // 5. Repeat, while k < len,
+        while k < len {
+            //    a. Let from be ! ToString(𝔽(len - k - 1)).
+            let from = PropertyKey::Integer((len - k - 1).try_into().unwrap());
+            //    b. Let Pk be ! ToString(𝔽(k)).
+            let pk = PropertyKey::try_from(k).unwrap();
+            //    c. Let fromValue be ? Get(O, from).
+            let from_value = get(agent, o, from)?;
+            //    d. Perform ! CreateDataPropertyOrThrow(A, Pk, fromValue).
+            create_data_property_or_throw(agent, a, pk, from_value)?;
```
aapoalas, line 2645: **issue:** Here the spec says `! CreateDataPropertyOrThrow`: This means that the spec knows this cannot throw, ie.. this is `unwrap()` (or even `unwrap_unchecked()` if we wanted to be dangerous about this, which we don't for now).

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -2624,8 +2624,31 @@ impl ArrayPrototype {
         todo!();
     }
 
-    fn to_reversed(_agent: &mut Agent, _this_value: Value, _: ArgumentsList) -> JsResult<Value> {
-        todo!();
+    fn to_reversed(agent: &mut Agent, this_value: Value, _: ArgumentsList) -> JsResult<Value> {
+        // 1. Let O be ? ToObject(this value).
+        let o = to_object(agent, this_value)?;
```
aapoalas, line 2629: > I tried to add a fast path similar to Array.prototype.reverse method but I wasn't able to figure out how to clone the array in order to not mutate the original array order. At least by doing array.clone() didn't work (it mutated the original array as well) and then the converting it to slice and back to an array triggers some issues between &[Option<Value>] and &[Value] that I couldn't find an easy fix - if we should add the fast path anyway, ideas are welcomed on how to implement it 🙏

As you noticed, we do not have an existing cloning algorithm for Arrays or their elements data. `array.clone()` just clones the Array reference, so this is the same as doing this in JS:

```ts
let array = [...];
let array2 = array; // array.clone();
```

In order to actually implement cloning, I spent a good while writing complicated and interesting ideas. Turns out Rust already has exactly what we need. If the Array is dense and trivial (contains no holes and no getters) then we can just use `Vec::extend_from_within(index..index+1)` to push an internal copy to the end of the Vec.

Where we need to use this the `ElementArrays` in `element_array.rs`: We can extend_from_within the correct `ElementArray2PowX` Vector with the index of our `from` Array's `elements: SealableElementsVector`. Now the last index in `ElementArray2PowX` is a copy of our `from` Array's elements data, and we can take that index as our clone'd Array's `elements: SealableElementsVector`. We also need to manually create an `ArrayHeapData` with that elements etc. but that's just side rumble at this point.

Once the cloning function is done, the last stop is to just do the `array.as_mut_slice(agent).reverse()` on the cloned Array.

Does this sound doable? Do you want to give it a go?

aapoalas (changes requested):
Great job! Few minor issues, but all in all this is perfectly good to go.

Adding a fast path is possible, but requires a tad bit more involved work that touches on the Array's backing "elements heap vectors". This isn't too-too bad but requires a bit of work. I'll be happy to help if you want to take a stab at it.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
ivandevp: Sure! I'll give it a try - I'm still not 100% clear on the implementation but let me revisit this along with the code and if I get questions I'll let you know :)

ivandevp: I tried following your suggestion on Discord and committed my attempt for implementation 😃

nova_vm/src/heap/element_array.rs
```diff
@@ -2304,6 +2304,81 @@ impl ElementArrays {
...
+            e2pow12,
+            e2pow16,
+            e2pow24,
+            e2pow32,
+        } = self;
+        let new_index = match new_key {
+            ElementArrayKey::Empty => ElementIndex::from_u32_index(0),
+            ElementArrayKey::E4 => {
+                let elements = e2pow4;
+                let index = ElementIndex::last_element_index(&elements.values).into_index();
```
aapoalas, line 2327: **issue:** This isn't correct; you're now always making a shallow copy of the last allocated elements in this particular `e2powX` array. What you want to do is to use the index from the `elements_vector` you took as a parameter, so that you make a shallow copy of that instead.

aapoalas (changes requested):
Some errors leaked in but overall this is indeed exactly how the fast path should work. Nice work! <3

nova_vm/src/ecmascript/builtins/array.rs
```diff
@@ -80,6 +80,19 @@ impl Array {
         agent[self].elements.is_trivial(agent)
     }
 
+    #[inline]
+    pub(crate) fn shallow_clone(self, agent: &mut Agent) -> Result<Array, ()> {
+        let elements = agent[self].elements;
+        let object_index = self.get_backing_object(agent);
+        let cloned_elements = agent.heap.elements.shallow_clone(elements.into());
+        let data = ArrayHeapData {
+            object_index,
```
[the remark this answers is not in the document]
ivandevp: Got it. Thanks for the clear explanation on this 🙏

nova_vm/src/heap/element_array.rs
```diff
@@ -2304,6 +2304,81 @@ impl ElementArrays {
                 .contains(&Some(element)),
         }
     }
+
+    /// This method creates a "shallow clone" of the elements of a trivial/dense array.
+    /// It does not do anything with descriptors and assumes there is a previous validation in place.
+    pub fn shallow_clone(&mut self, elements_vector: ElementsVector) -> SealableElementsVector {
+        let new_len = usize::try_from(elements_vector.len()).unwrap();
+        let new_key = ElementArrayKey::from(new_len);
```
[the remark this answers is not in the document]
ivandevp: I think I got the idea, I refactored a bit the method, tho I think there's no need to create both methods when they do pretty much the same but using a different input (key / length). I'm not happy with the current name I used for the inner method but let me know if you have any better suggestions, please :)

nova_vm/src/heap/element_array.rs
```diff
@@ -2304,6 +2373,18 @@ impl ElementArrays {
...
+        if new_key == elements_vector.cap {
```
aapoalas, line 2382: **issue:** As currently written, the comparison is the wrong way around: If the new key is the same as the old one, then we can use the new key. If it's different, then we must use the old key. **Note** that this ends up as just being "use the old key always".

Using the old key always is absolutely a valid choice. The downside is as I mentioned that it will overallocate when the source Array has a higher capacity than it would need nowadays. This is the `Array(256)` case I mentioned earlier.

Note that fixing that downside requires using an API different from `extend_from_within`; it would be an `extend_from_slice` from the source key+index to the `new_key` + new index (push + last_element_index).

But that's probably going to be too hard; you can just remove this if clause, and remove the `new_len`/`new_key` calculation. Just use the old key and old len directly.

ivandevp: Oh, I see. Alright, I removed the conditional checks and hope at some point I can add support to that use case you mentioned.

nova_vm/src/ecmascript/builtins/array.rs
```diff
@@ -80,6 +80,17 @@ impl Array {
         agent[self].elements.is_trivial(agent)
     }
 
+    pub(crate) fn shallow_clone(self, agent: &mut Agent) -> Array {
```
aapoalas, line 83: **suggestion:** This could maybe be renamed to "clone_array" (a nod to us only taking the array features) or "to_cloned" (a nod to the prototype methods that use this).

aapoalas, line 83: **needs:** Please add a comment to mention that the array should be simple (no descriptors), and that otherwise this will do somewhat odd things (getter/setter indexes become holes). I know you have that comment on the elements side, but we should also have it here since this is the more likely "entry point" to call the function.

nova_vm/src/heap/element_array.rs
```diff
@@ -2304,6 +2304,72 @@ impl ElementArrays {
                 .contains(&Some(element)),
         }
     }
+
+    /// This method creates a "shallow clone" of the elements of a trivial/dense array.
+    /// It does not do anything with descriptors and assumes there is a previous validation in place.
+    pub fn shallow_clone(&mut self, elements_vector: ElementsVector) -> SealableElementsVector {
```
aapoalas, line 2310: **note:** I think this function name is probably good as it is; my suggested renaming of the Array method does not transitively apply to this method :)

aapoalas (approved):
LGTM, and thank you for the work and for the patience! Great stuff, and the shallow clone helper function will be of great help in the future with other `toCloned` array prototype methods.

I left a few minor things that could be changed, the comment thing being most important. But otherwise this looks very much good to go! <3

nova_vm/src/ecmascript/builtins/array.rs
ivandevp: I opted for `to_cloned` in order to make it more readable when calling `array.to_cloned()`

aapoalas:
Looks all good to me! Thank you very much, and sorry for the delay in reviewing! <3

aapoalas (approved):

PR #455, opened 2024-11-03
feat(ecmascript): Implement dataview `get` and `set` methods

### LINKED ISSUE #152: DataView builtin functions

eliassjogreen:
I implemented the `get*` and `set*` methods but it may still not be entirely correct due to the `get_offset_by_byte` and `set_offset_by_byte` methods I added to `DataBlock`. I also added a few methods to the `Viewable` which I am somewhat unsure about if it's the correct approach.

Closes #152


nova_vm/src/ecmascript/builtins/array_buffer/abstract_operations.rs
```diff
@@ -366,39 +376,49 @@ pub(crate) fn get_raw_bytes_from_shared_block(
...
     // 4. Let elementSize be the Element Size value specified in Table 71 for Element Type type.
+    // 3. Let block be arrayBuffer.[[ArrayBufferData]].
+    let block = agent[array_buffer].get_data_block();
     // 5. If IsSharedArrayBuffer(arrayBuffer) is true, then
     // a. Assert: block is a Shared Data Block.
     // b. Let rawValue be GetRawBytesFromSharedBlock(block, byteIndex, type, isTypedArray, order).
     // 6. Else,
     // a. Let rawValue be a List whose elements are bytes from block at indices in the interval from byteIndex (inclusive) to byteIndex + elementSize (exclusive).
     // 7. Assert: The number of elements in rawValue is elementSize.
+    // TODO: Add [[LittleEndian]] to the agent record.
```
aapoalas, line 399: **note:** The LE status of the Agent record depends on just the machine's natural LE status, so this should instead be a build config thing basically. On BE build targets it'll be BE but otherwise LE.

nova_vm/src/ecmascript/builtins/array_buffer/abstract_operations.rs
```diff
@@ -366,39 +376,49 @@ pub(crate) fn get_raw_bytes_from_shared_block(
...
 /// isLittleEndian (a Boolean) and returns a List of byte values.
-pub(crate) fn numeric_to_raw_bytes(
-    _array_buffer: ArrayBuffer,
-    _type: (),
-    _value: Number,
-    _is_little_endian: bool,
-) {
+pub(crate) fn numeric_to_raw_bytes<T: Viewable>(
+    agent: &mut Agent,
+    gc: GcScope<'_, '_>,
```
aapoalas, line 418: **issue:** This cannot trigger GC and thus shouldn't take the GC scope.

nova_vm/src/ecmascript/builtins/data_view/abstract_operations.rs
```diff
@@ -143,3 +154,151 @@ pub(crate) fn is_view_out_of_bounds(
...
+    view: Value,
+    request_index: Value,
+    is_little_endian: Value,
+) -> JsResult<Value> {
+    // 1. Perform ? RequireInternalSlot(view, [[DataView]]).
+    // 2. Assert: view has a [[ViewedArrayBuffer]] internal slot.
+    let view = require_internal_slot_data_view(agent, view)?;
+
+    // 3. Let getIndex be ? ToIndex(requestIndex).
+    let get_index = to_index(agent, gc, request_index)? as usize;
```
aapoalas, line 178: **suggestion:** Do an fast-path for if index and LE being integer and Boolean (or undefined) respectively. We can avoid the calls and having to pass GC token (this will become very relevant in the future.)

nova_vm/src/ecmascript/builtins/array_buffer/abstract_operations.rs
```diff
@@ -422,29 +447,40 @@ pub(crate) fn numeric_to_raw_bytes(
...
 /// optional argument isLittleEndian (a Boolean) and returns UNUSED.
-pub(crate) fn set_value_in_buffer(
-    _array_buffer: ArrayBuffer,
-    _byte_index: u32,
-    _type: (),
-    _value: Value,
+#[allow(clippy::too_many_arguments)]
+pub(crate) fn set_value_in_buffer<T: Viewable>(
+    agent: &mut Agent,
+    gc: GcScope<'_, '_>,
```
aapoalas, line 453: **note:** Same thing here, this cannot trigger GC.

nova_vm/src/ecmascript/builtins/structured_data/data_view_objects/data_view_prototype.rs
```diff
@@ -248,184 +249,364 @@ impl DataViewPrototype {
...
-        _: ArgumentsList,
+        agent: &mut Agent,
+        gc: GcScope<'_, '_>,
+        this_value: Value,
+        arguments: ArgumentsList,
     ) -> JsResult<Value> {
-        todo!()
+        let byte_offset = arguments.get(0);
+        // 2. If littleEndian is not present, set littleEndian to false.
+        let little_endian = if arguments.len() > 1 {
```
aapoalas, line 289: **thought:** Because `to_boolean` cannot throw any errors, we could perform that transform here and call `get_view_value` with an actual bool.

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -61,18 +67,248 @@ mod private {
     impl Sealed for f64 {}
 }
 
-pub trait Viewable: private::Sealed {}
+pub trait Viewable: private::Sealed + Copy {
+    fn is_bigint_type() -> bool;
```
aapoalas, line 71: **nitpick:** this could be an associated const.

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -61,18 +67,248 @@ mod private {
...
+    fn into_be_value(self, agent: &mut Agent) -> Value;
+    fn into_le_value(self, agent: &mut Agent) -> Value;
+    fn from_le_value(agent: &mut Agent, gc: GcScope<'_, '_>, value: Value) -> Self;
+    fn from_be_value(agent: &mut Agent, gc: GcScope<'_, '_>, value: Value) -> Self;
+}
```
aapoalas, line 76: **note:** to and from ne (natural alignment of the machine) might be of use as well.

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -61,18 +67,248 @@ mod private {
...
+
+impl Viewable for u8 {
+    fn is_bigint_type() -> bool {
+        false
+    }
+
+    fn into_be_value(self, _: &mut Agent) -> Value {
+        Value::from(self.to_be())
```
aapoalas, line 84: **nitpick:** No need to call these no-ops.

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -61,18 +67,248 @@ mod private {
...
+    }
+
+    fn into_le_value(self, _: &mut Agent) -> Value {
+        Value::from(self.to_le())
+    }
+
+    fn from_be_value(agent: &mut Agent, gc: GcScope<'_, '_>, value: Value) -> Self {
+        value.to_real(agent, gc).unwrap() as Self
```
aapoalas, line 92: **suggestion:** These methods definitely call for fast-paths for integers.

Eventually we'll probably want separate methods for a "try from Value" variants that can't trigger GC.

aapoalas (changes requested):
Had a few minor issues and some suggestions, but in general looks pretty damn good to me! Good job and thank you <3

nova_vm/src/ecmascript/builtins/array_buffer/abstract_operations.rs
eliassjogreen: Required because the `to_*` conversion functions called by `from_le_value` and `from_be_value` require the GC token.

nova_vm/src/ecmascript/types/spec/data_block.rs
eliassjogreen: I just realized these should follow the conversion table of typed arrays:
https://tc39.es/ecma262/#table-the-typedarray-constructors

But yes, making these (or even the `to_*` functions) not require GC would be a very nice thing.

nova_vm/src/ecmascript/builtins/array_buffer/abstract_operations.rs
eliassjogreen: The `from_le_value` and `from_be_value` require the GC scope due to the `to_*` functions they call internally.

nova_vm/src/ecmascript/builtins/data_view/abstract_operations.rs
```diff
@@ -143,3 +154,147 @@ pub(crate) fn is_view_out_of_bounds(
...
+) -> JsResult<Value> {
+    // 1. Perform ? RequireInternalSlot(view, [[DataView]]).
+    // 2. Assert: view has a [[ViewedArrayBuffer]] internal slot.
+    let view = require_internal_slot_data_view(agent, view)?;
+
+    // 3. Let getIndex be ? ToIndex(requestIndex).
+    let get_index = to_index(agent, gc.reborrow(), request_index)? as usize;
+
+    // 4. If IsBigIntElementType(type) is true, let numberValue be ? ToBigInt(value).
+    let number_value = if T::IS_BIGINT {
```
aapoalas, line 248: **suggestion:** I think if we add a `T::NumberValueType` which is `Number` for non-64-bit viewables and `BigInt` for the rest, then we can here do `let number_value: T::NumberValueType = if T::IS_BIGINT`, and then the `from_xe_value` variants won't need `gc` anymore as they're always converting from `Number` to  native type or from `BigInt` to native type.

nova_vm/src/ecmascript/builtins/structured_data/data_view_objects/data_view_prototype.rs
aapoalas, line 1: **praise:** This file is a joy on the eyes <3 Great work with this <3

nova_vm/src/ecmascript/builtins/data_view/abstract_operations.rs
aapoalas: Yeah, `to_boolean` can be skipped unconditionally, I forgot that it cannot trigger GC anyhow.

We can also skip the "fast path" for `to_index`. Eventually we'll want to employ those... Or maybe I'll change the APIs for `to_index` and similar kinds of functions now that I think about it... The thing is that if we know statically that we're not calling any JavaScript code, then we don't need to prepare for GC which means we don't won't need to perform any rooting of Values using `Local<Value>`. But actually indeed it's probably better to eventually turn `to_index` into something like this:

```rs
fn to_index(agent: &mut Agent, index: Value, on_needs_gc: FnMut(&mut Agent, Object) -> JsResult<Number>) -> JsResult<Number> { ... }
```

... well, not this exactly. But something where effectively the caller provides a callback that the method can use to signal that it is about to call into JavaScript and the caller should store its Values into Locals.

nova_vm/src/ecmascript/types/language/bigint.rs
```diff
@@ -502,7 +520,6 @@ impl BigInt {
         }
     }
 
-    /// ### [
```
aapoalas, line 505: **praise:** Nice catch!

aapoalas (approved):
Looks good to me, though I had a suggestion through which we can still avoid passing GC into the `set_view_value` method.

nova_vm/src/ecmascript/builtins/data_view/abstract_operations.rs
eliassjogreen: I'll leave that for a future improvement of the `Viewable` trait. We'll also need a `IS_CLAMPED` trait const or something for when we implement clamped arrays.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -707,6 +707,69 @@ pub(crate) fn string_to_big_int(_agent: &mut Agent, _argument: String) -> Option
...
+
+    argument: Value,
+) -> JsResult<i64> {
+    // 1. Let n be ? ToBigInt(argument).
+    let n = to_big_int(agent, gc, argument)?;
+
+    // 2. Let int64bit be ℝ(n) modulo 2**64.
+    match n {
+        BigInt::BigInt(heap_big_int) => {
+            let int64bit = agent[heap_big_int].data.clone() % u64::MAX;
```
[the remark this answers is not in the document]
eliassjogreen: Correct! Managed to wrangle the bigints to i64 and u64 respectively. Also tested it against the test262 tests and it seems to function correctly.

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -707,6 +707,69 @@ pub(crate) fn string_to_big_int(_agent: &mut Agent, _argument: String) -> Option
...
+///
+/// The abstract operation ToBigInt64 takes argument argument (an ECMAScript
+/// language value) and returns either a normal completion containing a BigInt
+/// or a throw completion. It converts argument to one of 2**64 BigInt values
+/// in the inclusive interval from ℤ(-2**63) to ℤ(2**63 - 1).
+#[inline(always)]
+pub(crate) fn to_big_int64(
+    agent: &mut Agent,
+    gc: GcScope<'_, '_>,
+
```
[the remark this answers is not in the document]
eliassjogreen: Ah, lots (~700) of matches for those empty lines after the gc parameter, might be nice to do a formatting and cleaning up PR in the future for those.

eliassjogreen:
@aapoalas Ready to merge when you are :)

nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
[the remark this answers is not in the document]
aapoalas: Yeah. I thought and trusted that cargo fmt would do it for me automatically but seemingly it didn't. Hence why they got left in.

PR #456, opened 2024-11-04
feat(ecmascript): Implement `parseInt`

eliassjogreen:


nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -725,15 +728,119 @@ impl GlobalObject {
...
+            if radix == 10 && string.is_number() {
+                return Ok(string);
+            }
+        }
+
+        // 1. Let inputString be ? ToString(string).
+        let s = to_string(agent, gc.reborrow(), string)?;
+        let s = s.as_str(agent).to_owned();
+
+        // 2. Let S be ! TrimString(inputString, start).
```
aapoalas, line 772: **suggestion:** Because steps 2, 3, 4, 5 cannot throw they cannot be observed. As such, we can move them freely. If we move them to happen after step 6, we can also move `let s = s.as_str(agent)` to happen after `r = to_int32(agent, gc, radix)?`. That then means that we do not need the `to_owned` heap copy anymore and can operate entirely on a borrowed string.

nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -725,15 +728,119 @@ impl GlobalObject {
...
+        // 14. Let mathInt be the integer value that is represented by Z in
+        //     radix-R notation, using the letters A through Z and a through z
+        //     for digits with values 10 through 35. (However, if R = 10 and Z
+        //     contains more than 20 significant digits, every significant
+        //     digit after the 20th may be replaced by a 0 digit, at the option
+        //     of the implementation; and if R is not one of 2, 4, 8, 10, 16,
+        //     or 32, then mathInt may be an implementation-approximated
+        //     integer representing the integer value denoted by Z in radix-R
+        //     notation.)
+        let math_int = i128::from_str_radix(z, r).unwrap() as f64;
```
aapoalas, line 835: **suggestion:** We should try to avoid the f64 conversion if at all we can. Some `try_from` shenanigans would probably be a good idea here to try and avoid that. Also, for `r = 10` and `r = 2` various optimisations based on the length of `z` may be beneficial.

eg. `r = 10` if `z` length is less than 16 then we're guaranteed a safe integer value. If less than 3 then it's a guaranteed u8. If less than 5 then u16. So on. Same sort of logic can be used for `r = 2`.

aapoalas (approved):
LGTM! I had some suggestions.

nova_vm/src/ecmascript/builtins/global_object.rs
eliassjogreen: Implemented these fun fast-paths for most known safe integer values, and a macro for easily adding more if the need arises for any common radixes.

nova_vm/src/ecmascript/builtins/global_object.rs
eliassjogreen: Fixed!

eliassjogreen:
Ready to merge :)

I did a bunch of micro benchmarks on the different combinations of matchers and found that they were all within a very small margin but found that the one which can be found pushed was the fastest by 0.33%:
```
  ./target/release/nova_cli_fast eval scripts/parseInt/fast.js ran
    1.02 ± 0.33 times faster than ./target/release/nova_cli_fast_slimmer eval scripts/parseInt/fast.js
    1.02 ± 0.33 times faster than ./target/release/nova_cli_slow eval scripts/parseInt/fast.js
    1.02 ± 0.32 times faster than ./target/release/nova_cli_fast_slim_unwrap_unsafe eval scripts/parseInt/fast.js
    1.05 ± 0.32 times faster than ./target/release/nova_cli_fast_unwrap_unsafe eval scripts/parseInt/fast.js
    1.05 ± 0.34 times faster than ./target/release/nova_cli_fast_slim eval scripts/parseInt/fast.js
```
Might be statistically insignificant and might even compile down to the same machine code but was pretty consistently a tiiiiny bit better so thats what we have.

nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -725,15 +728,180 @@ impl GlobalObject {
...
+            || string.is_empty_string()
+        {
+            return Ok(Value::nan());
+        }
+
+        // OPTIMIZATION: If the string is a number and the radix is 10, return the number.
+        if let Value::Integer(radix) = radix {
+            let radix = radix.into_i64();
+            if radix == 10 && string.is_number() {
+                return Ok(string);
```
aapoalas, line 764: **issue (nonblocking): I forgot that I had tried this, but this fast path is wrong for non-integers. I'm not 100% sure on this but probably this is equal to a `trunc` call. Eg. `parseInt(Math.PI) === 3` IIRC.

PR #457, opened 2024-11-04
feat(heap): Implement a simple string interner

eliassjogreen:
A quite simple implementation of a string interner as a `AHashMap<StringHeapData, StringIndex>` lookup table. Seems to be broken and breaking tests, probably for a good reason. Ideally I would like the lookup table to not take `StringHeapData` as the key and instead take something like `&'a str` but that would require a lifetime for the heap. Far from mergeable as is.

> [!WARNING]
> Needs to be implemented using a `HashTable<...>` instead. See `Map`/`Set` for reference.

Greatly speeds up operations where there are lots of strings.


eliassjogreen:
Got stuck trying to implement it using `HashTable`. May re-open in the future if.

PR #459, opened 2024-11-10
feat(ecmascript): Implement `parseFloat`

eliassjogreen:
Uses [fast_float](https://docs.rs/fast-float/latest/fast_float/).


nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -719,14 +719,72 @@ impl GlobalObject {
...
+        }
+
+        if trimmed_string.starts_with("-Infinity") {
+            return Ok(Value::neg_inf());
+        }
+
+        // This check is used to prevent fast_float from parsing "inf", "+inf" as Infinity and "-inf" as -Infinity.
+        // Copied from boa. Source https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/number/globals.rs#L314-L319
+        // Copyright (c) 2019 Jason Williams
+        let trimmed_string_prefix = trimmed_string
```
aapoalas, line 764: **issue:** I think we should to avoid the string heap copy here. We could either copy the first four bytes onto the stack (and do the lowercasing manually), or perhaps just manually check the first four bytes in a helper function.

It might also be worth it to only do this after we parse float and if we get an infinite result.

eliassjogreen: Done!

nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -719,14 +719,65 @@ impl GlobalObject {
...
+            return Ok(Value::neg_inf());
+        }
+
+        if let Ok((f, len)) = fast_float::parse_partial::<f64, _>(trimmed_string) {
+            if len == 0 {
+                return Ok(Value::nan());
+            }
+
+            // NOTE: This check is used to prevent fast_float from parsing "inf", "+inf" as Infinity and "-inf" as -Infinity.
+            if f.is_infinite() {
```
aapoalas, line 767: **praise:** Oh nice! The `eq_ignore_ascii_case` API is a steal of a find here! Great stuff!!!

aapoalas:
Damn, seems like a strict equality comparison isn't enough as "infinity" also must not parse into Infinity.

eliassjogreen:
Dang, pushed a fix.

aapoalas:
Thank you very much! Awesome work! <3

PR #461, opened 2024-11-14
feat(ecmascript): Array.prototype.unshift

CarrotzRule123:
This PR implements `Array.prototype.unshift`


load1n9 (changes requested):
run `cargo fmt` but other than that lgtm!

tests/expectations.json
```diff
@@ -457,22 +457,12 @@
...
   "built-ins/Array/prototype/unshift/S15.4.4.13_A2_T1.js": "CRASH",
   "built-ins/Array/prototype/unshift/S15.4.4.13_A2_T2.js": "CRASH",
-  "built-ins/Array/prototype/unshift/S15.4.4.13_A2_T3.js": "CRASH",
   "built-ins/Array/prototype/unshift/S15.4.4.13_A3_T2.js": "CRASH",
-  "built-ins/Array/prototype/unshift/S15.4.4.13_A4_T1.js": "CRASH",
-  "built-ins/Array/prototype/unshift/S15.4.4.13_A4_T2.js": "CRASH",
-  "built-ins/Array/prototype/unshift/call-with-boolean.js": "CRASH",
-  "built-ins/Array/prototype/unshift/clamps-to-integer-limit.js": "CRASH",
-  "built-ins/Array/prototype/unshift/length-near-integer-limit.js": "CRASH",
+  "built-ins/Array/prototype/unshift/clamps-to-integer-limit.js": "TIMEOUT",
```
eliassjogreen, line 463: Does this pass when running `cargo run --release --bin test262 -- eval-test built-ins/Array/prototype/unshift/clamps-to-integer-limit.js`? In that case we can remove it from the expectations and add it to the `skip.json`

CarrotzRule123: It still gives `TIMEOUT` though

eliassjogreen: Huh, really slow test... It's probably fine to leave then

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3084,14 +3084,85 @@ impl ArrayPrototype {
...
+                    // 2. Perform ? DeletePropertyOrThrow(O, to).
+                    delete_property_or_throw(agent, gc.reborrow(), o, to)?;
+                }
+                // vi. Set k to k - 1.
+                k -= 1;
+            }
+            // d. Let j be +0𝔽.
+            let mut j = 0;
+            // e. For each element E of items, do
+            for e in items.iter() {
```
load1n9, line 3147: ik the instructions say to use an iterator as such but clippy hates this so maybe
do something like `for (j, e) in items.iter().enumerate()`

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3084,14 +3084,85 @@ impl ArrayPrototype {
...
+                // i. Perform ? Set(O, ! ToString(j), E, true).
+                set(agent, gc.reborrow(), o, j.try_into().unwrap(), *e, true)?;
```
load1n9, line 3149: clippy is giving up an error here as well, if we cant fix it `#[allow(clippy::unnecessary_fallible_conversions)]` this should work

load1n9 (changes requested):
other than the issues below  lgtm!

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
aapoalas: Just `into()`

load1n9: yeah that seems to work on my end

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
aapoalas: **issue:** Yeah, just use `.enumerate()` to get the `j` index.

tests/expectations.json
aapoalas: Ah yeah, it's a very valid timeout. The test checks if we create an object like: `{ length: 2 ** 53 }` then does the unshift try every single index in the range `0..length` or does it have some short-cut that it takes, seeing that the object and its prototypes have no indexed properties.

That being said, it's also the sort of stuff that is just _stupid_ and I hate that it exists :D

tests/expectations.json
```diff
@@ -457,22 +457,12 @@
   "built-ins/Array/prototype/toSpliced/unmodified.js": "CRASH",
   "built-ins/Array/prototype/toString/non-callable-join-string-tag.js": "CRASH",
   "built-ins/Array/prototype/unshift/S15.4.4.13_A1_T1.js": "CRASH",
-  "built-ins/Array/prototype/unshift/S15.4.4.13_A1_T2.js": "CRASH",
   "built-ins/Array/prototype/unshift/S15.4.4.13_A2_T1.js": "CRASH",
   "built-ins/Array/prototype/unshift/S15.4.4.13_A2_T2.js": "CRASH",
-  "built-ins/Array/prototype/unshift/S15.4.4.13_A2_T3.js": "CRASH",
   "built-ins/Array/prototype/unshift/S15.4.4.13_A3_T2.js": "CRASH",
```
aapoalas, line 462: **issue (nonblocking):** It seems we're not treating negative lengths in the way that the spec expects. Not sure where the issue lies though. This is the code it's running:

```js
var obj = {};
obj.unshift = Array.prototype.unshift;
obj[0] = "";
obj.length = -4294967295;

var unshift = obj.unshift("x", "y", "z");
```

Could you check what the crash is coming from using `eval-test` and/or a small test file like a `test.js`? And if the issue is in the unshift implementation then of course fix it. If it is not, then possibly make an issue about it.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3084,14 +3084,85 @@ impl ArrayPrototype {
...
-        _this_value: Value,
-        _: ArgumentsList,
+        agent: &mut Agent,
+        mut gc: GcScope<'_, '_>,
+        this_value: Value,
+        items: ArgumentsList,
     ) -> JsResult<Value> {
-        todo!();
+        // 1. Let O be ? ToObject(this value).
+        let o = to_object(agent, this_value)?;
```
aapoalas, line 3105: **suggestion:** It's possible to add a fast-path here if `this_value` is an Array and it is trivial (contains no non-default descriptors, so no getters, setters, or unwritable, unconfigurable, or unenumerable properties) and dense (contains no holes), and its length property is writable. We'll first need to do a `array.reserve(agent, array.len() + items.len())`, then change the length, access the memory as a mutable slice, copy within it to move old items towards the end, and finally copy data from `items` to the beginning of the memory slice.

aapoalas (changes requested):
Hello and welcome to the Nova project! Thank you very much for taking the time to write this function! <3

Clippy had two issues, we'll want to fix those. Aside from those it seems that negative object lengths (maybe?) give the implementation some issues. Some test262 tests that we would expect to pass do not pass. So we'd like to fix that as well before we merge.

Code-wise, there's a fast path you could add if you want. I found no code issue though, looks really good! Great job, and thank you again! Great to have you here <3

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3084,14 +3084,107 @@ impl ArrayPrototype {
...
+        // 4. If argCount > 0, then
+        if arg_count > 0 {
+            // a. If len + argCount > 2**53 - 1, throw a TypeError exception.
+            if (len + arg_count as i64) > SmallInteger::MAX_NUMBER {
+                return Err(agent.throw_exception_with_static_message(
+                    ExceptionType::TypeError,
+                    "Array length overflow",
+                ));
+            }
+            if let Value::Array(array) = this_value {
```
aapoalas, line 3119: **issue:** Sorry I didn't notice this at first, but the fast path should be at the very beginning of the function: Before even doing `o = to_object` etc. What we want to optimise for is something like `[b,c,d].unshift(a)` as that is the common case, and there the `to_object` call is just going to return `this_value` directly anyway so that call isn't useful.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3084,14 +3084,107 @@ impl ArrayPrototype {
...
+                // Fast path: Array is dense and contains no descriptors. No JS
+                // functions can thus be called by unshift.
+                if array.is_trivial(agent) && array.is_dense(agent) {
```
aapoalas, line 3122: **issue:** I think the array's length value needs to be writable as well, and should be checked here.

aapoalas (changes requested):
Minor issue with the placement of the fast path (and checking for length writability) but otherwise looks perfect to me. Thank you! <3

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3101,6 +3101,35 @@ impl ArrayPrototype {
         this_value: Value,
         items: ArgumentsList,
     ) -> JsResult<Value> {
+        // Fast path: Array is dense and contains no descriptors. No JS
+        // functions can thus be called by unshift.
+        if let Value::Array(array) = this_value {
+            if array.is_trivial(agent)
+                && array.is_dense(agent)
+                && agent[array].elements.len_writable
```
aapoalas, line 3109: **issue:** Before going into this branch, we should also check that `len + arg_count <= u32::MAX`. The `some_usize as u32` doesn't saturate (clamp itself to the maximum value) but instead wraps around (ignores highest bits).

Right now, trying to unshift more data into the Array than can be indexed using u32 will cause a panic.

CarrotzRule123: Fixed by https://github.com/trynova/nova/pull/461/commits/ee619df0d1365259f4cd97a444d652c96aa23894

aapoalas: **issue:** Instead of that, do this:
```rs
if let Value::Array(array) = this_value {
  let len = array.len(agent);
  let arg_count = items.len();
  let final_len = u32::try_from(len + arg_count);
  if final_len.is_some() && array.is_trivial(agent) && array.is_dense(agent) && array.length_writable(agent) {
   // ...
```
(The last `length_writable` would be a worthwhile addition to the Array's API :) )

We need to make sure that the final length can be expressed as a `u32`; you only checked that it is within `u53`. And of course, we don't want to throw an error if we're eg. 1 over `u32::MAX`, that wouldn't be correct according to the spec. We'll just take the slow path in that case.

aapoalas (dismissed):
Looks good to me! If you want to fix the nitpicks then please do, otherwise I'll merge this a bit later.

Thank you very much for your time, patience, and work and welcome to the contributors list :)

CarrotzRule123:
Thanks for all the detailed explanations!
Also, thanks for taking the time and patience to review my code!

load1n9 (approved):
LGTM!

PR #467, opened 2024-11-22
feat(ecmascript): Array.prototype.toSorted

CarrotzRule123:
This PR implements `Array.prototype.toSorted`


load1n9 (approved):
LGTM!

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3034,13 +3034,57 @@ impl ArrayPrototype {
...
+        //     that captures comparator and performs the following steps when
+        //     called:
+        //       a. Return ? CompareArrayElements(x, y, comparator).
+        // 6. Let sortedList be ? SortIndexedProperties(O, len, SortCompare, read-through-holes).
+        let sorted_list: Vec<Value> =
+            sort_indexed_properties::<false, false>(agent, gc.reborrow(), o, len, comparator)?;
+        // 7. Let j be 0.
+        let mut j = 0;
+        // 8. Repeat, while j < len,
+        while j < len {
```
aapoalas, line 3073: **suggestion:** You can add a fast path if `a` is `Value::Array` where you take Array slice `as_mut` and copy from the `sorted_list` directly.

aapoalas (approved):
LGTM! One suggestion for a fast path.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
CarrotzRule123: `array_create` always returns an `Array` tho, does that mean I should take the fast path every time?

aapoalas: Yup :)

And this is actually to be expected: The spec uses `!` (`unwrap()` for us) because it knows exactly what path the code will take. It knows that the data property doesn't already exist and that the array being created isn't a Proxy. Because this is `CreateDataProperty` which is equivalent to `Object.defineProperty` it also means that no setters are called. Effectively, no JavaScript code can run here.

Thus, we too can simply take the path of "least resistance" and directly do what the code would end up doing anyway, which is to write data into the Array backing store.

CarrotzRule123: Doesn't that mean everything after 7. is unreachable?

aapoalas: Well, not "unreachable" in the sense that it shouldn't happen but yeah, it's "unreachable" in the sense that we don't need to do those things concretely as written by the spec.

If you want to verify that what we're doing here is correct, the spec path is basically this:
1. Loop over j = 0..len
2. Get j'th Value from sorted Values
3. Call https://tc39.es/ecma262/#sec-createdatapropertyorthrow with that Value
4. Call https://tc39.es/ecma262/#sec-createdataproperty
5. Call https://tc39.es/ecma262/#sec-array-exotic-objects-defineownproperty-p-desc
    - Here we take path 2.h.
6. Call https://tc39.es/ecma262/#sec-ordinarydefineownproperty
    - Here we take path 2.d.i with `[[Writable]]: true, [[Enumerable]]: true, [[Configurable]]: true` and return `true` from 2.e.

And since in Nova we avoid storing `WEC` descriptors when all the values are true, the only data we store in OrdinaryDefineOwnProperty's step 2.d.i is the value itself, and we store that into the Array's backing store.

So, in the end we do:
1. Loop over j = 0..len
2. Get j'th Value from sorted Values
3. Store that Value to Array's backing store.

CarrotzRule123: Alright, thanks for the explanation!

aapoalas (approved):
LGTM, thank you very much. Great stuff <3

PR #470, opened 2024-11-26
feat(ecmascript): Array.prototype.with

CarrotzRule123:
This PR implements `Array.prototype.with`


nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3247,13 +3247,80 @@ impl ArrayPrototype {
...
+        this_value: Value,
+        arguments: ArgumentsList,
     ) -> JsResult<Value> {
-        todo!();
+        // Fast path: Array is dense and contains no descriptors. No JS
+        // functions can thus be called by with.
+        if let Value::Array(array) = this_value {
+            let len = array.len(agent) as i64;
+            let index = arguments.get(0);
+            let relative_index =
```
aapoalas, line 3262: **suggestion:** Let's conditionalise the fast path with the `index` parameter being an Integer as well so we can avoid the `to_integer_or_infinity` call and the possibility of GC it triggers:

```rs
let index = arguments.get(0);
let value = arguments.get(1);
if let (Value::Array(array), Value::Integer(index)) = (this_value, index) {
  // ...
}
```

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3247,13 +3247,80 @@ impl ArrayPrototype {
...
+                to_integer_or_infinity(agent, gc.reborrow(), index)?.into_i64(agent);
+            let actual_index = if relative_index >= 0 {
+                relative_index
+            } else {
+                len + relative_index
+            };
```
aapoalas, line 3268: **issue:** Throw an error on the fast path if the index is out of bounds after normalisation. We don't need to really go back to the slow path just to get the error thrown from there.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3247,13 +3247,80 @@ impl ArrayPrototype {
...
+            if array.is_trivial(agent)
+                && array.is_dense(agent)
+                && !(actual_index >= len || actual_index < 0)
```
aapoalas, line 3271: **nitpick:** Not relevant after adding the throw path but: I'd remove the double-negation here by going `0 <= actual_index && actual_index < len`.

aapoalas (approved):
LGTM! You're really stamping these out, good stuff and thank you so much <3

I'd make the fast path a bit more straightforward by only engaging it when the index value is an integer, and you can just throw the RangeError from there instead of needing to go back to the slow path for that.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
CarrotzRule123: Out of curiosity, is this something that rustc and llvm could optimize?

aapoalas: Yeah, I'm sure the resulting code is equivalent either way. It's just a matter of readability: Double negatives tend to confuse people :)

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -3247,13 +3247,82 @@ impl ArrayPrototype {
...
+                if actual_index >= len || actual_index < 0 {
+                    return Err(agent.throw_exception_with_static_message(
+                        gc.nogc(),
+                        ExceptionType::RangeError,
+                        "invalid or out-of-range index",
+                    ));
+                }
+                // Fast path: Set new value in cloned array.
+                let cloned_array = array.to_cloned(agent);
+                cloned_array.as_mut_slice(agent)[actual_index as usize] = Some(arguments.get(1));
```
aapoalas, line 3279:
```suggestion
                cloned_array.as_mut_slice(agent)[actual_index as usize] = Some(value);
```

PR #471, opened 2024-11-27
feat(ecmascript): `TypedArray.prototype.join`

aapoalas:


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -595,13 +593,276 @@ impl TypedArrayPrototype {
...
+            (
+                to_string(agent, gc.reborrow(), separator)?
+                    .unbind()
+                    .bind(gc.nogc()),
+                true,
+            )
+        };
+        if len == 0 {
+            return Ok(String::EMPTY_STRING.into_value());
+        }
```
eliassjogreen, line 678: **Nitpick:** Couldn't this be moved to before step 4?

**Suggestion:** A fast-path could also be added for `len == 1`, then we wouldn't need to do all of the allocation for joining. But I guess that would be a very uncommon case, so might not be worth it.

tests/expectations.json
```diff
@@ -11019,18 +11003,7 @@
   "built-ins/TypedArray/prototype/with/this-value-invalid.js": "CRASH",
   "built-ins/TypedArray/resizable-buffer-length-tracking-1.js": "CRASH",
   "built-ins/TypedArray/resizable-buffer-length-tracking-2.js": "CRASH",
-  "built-ins/TypedArrayConstructors/BigInt64Array/prototype/BYTES_PER_ELEMENT.js": "CRASH",
   "built-ins/TypedArrayConstructors/BigUint64Array/prototype.js": "CRASH",
```
eliassjogreen, line 11006: **Unrelated:** Why does the `BigUint64Array` prototype test still fail yet all other TypedArray prototype tests don't?

eliassjogreen (changes requested):
LGTM. Just a random thought or two.

eliassjogreen (approved):

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
aapoalas: Nope, there's a throw-path in `? ToString(separator)` :(

eliassjogreen: Damn, right...

tests/expectations.json
aapoalas: Hmm. That's a good question.

aapoalas: Alright: Some failures at least are caused by the error that is thrown if you run `String(new BigUint64Array())`.

Effectively, it's expected because of our `print` function's method.

PR #480, opened 2024-12-07
fix: breaking gc stuff with arraybuffer and regexp disabled

load1n9:


aapoalas (approved):

aapoalas:
Thank you <3

I really gotta set up an addition build for this...

PR #485, opened 2024-12-11
chore: replace cpu flag with fun clippy command

load1n9:


eliassjogreen:
Why??? 😭

aapoalas:
Mm, I don't quite understand what this does. Can you explain a bit?

load1n9:
@eliassjogreen have clippy while we figure out what flags we want

eliassjogreen:
Most likely we don't want any flags, and if we want any flags it's probably the one we have currently. So I say we keep it this way. Any possible build options we wan't to tweak we can do in the cargo profiles. E.g. opt-level, LTO, etc.

MierenManz:
Do we currently provide builds? Or is everything build by the user?

Cuz we nee to remove the native cpu flag when providing builds because older systems may not support the features our build cpu does support

eliassjogreen:
No, we don't yet provide pre-built executables. Maybe if we should start distributing C-bindings and a pre-built dynamically linked libnova? But the nova cli is mainly for debugging purposes so I don't see any reason to distribute that.

PR #489, opened 2024-12-31
feat(ecmascript): String.prototype toLocaleLowerCase and toLocalUpperCase

yossydev:
I implemented the toLocaleLowerCase and toLocaleUpperCase methods, which were previously marked as todo. Since their behavior is essentially the same as the existing toLowerCase and toUpperCase methods, I implemented them based on the existing functionality.

https://github.com/trynova/nova/blob/6e3de25936e3beb5901fd9282e631b07c542cbc3/tests/test262_runner.rs#L37-L38
it seems that the Intl402 functionality hasn’t been implemented yet. Therefore, I believe the current implementation should work fine.

If you notice any issues or areas of concern, I would greatly appreciate your feedback.

ref: https://github.com/trynova/nova/issues/141


aapoalas (approved):
Hello and welcome to the project! Thank you for implementing these two methods <3

I'm not quite 100% sure on if Rust's `to_uppercase` and `to_lowercase` implement the Unicode Default Case Conversion (Rust's documentation talks of "Unicode Derived Core Property") or something other. Either way, it's basically a given that either our `toLowerCase` or `to_locale_lower_case` is doing the wrong thing in some corner cases (like one character in Turkish, IIRC).

Still, this is absolutely good to have and since I don't know what the difference should be and how we'd get it, I'll say this LGTM! Thank you again, and あけましておめでとう!

PR #490, opened 2024-12-31
feat: 2024 → 2025

yossydev:
Happy New Year!

I hope you don’t mind, but I went ahead and updated the “2024” part.

### DIFF
--- a/nova_vm/src/ecmascript/execution/agent.rs
+++ b/nova_vm/src/ecmascript/execution/agent.rs
@@ -5,7 +5,7 @@
 //! ## Notes
 //!
 //! - This is inspired by and/or copied from Kiesel engine:
-//!   Copyright (c) 2023-2024 Linus Groh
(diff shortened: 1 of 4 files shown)


aapoalas:
Hey, thank you for this but I don't think this needs to be or really should be done: We took inspiration/copied code from Kiesel during the end of 2023 and early 2024, but not since then. Hence, the copyright years staying as they are is _probably_ correct..? I'm no lawyer though.

yossydev:
> Hey, thank you for this but I don't think this needs to be or really should be done: We took inspiration/copied code from Kiesel during the end of 2023 and early 2024, but not since then. Hence, the copyright years staying as they are is _probably_ correct..? I'm no lawyer though.

Thank you for your comment!

I thought it would be better to update it since it was marked as “2024,” so I made the change, but it turns out it wasn’t necessary!

I’ll go ahead and close this then!🙏🏻

Issue #142, opened 2024-04-15
RegExp.prototype builtin functions


aapoalas:

aapoalas:
All methods have been implemented.

PR #496, opened 2025-01-03
feat(ecmascript): regexp get flags

yossydev:
I have implemented get flags for the proxy.

I believe more test262 tests will pass once https://github.com/trynova/nova/blob/71f28ae7b10e2d8335f0720be6ca353884994be4/nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_constructor.rs#L47-L55 is addressed, but I plan to handle that in a separate PR!

ref: https://github.com/trynova/nova/issues/142


nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
```diff
@@ -173,19 +175,155 @@ impl RegExpPrototype {
     fn get_dot_all(
         _agent: &mut Agent,
         _this_value: Value,
-        _: ArgumentsList,
+        _args: ArgumentsList,
         _gc: GcScope<'_, '_>,
     ) -> JsResult<Value> {
         todo!()
     }
```
eliassjogreen, line 183: **suggestion:** Add a doc comment `/// ### [22.2.6.4 get RegExp.prototype.flags](https://tc39.es/ecma262/#sec-get-regexp.prototype.flags)` linking to the spec.

yossydev: Thank you for the review!
I’ve added a doc comment to get_flags! https://github.com/trynova/nova/pull/496/commits/f67ad6e9db226e8c149ff3ab08af6fd708cf0f2a

Also, there were some unintended changes in get_dot_all, but I’ve excluded those 🙏.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
```diff
@@ -179,13 +181,150 @@ impl RegExpPrototype {
...
         _: ArgumentsList,
-        _gc: GcScope<'_, '_>,
+        mut gc: GcScope<'_, '_>,
     ) -> JsResult<Value> {
-        todo!()
+        // 1. Let R be the this value.
+        let r = this_value;
+
+        // 2. If R is not an Object, throw a TypeError exception.
+        if !r.is_object() {
```
aapoalas, line 195:
```suggestion
        let Ok(r) = Object::try_from(t) else {
```

This lets you avoid the unwraps and such later.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
```diff
@@ -179,13 +181,150 @@ impl RegExpPrototype {
...
+                ExceptionType::TypeError,
+                "value is not object",
+                gc.nogc(),
+            ));
+        };
+
+        // 3. Let codeUnits be a new empty List.
+        let mut code_units = Vec::new();
+
+        let Some(obj) = Object::internal_prototype(Object::try_from(r).unwrap(), agent) else {
```
aapoalas, line 206: **issue:** Spec does not say to read r's prototype.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
```diff
@@ -179,13 +181,150 @@ impl RegExpPrototype {
...
+            return Err(agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "value is not object",
+                gc.nogc(),
+            ));
+        };
+
+        let has_indices_args = get(
+            agent,
+            obj,
```
aapoalas, line 216: **issue:** Spec says to `Get(r ,"hasIndices")` but you're now effectively doing `Get(r[[Prototype]], "hasIndices")` which isn't (exactly) the same thing.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
```diff
@@ -179,13 +181,150 @@ impl RegExpPrototype {
...
+        if !r.is_object() {
+            return Err(agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "value is not object",
+                gc.nogc(),
+            ));
+        };
+
+        // 3. Let codeUnits be a new empty List.
+        let mut code_units = Vec::new();
```
eliassjogreen, line 204: **suggestion:** Probably entirely unnecessary, overkill and optimized by the compiler either way but we could avoid heap allocations here because we know that regexp flags in javascript are at most 8 and could easily be represented as a set, bit field or array with capacity 8. Possibly just using a `[char; 8]` could do the trick.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
yossydev: I see, I did it without realizing.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
aapoalas: Even better actually: It seems to be impossible to have all 8 flags on at the same time, try `new RegExp("a", "dgimsuvy")`, so 7 characters is the maximum it will ever take to represent the whole flags string. That happens to also be our SmallString limit so that's nice :)

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
aapoalas, line 216: **issue:** All of the get's should be done on `r` and not on `obj`.

aapoalas (changes requested):
Issue regarding the prototype, but aside from that this looks great! Nice work <3

If you want to, you could also do `let r = r.scope(agent, gc.nogc());` before the first `Get` call, and then in each `get` use `r.get(agent)` as the argument: This means that the function is then interleaved garbage-collector proof and won't need any work in the future :)

Making RegExp constructors work requires quite a bit more work, and especially making RegExp matching work will be a whole other ball game! That requires either bringing in a RegExp engine dependency, or building one :)

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
yossydev: > because we know that regexp flags in javascript are at most 8 and could easily

Exactly…!! Thank you for the review!

https://github.com/trynova/nova/pull/496/commits/87184e998dde62fc00b56641e64158998f19ea13

I tried implementing it using the `array with capacity 8` method, but does it align with the intent?

yossydev: > 7 characters is the maximum

Oh, sorry, I missed that. So the maximum is 7, got it.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
yossydev: Thank you for the review! I fixed it in https://github.com/trynova/nova/pull/496/commits/fe908ba8060f7863f3a6c4ac24bb86e6974f5660!

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
```diff
@@ -179,13 +181,150 @@ impl RegExpPrototype {
...
+            ));
+        };
+
+        let has_indices_args = get(
+            agent,
+            obj,
+            BUILTIN_STRING_MEMORY.hasIndices.into(),
+            gc.reborrow(),
+        )?;
+        // 4. Let hasIndices be ToBoolean(? Get(R, "hasIndices")).
```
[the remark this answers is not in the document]
yossydev: I was a bit unsure about the position of the comment, but your suggestion definitely feels more natural.

I’ve fixed it in [05850b3](https://github.com/trynova/nova/commit/05850b3)!

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
aapoalas: ie. You should just remove the entire `internal_prototype` call and the `obj` local variable. Neither are asked for by the spec.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
aapoalas: Still needs to be done.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
```diff
@@ -179,13 +181,154 @@ impl RegExpPrototype {
...
+
+        // 19. If sticky is true, append the code unit 0x0079 (LATIN SMALL LETTER Y) to codeUnits.
+        if sticky {
+            code_units[7] = Some('v');
+        };
+
+        // 20. Return the String value whose code units are the elements of the List codeUnits. If codeUnits has no elements, the empty String is returned.
+        Ok(Value::from_string(
+            agent,
+            code_units.iter().filter_map(|&opt| opt).collect(),
```
aapoalas, line 329: **issue (nonblocking):**  This is still allocating a string.

The way I'd do this is this:
```rs
    let mut str: [u8; 3] = [0; 3];
    let mut i = 0;
    
    if a > 10 {
        str[i] = b'a';
        i += 1;
    }
    if b > 10 {
        str[i] = b'b';
        i += 1;
    }
    if c > 10 {
        str[i] = b'c';
        i += 1;
    }
    
    let res: &str = unsafe { std::str::from_utf8_unchecked(&str[0..i]) };
```
ie. Keep track of where in the `code_units` you're assigning to, and finally take a slice of the array and interpret that as a UTF-8 string: We're guaranteed to get a proper UTF-8 string without checking because we know exactly the characters we put into the array, and those are all ASCII characters.

aapoalas (changes requested):
Still issues around the `R` vs `R[[Prototype]]`.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
aapoalas: Oh, now that I think about it: The maximum for real RegExps is 7 but a user-created RegExp-like object could easily return `true` for all of the getter-questions, so it's better to err for safety and keep the maximum capacity of 8.

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
yossydev: Apologies, I missed this earlier…
This is probably related to https://github.com/trynova/nova/pull/496#discussion_r1902783618, right? Thank you for the review!

I addressed it in [4ab2b661ff769d7d1ae735c9e6643df83ee78f0d](https://github.com/trynova/nova/commit/4ab2b661ff769d7d1ae735c9e6643df83ee78f0d)!

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
yossydev: Thanks again! I made the fixes based on your comments!
https://github.com/trynova/nova/pull/496/commits/d0799132bbc5faa07dd65a01384c6ce8994092a9

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
```diff
@@ -179,13 +180,152 @@ impl RegExpPrototype {
...
+        let Ok(r) = Object::try_from(r) else {
+            return Err(agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "value is not object",
+                gc.nogc(),
+            ));
+        };
+
+        // 3. Let codeUnits be a new empty List.
+        let mut code_units: [u8; 7] = [0; 7];
```
aapoalas, line 203: **issue:** Just to be safe against bad user attacks, let's make this 8 after all.
```suggestion
        let mut code_units: [u8; 8] = [0; 8];
```

aapoalas (approved):
LGTM! Thank you, and good work <3

nova_vm/src/ecmascript/builtins/text_processing/regexp_objects/regexp_prototype.rs
yossydev: Ahhh, my bad, it was still 7…
You even made the corrections for me! Sorry for the trouble!

aapoalas: No worries and nothing to say "sorry" about :) It's no trouble, rather the exact opposite. I'm so happy that you're contributing to Nova, and I'm terribly over-eager to get the PRs merged so if it's simple stuff then I'll rush to writing a suggestion snippet, and to get the code in just that little faster I might even merge it myself if it's a trivial change like this :)

aapoalas: Also just FYI: With these sorts of trivial suggestion commits, I will remove the "coauthored by Aapo Alasuutari" text from the final commit message because I don't particularly want to "steal valor" by making it seem like I was part of writing the commit.

PR #497, opened 2025-01-03
feat(ecmascipt): is_reg_exp

yossydev:
I implemented [isRegExp](https://tc39.es/ecma262/#sec-isregexp) and used it in places that were previously marked as TODO. This improved several test262 cases.


nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -121,6 +125,43 @@ pub(crate) fn try_is_extensible(
     o.try_is_extensible(agent, gc)
 }
 
+/// ### [7.2.6 IsRegExp ( argument )](https://tc39.es/ecma262/#sec-isregexp)
+///
+/// The abstract operation IsRegExp takes argument
+/// argument (an ECMAScript language value) and returns either a normal completion containing a Boolean or a throw completion.
+/// It performs the following steps when called
```
eliassjogreen, line 132:
```suggestion
```
**nitpick:** We usually omit the `"It performs the following steps when called"` from the spec comment.

See the [contributing](https://github.com/trynova/nova/blob/main/CONTRIBUTING.md#align-with-the-ecmascript-specification) docs:
> Once you've copied the specification text in as a comment, you can start turning the commented out abstract operation into a function. Start by changing the header and comment part of the abstract operation into a doc comment and make the header a level three header with a link to the original specification. Preferably, wrap the comment part into multiple lines (generally 80 is used as the line width point). Finally, remove the repetitive and redundant "It performs the following steps when called:" part.

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -121,6 +125,43 @@ pub(crate) fn try_is_extensible(
...
+
+    // 3. If matcher is not undefined, return ToBoolean(matcher).
+    if matcher.is_undefined() {
+        return Ok(to_boolean(agent, matcher));
+    }
+
+    // 4. If argument has a [[RegExpMatcher]] internal slot, return true.
+    if let Value::RegExp(_) = argument {
+        return Ok(true);
+    }
```
eliassjogreen, line 158: **suggestion:** This should be the most common case and therefor probably be done first? The `get` abstract operation is currently really slow so when we can we probably want to do that last. But I am unsure if we are allowed to re-order these or if step 2 possibly is mutating (cc @aapoalas)?

aapoalas: Unfortunately reordering those lines would be observable and thus isn't allowed :(

image

The "optimal" strategy here would be to split this into a `try_` method that takes a `NoGcScope` and uses `try_get`: Callers of `is_reg_exp` could then optimistically call that method and if they get a `TryResult::Break(())` then they'd root stuff and prepare to call into JavaScript: Because `Symbol.match` is very unlikely to be a getter, that path is basically never taken and rooting is essentially always avoided.

aapoalas (approved):
LGTM!

PR #498, opened 2025-01-03
feat(test262): Implement the `detachArrayBuffer` test harness function

eliassjogreen:


aapoalas (approved):
LGTM, nice!

aapoalas (changes requested):
Actually, on second thought: Maybe rather make the detach a public method on the ArrayBuffer struct that just calls the AO. It's maybe a bit cleaner than publishing AOs.

nova_vm/src/heap/heap_gc.rs
```diff
@@ -1023,6 +1025,8 @@ fn sweep(
     let Heap {
         #[cfg(feature = "array-buffer")]
         array_buffers,
+        #[cfg(feature = "array-buffer")]
+        array_buffer_detach_keys,
```
aapoalas, line 1029: **praise:** You dog, you! This is _exactly_ what I wanted <3 Detach keys are rare so saving them in the AB itself would be a waste. I could kiss you for doing it like this <3

nova_vm/src/ecmascript/builtins/array_buffer/abstract_operations.rs
```diff
@@ -18,6 +18,8 @@ use crate::{
     Heap,
 };
 
+// TODO: Implement the contents of the `DetachKey` struct?
```
aapoalas, line 21: **thought:** I wonder what the correct choice here would be... `Box<dyn Any>` or something like that might be the trick... But I'm not sure.

But let's leave it for later :)

aapoalas (approved):

PR #504, opened 2025-01-04
feat(ecmascript): proxy internal has prototype

yossydev:
Implemented the has prototype for the proxy and passed Test262!

doc: https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-hasproperty-p


tests/expectations.json
```diff
@@ -3147,32 +3143,18 @@
...
   "built-ins/Proxy/has/return-false-target-prop-exists-using-with.js": "CRASH",
-  "built-ins/Proxy/has/return-false-target-prop-exists.js": "CRASH",
   "built-ins/Proxy/has/return-false-targetdesc-not-configurable-using-with.js": "CRASH",
-  "built-ins/Proxy/has/return-false-targetdesc-not-configurable.js": "CRASH",
-  "built-ins/Proxy/has/return-is-abrupt-in.js": "CRASH",
   "built-ins/Proxy/has/return-is-abrupt-with.js": "CRASH",
   "built-ins/Proxy/has/return-true-target-prop-exists-using-with.js": "CRASH",
-  "built-ins/Proxy/has/return-true-target-prop-exists.js": "CRASH",
-  "built-ins/Proxy/has/return-true-without-same-target-prop.js": "CRASH",
   "built-ins/Proxy/has/trap-is-missing-target-is-proxy.js": "CRASH",
```
yossydev, line 3154: I feel like this should pass, but the following part returns false, causing the test to fail.

```js
// ref: https://github.com/tc39/test262/blob/a799a4ac23a466950158f940a977968a743f3059/test/built-ins/Proxy/has/trap-is-missing-target-is-proxy.js#L29-L33

var functionTarget = new Proxy(function() {}, {});
var functionProxy = new Proxy(functionTarget, {});

assert("name" in functionProxy);
assert("length" in Object.create(functionProxy));
```

I’m not quite sure about the cause, so if you have any ideas, please let me know.

aapoalas: I'll take a quick look.

aapoalas: Alright, it's actually failing on the line testing "lastIndex" property. This is because our RegExpAlloc method currently doesn't define it. That is also intentional, as we don't want to just use the backing object for this property; it's usually the only property in a RegExp object and creating the whole backing object and all that it entails for every RegExp is wasteful.

What we'd rather do is keep a `last_index: Value` in the RegExpHeapData and change the internal methods of RegExp to account for that property. One annoying part is that, if I'm not badly mistaken, the property can made `writable: false` so we will also need one bit for that somewhere, but it's unlikely we have any good place for that and as a result will likely need to pay 4 bytes for that little thing.

aapoalas: Second thought: Storing a Value is insane. A u32 will suffice 99% of the time.

aapoalas (approved):
LGTM! Nice work again <3

tests/expectations.json
yossydev: Thank you for looking into it! I see, it was related to the RegExp part.
Thank you!

PR #507, opened 2025-01-05
feat(ecmascript): proxy set method

yossydev:
I implemented the set method for the proxy.
I think its contents are mostly the same as the get method.

doc:https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-set-p-v-receiver


aapoalas (approved):
LGTM, thank you <3 You're really cranking these out!

PR #510, opened 2025-01-05
feat(ecmascript): proxy call method

yossydev:
I implemented the call method (also part of apply), but the number of crashes in test262 has increased. For example:

https://github.com/tc39/test262/blob/ff9763729d242b54a9f11bbc614670db3888c8d2/test/built-ins/Proxy/apply/call-result.js

When `p.call(Function.prototype.call)` is invoked, and p.apply is called, it results in the error:

Uncaught exception: TypeError: Not a callable value

I believe this is related to the following section of code:

https://github.com/trynova/nova/blob/55e48f94ed54b309f724ecbae28c2bd3f4b243ee/nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs#L67-L89

Here, the issue is that it returns None. As the comment suggests, I suspect that if we change `Option<Function>` to `Option<Callable>`, the `try_into()` method would return true and resolve the issue.

doc: https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-call-thisargument-argumentslist

### DIFF
--- a/nova_vm/src/ecmascript/builtins/proxy.rs
+++ b/nova_vm/src/ecmascript/builtins/proxy.rs
@@ -798,7 +798,6 @@ impl InternalMethods for Proxy {
             mut handler,
         } = validate_non_revoked_proxy(agent, self, gc.nogc())?;
         // 5. Let trap be ? GetMethod(handler, "apply").
-        let scoped_target = target.scope(agent, gc.nogc());
         let trap = if let TryResult::Continue(trap) = try_get_object_method(
             agent,
             handler,
@@ -808,6 +807,7 @@ impl InternalMethods for Proxy {
             trap?
         } else {
             let scoped_handler = handler.scope(agent, gc.nogc());
+            let scoped_target = target.scope(agent, gc.nogc());
             let trap = get_object_method(
                 agent,
                 handler,
--- a/nova_vm/src/ecmascript/builtins/proxy.rs
+++ b/nova_vm/src/ecmascript/builtins/proxy.rs
@@ -835,9 +835,9 @@ impl InternalMethods for Proxy {
         // 7. Let argArray be CreateArrayFromList(argumentsList).
         let arg_array = create_array_from_list(agent, &[arguments_list], gc.nogc());
         // 8. Return ? Call(trap, handler, « target, thisArgument, argArray »).
-        return call(
+        return call_function(
             agent,
-            trap.into_value(),
+            trap.unbind(),
             handler.into_value(),
             Some(ArgumentsList(&[
                 target.into_value(),
--- a/nova_vm/src/ecmascript/builtins/proxy.rs
+++ b/nova_vm/src/ecmascript/builtins/proxy.rs
@@ -773,6 +773,13 @@ impl InternalMethods for Proxy {
         todo!();
     }
 
+    /// ### [10.5.12 [[Call]] ( thisArgument, argumentsList )](https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-call-thisargument-argumentslist)
+    ///
+    /// The [[Call]] internal method of a Proxy exotic object O takes
+    /// arguments thisArgument (an ECMAScript language value)
+    /// and argumentsList (a List of ECMAScript language values)
+    /// and returns either a normal completion containing an ECMAScript
+    /// language value or a throw completion.
     fn internal_call(
         self,
         agent: &mut Agent,
--- a/nova_vm/src/ecmascript/builtins/proxy.rs
+++ b/nova_vm/src/ecmascript/builtins/proxy.rs
@@ -10,7 +10,10 @@ use data::ProxyHeapData;
 use crate::{
     ecmascript::{
         abstract_operations::{
-            operations_on_objects::{call_function, get_object_method, try_get_object_method},
+            operations_on_objects::{
+                call, call_function, create_array_from_list, get_object_method,
+                try_get_object_method,
+            },
             testing_and_comparison::{is_extensible, same_value},
             type_conversion::to_boolean,
         },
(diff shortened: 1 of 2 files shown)


aapoalas:
Yup, we've not had callable proxies so `Callable` doesn't exist at all.

Do you want to try your hand at implementing it? It would basically just be 90% direct copy of the `Function` enum but with `Proxy` added to it. The `is_callable` function will then need to be changed to check if a `Proxy`'s target `is_callable`.

yossydev:
> Yup, we've not had callable proxies so `Callable` doesn't exist at all.
> 
> Do you want to try your hand at implementing it? It would basically just be 90% direct copy of the `Function` enum but with `Proxy` added to it. The `is_callable` function will then need to be changed to check if a `Proxy`'s target `is_callable`.

I’d love to do it! I’ll probably start implementing it tomorrow or the day after!

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -770,14 +773,79 @@ impl InternalMethods for Proxy {
...
+        // 1. Perform ? ValidateNonRevokedProxy(O).
+        // 2. Let target be O.[[ProxyTarget]].
+        // 3. Let handler be O.[[ProxyHandler]].
+        // 4. Assert: handler is an Object.
+        let NonRevokedProxy {
+            mut target,
+            mut handler,
+        } = validate_non_revoked_proxy(agent, self, gc.nogc())?;
+        // 5. Let trap be ? GetMethod(handler, "apply").
+        let scoped_target = target.scope(agent, gc.nogc());
```
aapoalas, line 801: **nitpick:** Here we probably shouldn't be unconditionally scoping `target`: It's very unlikely that the `apply` property is itself a getter so it's likely that the `try_get_object_method` will succeed. If it does, then `target` is guaranteed to still be valid during steps `6.a` and `8`: Both of those can trigger GC, but the `target` value is no longer needed after those steps so effectively it's very likely that scoping `target` is unnecessary.

You should just move the scoping of `target` into the `else` branch to live together with the handler scoping.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -770,14 +773,79 @@ impl InternalMethods for Proxy {
...
+            let gc = gc.nogc();
+            let trap = trap.map(|f| f.bind(gc));
+            handler = scoped_handler.get(agent).bind(gc);
+            target = scoped_target.get(agent).bind(gc);
+            trap
+        };
+        // 6. If trap is undefined, then
+        let Some(trap) = trap else {
+            // a. Return ? Call(target, thisArgument, argumentsList).
+            return call(
```
aapoalas, line 827: **issue:** Use `call_function` since we're guaranteed to get a Function (well, a `Callable` once we have that). `call` method checks that the function being called is callable, but that's a redundant check if we statically know that we're calling a callable object.

aapoalas (changes requested):
One nit and one issue (in two places though; remember to fix both `call()`s to be `call_function()`).

But generally looks good to me! Thank you again <3 We're really going places this week; we're now well past 50% passing rate!

nova_vm/src/ecmascript/builtins/proxy.rs
yossydev: In this part, since target is an object, I was using call :eyes:

```rust
let Some(callback_fn) = is_callable(target, gc.nogc()) else {
        ExceptionType::TypeError,
        "Callback function is not a function",
        gc.nogc(),
    ));
};
```

↑ Would it be necessary to insert a process like this beforehand…?? 👀

aapoalas: Oh, I had missed that we're performing the Call AO on target and not trap.

I guess here using the normal Call method is fine then. Technically I'm pretty sure that it should be impossible to end up in here unless target is itself callable (since this method should never be called on anything but callable Proxies), but it's safer and pretty simple to just use the Call method and throw an error if this isn't a callable Proxy.

nova_vm/src/ecmascript/builtins/proxy.rs
yossydev: Thank you! You’re absolutely right!
I’ve fixed it! https://github.com/trynova/nova/pull/510/commits/ba71fc8c76fe4d7125be616a9a81fe806bd0e822

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -770,14 +773,79 @@ impl InternalMethods for Proxy {
...
+                Some(ArgumentsList(&[arguments_list])),
+                gc,
+            );
+        };
+        // 7. Let argArray be CreateArrayFromList(argumentsList).
+        let arg_array = create_array_from_list(agent, &[arguments_list], gc.nogc());
+        // 8. Return ? Call(trap, handler, « target, thisArgument, argArray »).
+        return call(
+            agent,
+            trap.into_value(),
```
aapoalas, line 840: But here at least we can use `call_function`

yossydev: Thank you for the comment! I’m encountering the following error:

```rs
error[E0505]: cannot move out of `gc` because it is borrowed
   --> nova_vm/src/ecmascript/builtins/proxy.rs:850:13
    |
791 |         mut gc: GcScope<'_, '_>,
    |         ------ binding `gc` declared here
...
808 |             gc.nogc(),
    |             -- borrow of `gc` occurs here
...
841 |         return call_function(
    |                ------------- borrow later used by call
...
850 |             gc,
    |             ^^ move out of `gc` occurs here
```

Using call works without issues, but with call_function, I get this error. Honestly, I’m not entirely sure why this happens. Do you have any ideas about what might be causing this? :eyes:

nova_vm/src/ecmascript/builtins/proxy.rs
yossydev: ok, thank you!

nova_vm/src/ecmascript/builtins/proxy.rs
aapoalas: When doing `call_function(agent, trap, .., gc)` add `.unbind()` to the trap parameter.

This is basically the GC safety system in action, it just doesn't like function interface boundaries :)

yossydev: I was able to get it working by applying the fix in https://github.com/trynova/nova/pull/510/commits/4c31cd728b7b053caadc137791c0ceda70e5361c!

Thank you!!! I haven’t really gone through this part of the code much, so I’ll make sure to read it. Rust is tough…

aapoalas: Don't worry, the GC safety system is something I'm only managing to get working right now, and it's pretty awkward even though it does work. This also isn't exactly your run of the mill Rust: This is very much abusing the Rust borrow checker's mechanisms to get static checking of ... well, of rooting and garbage collector safepoints.

I'll probably be writing a more complete contributor's guide to this stuff, and a more generally accessible blog post as well. I'll also give a talk on this at FOSDEM at the turn of the month: https://fosdem.org/2025/schedule/event/fosdem-2025-4394-abusing-reborrowing-for-fun-profit-and-a-safepoint-garbage-collector/

aapoalas (approved):
LGTM, good job! <3

nova_vm/src/ecmascript/builtins/proxy.rs
yossydev: Oh, that’s exciting!!!
I’m looking forward to the blog, but do you know if there will be an archive of the talk as well…?

aapoalas: Yes, unless something goes badly wrong the video will be available online.

yossydev:
> Yup, we've not had callable proxies so `Callable` doesn't exist at all.
> 
> Do you want to try your hand at implementing it? It would basically just be 90% direct copy of the `Function` enum but with `Proxy` added to it. The `is_callable` function will then need to be changed to check if a `Proxy`'s target `is_callable`.

hi, @aapoalas 

The Callable struct would look like the following, right?

```rs
#[derive(Clone, Copy, PartialEq)]
#[repr(u8)]
pub enum Callable<'a> {
    BoundFunction(BoundFunction<'a>) = BOUND_FUNCTION_DISCRIMINANT,
    BuiltinFunction(BuiltinFunction<'a>) = BUILTIN_FUNCTION_DISCRIMINANT,
    ECMAScriptFunction(ECMAScriptFunction<'a>) = ECMASCRIPT_FUNCTION_DISCRIMINANT,
    BuiltinGeneratorFunction = BUILTIN_GENERATOR_FUNCTION_DISCRIMINANT,
    BuiltinConstructorFunction(BuiltinConstructorFunction<'a>) =
        BUILTIN_CONSTRUCTOR_FUNCTION_DISCRIMINANT,
    BuiltinPromiseResolvingFunction(BuiltinPromiseResolvingFunction<'a>) =
        BUILTIN_PROMISE_RESOLVING_FUNCTION_DISCRIMINANT,
    BuiltinPromiseCollectorFunction = BUILTIN_PROMISE_COLLECTOR_FUNCTION_DISCRIMINANT,
    BuiltinProxyRevokerFunction = BUILTIN_PROXY_REVOKER_FUNCTION,
    Proxy(Proxy) = PROXY_DISCRIMINANT,
}
```
And in this case, would you also add things like unbind to the Callable enum? Would it essentially be the same implementation as Function? 👀

aapoalas:
Yup, exactly

PR #513, opened 2025-01-05
feat(ecmascript): Add support for proxies in `is_array` abstract operation

eliassjogreen:


nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
```diff
@@ -52,16 +55,29 @@ pub(crate) fn require_object_coercible(
...
-    // TODO: Proxy
-    // 3. If argument is a Proxy exotic object, then
-    // a. Perform ? ValidateNonRevokedProxy(argument).
-    // b. Let proxyTarget be argument.[[ProxyTarget]].
-    // c. Return ? IsArray(proxyTarget).
-    // 4. Return false.
+pub(crate) fn is_array(
+    agent: &mut Agent,
+    argument: impl IntoValue,
+    gc: NoGcScope<'_, '_>,
```
aapoalas, line 61: **thought:** Whew, thank heavens this at least cannot call into JavaScript :D

aapoalas (approved):
LGTM!

nova_vm/src/ecmascript/abstract_operations/testing_and_comparison.rs
eliassjogreen: Yep! And we should really add that feature flag for proxies so this can just be a check for the array variant...

aapoalas:
Need to resolve the conflict: I'll merge immediately after :)

eliassjogreen:
I just went out so if you could do that that would be great :)

PR #515, opened 2025-01-07
feat(ecmascript): proxy set prototype of

yossydev:
I implemented the setPrototypeOf for Proxy!

doc: https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-setprototypeof-v


nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -227,13 +227,85 @@ impl InternalMethods for Proxy {
...
+        // 5. Let trap be ? GetMethod(handler, "setPrototypeOf").
+        let trap = if let TryResult::Continue(trap) = try_get_object_method(
+            agent,
+            handler,
+            BUILTIN_STRING_MEMORY.setPrototypeOf.into(),
+            gc.nogc(),
+        ) {
+            trap?
+        } else {
+            let scoped_target: crate::engine::Scoped<'_, Object> = target.scope(agent, gc.nogc());
```
aapoalas, line 258: **nitpick:** Type here seems unnecessary, or at least should use the "Scoped" directly instead of the full qualified name.

yossydev: Thank you! I didn’t intend to add explicit types, but it seems they were there after all! I’ve removed them to match the rest. https://github.com/trynova/nova/pull/515/commits/0fcc470c2191c1b5e080b66197bd802ad3a9e4cb

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -227,13 +227,85 @@ impl InternalMethods for Proxy {
...
+        // 6. If trap is undefined, then
+        let Some(trap) = trap else {
+            // a. Return ? target.[[SetPrototypeOf]](V).
+            return target.internal_set_prototype_of(agent, prototype, gc);
+        };
+        // 7. Let booleanTrapResult be ToBoolean(? Call(trap, handler, « target, V »)).
+        let argument = call_function(
+            agent,
+            trap.unbind(),
+            handler.into_value(),
```
aapoalas, line 282: **nitpick:** Here we'd actually need to have the target scoped again as it's used after this call.

But, that will become a build time failure later down the line anyway so I'm not too worried about it. Let's just get this merged :)

aapoalas (approved):
LGTM, thank you for all this Proxy work! If I'm not badly mistaken, we're pretty close to having all the methods implemented now right?

yossydev:
> LGTM, thank you for all this Proxy work! If I'm not badly mistaken, we're pretty close to having all the methods implemented now right?

Yes, it will be done by the end of the week!

Issue #158, opened 2024-04-15
AsyncGeneratorFunction builtin functions


aapoalas:

aapoalas:
I believe all the related builtin functions for this have been implemented.

PR #516, opened 2025-01-07
feat(ecmascript): async generator function constructor

yossydev:
ref: https://github.com/trynova/nova/issues/158
doc: https://tc39.es/ecma262/#sec-asyncgeneratorfunction


aapoalas (dismissed):
LGTM! Let's see if we can get the CI to pass as well...

PR #517, opened 2025-01-08
feat(ecmascript): Proxy getOwnProperty

yossydev:
I have implemented getOwnProperty for the proxy and the complete_property_descriptor function needed within it. The amount of code turned out to be larger than I expected, and there are a lot of conditional expressions. I believe it can be written more cleanly, so I’d appreciate it if you could review it. 🙏

doc: https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-getownproperty-p


yossydev:
Oops, I left the error messages as placeholders since I was planning to fix them later! I’ll go ahead and fix them now. 🙏

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -463,13 +465,166 @@ impl InternalMethods for Proxy {
...
+                    gc.nogc(),
+                ));
+            };
+            // e. Return undefined.
+            return Ok(None);
+        };
+        // 11. Let extensibleTarget be ? IsExtensible(target).
+        let extensible_target = is_extensible(agent, target, gc.reborrow())?;
+        // 12. Let resultDesc be ? ToPropertyDescriptor(trapResultObj).
+        let result_desc = if let TryResult::Continue(result_desc) =
```
aapoalas, line 571: **issue:** To simplify things, let's just use the `to_property_descriptor` method here directly.

nova_vm/src/ecmascript/types/spec/property_descriptor.rs
```diff
@@ -493,6 +493,63 @@ impl PropertyDescriptor {
         TryResult::Continue(Ok(desc))
     }
 
+    /// ### [6.2.6.6 CompletePropertyDescriptor ( Desc )](https://tc39.es/ecma262/#sec-completepropertydescriptor)
+    ///
+    /// The abstract operation CompletePropertyDescriptor takes
+    /// argument Desc (a Property Descriptor) and returns unused.
+    pub(crate) fn complete_property_descriptor(
+        desc: Self,
```
aapoalas, line 501: **issue:** You cannot take descriptor by value here, as that means any changes you make here are made invisible to the caller and since this method returns UNUSED, the end result is that the caller gets absolutely nothing.

You want to take `&mut self` here instead and use `self` in place of `desc`.

nova_vm/src/ecmascript/types/spec/property_descriptor.rs
```diff
@@ -493,6 +493,63 @@ impl PropertyDescriptor {
...
+            get: None,
+            set: None,
+            enumerable: Some(false),
+            configurable: Some(false),
+        };
+        // 2. If IsGenericDescriptor(Desc) is true or IsDataDescriptor(Desc) is true, then
+        if desc.is_generic_descriptor() || desc.is_data_descriptor() {
+            // a. If Desc does not have a [[Value]] field, set Desc.[[Value]] to like.[[Value]].
+            let Some(_) = desc.value else {
+                like.value = desc.value;
```
aapoalas, line 518: **issue:** This is the wrong way around: `"set Desc.[[Value]] to like.[[Value]]` means `desc.value = like.value`.

Same issue repeats on every assignment line.

nova_vm/src/ecmascript/types/spec/property_descriptor.rs
```diff
@@ -493,6 +493,63 @@ impl PropertyDescriptor {
...
+    /// The abstract operation CompletePropertyDescriptor takes
+    /// argument Desc (a Property Descriptor) and returns unused.
+    pub(crate) fn complete_property_descriptor(
+        desc: Self,
+        _agent: &mut Agent,
+        _gc: NoGcScope<'_, '_>,
+    ) -> JsResult<()> {
+        // 1. Let like be the Record { [[Value]]: undefined, [[Writable]]: false, [[Get]]: undefined, [[Set]]: undefined, [[Enumerable]]: false, [[Configurable]]: false }.
+        let mut like: PropertyDescriptor = PropertyDescriptor {
+            value: None,
```
aapoalas, line 507: **issue:** This is a bit confusing, but in this case the `value` should be `Some(Value::Undefined)`, ie. it is present but has the value `undefined`.

That same logic doesn't work for `get` and `set` though, since we type them as `Option<Function>`. There `undefined` indeed means `None` :)

nova_vm/src/ecmascript/types/spec/property_descriptor.rs
```diff
@@ -493,6 +493,63 @@ impl PropertyDescriptor {
...
+                return Ok(());
+            };
+            // b. If Desc does not have a [[Set]] field, set Desc.[[Set]] to like.[[Set]].
+            let Some(_) = desc.set else {
+                like.set = desc.set;
+                return Ok(());
+            };
+        };
+        // 4. If Desc does not have an [[Enumerable]] field, set Desc.[[Enumerable]] to like.[[Enumerable]].
+        let Some(_) = desc.enumerable else {
```
aapoalas, line 540: **nitpick:** Since you're not binding the contents of `Some(_)`, you can just use `desc.enumerable.is_none()`.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -463,13 +465,166 @@ impl InternalMethods for Proxy {
...
+            return Err(agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "valid",
+                gc.nogc(),
+            ));
+        };
+        // 16. If resultDesc.[[Configurable]] is false, then
+        if result_desc.configurable == Some(false) {
+            // b. If resultDesc has a [[Writable]] field and resultDesc.[[Writable]] is false, then
+            if let Some(false) = result_desc.writable {
```
aapoalas, line 599: **nitpick:** You don't need to use the `let Some(false)` binding scheme but can instead use perhaps a little more self-explanatory `result_desc.writable == Some(false)`.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -463,13 +465,166 @@ impl InternalMethods for Proxy {
...
+                // i. Assert: targetDesc has a [[Writable]] field.
```
aapoalas, line 600: **nitpick:** You're missing `assert!(target_desc.writable.is_some());`

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -463,13 +465,166 @@ impl InternalMethods for Proxy {
...
+                // ii. If targetDesc.[[Writable]] is true, throw a TypeError exception.
+                if let Some(target_desc) = target_desc {
+                    if target_desc.writable == Some(true) {
```
aapoalas, line 603: **issue:** `target_desc.configurable != Some(true)` needs to be also explicitly checked against.

This might become a bit easier to parse if you just wrote it in the spec order:
```rs
// a. If targetDesc is undefined or targetDesc.[[Configurable]] is true, then
if target_desc.map_or(true, |d| d.configurable == Some(true)) {
  // i. Throw a TypeError exception.
 // "proxy can't report a non-existent property as non-configurable" goes here
}
let target_desc = target_desc.unwrap();
if result_desc.writable == Some(true) {
  assert!(target_desc.writable.is_some());
  // ...
}
```

aapoalas (changes requested):
Some issues around, but not too bad. Good work on a harder part of the spec! Definitely hard to parse what the spec wants at these twists and turns :)

Should be fairly straightforward to get this to pass basically all the tests! <3

yossydev:
@aapoalas 
Thank you so much for your review! I’ve addressed your comments in the following commit: fb931e9b1fe96e20eeb142e6cd5de3612df81ee9

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -463,13 +465,166 @@ impl InternalMethods for Proxy {
...
+        let Some(trap) = trap else {
+            // a. Return ? target.[[GetOwnProperty]](P).
+            return target.internal_get_own_property(agent, property_key, gc);
+        };
+        // 7. Let trapResultObj be ? Call(trap, handler, « target, P »).
+        let trap_result_obj = call_function(
+            agent,
+            trap.unbind(),
+            handler.into_value(),
+            Some(ArgumentsList(&[target.into(), property_key.into()])),
```
aapoalas, line 521: **issue:** `P` needs to use the `convert_to_value` method or something like that. The `Into` conversion is unfortunately very "dangerous" in that it doesn't actually do what you'd kind of expect. If in JS we're doing something like `obj[0]` or `obj["0"]` then we'll convert that into `PropertyKey::Integer(0)` and we have separate handling for that.

But then when you do `let p: Value = PropertyKey::Integer(0).into();` it actually turns that into `Value::Integer(0)` directly, meaning that now the trap is being called with `obj, 0` instead of `obj, "0"` as the spec would expect.

Eventually we'll probably want to get rid of the whole `into` conversion, it's just _very_ convenient in some places.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -463,13 +465,166 @@ impl InternalMethods for Proxy {
...
+            return Err(agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "valid",
+                gc.nogc(),
+            ));
+        };
+        // 16. If resultDesc.[[Configurable]] is false, then
+        if result_desc.configurable == Some(false) {
+            // a. If targetDesc is undefined or targetDesc.[[Configurable]] is true, then
+            if target_desc
```
aapoalas, line 594: **issue:** It should be possible to avoid the clone here with something like `target_desc.as_ref().map_or` kind of thing maybe?

nova_vm/src/ecmascript/types/spec/property_descriptor.rs
```diff
@@ -493,6 +493,63 @@ impl PropertyDescriptor {
         TryResult::Continue(Ok(desc))
     }
 
+    /// ### [6.2.6.6 CompletePropertyDescriptor ( Desc )](https://tc39.es/ecma262/#sec-completepropertydescriptor)
+    ///
+    /// The abstract operation CompletePropertyDescriptor takes
+    /// argument Desc (a Property Descriptor) and returns unused.
+    pub(crate) fn complete_property_descriptor(
+        &mut self,
+        _agent: &mut Agent,
```
aapoalas, line 502: **issue:** Let's just take out both the parameters since we're not using them.

nova_vm/src/ecmascript/types/spec/property_descriptor.rs
```diff
@@ -493,6 +493,63 @@ impl PropertyDescriptor {
...
+        _gc: NoGcScope<'_, '_>,
+    ) -> JsResult<()> {
+        // 1. Let like be the Record { [[Value]]: undefined, [[Writable]]: false, [[Get]]: undefined, [[Set]]: undefined, [[Enumerable]]: false, [[Configurable]]: false }.
+        let like: PropertyDescriptor = PropertyDescriptor {
```
aapoalas, line 506: **nitpick:** `like: PropertyDescriptor = PropertyDescriptor { ... }` is redundant, can be just `like = PropertyDescriptor { ... }`.

nova_vm/src/ecmascript/types/spec/property_descriptor.rs
```diff
@@ -493,6 +493,63 @@ impl PropertyDescriptor {
...
+            set: None,
+            enumerable: Some(false),
+            configurable: Some(false),
+        };
+        // 2. If IsGenericDescriptor(Desc) is true or IsDataDescriptor(Desc) is true, then
+        if self.is_generic_descriptor() || self.is_data_descriptor() {
+            // a. If Desc does not have a [[Value]] field, set Desc.[[Value]] to like.[[Value]].
+            if self.value.is_none() {
+                self.value = like.value;
+                return Ok(());
```
aapoalas, line 519: **issue:** None of these `self.x = like.x` branches should not be returning.

aapoalas (changes requested):
Found a few more issues. Sorry about the back-and-forth :(

yossydev:
> Found a few more issues. Sorry about the back-and-forth :(

Don’t worry! Please feel free to review my work without hesitation!

c30250b93464b58bfba07898702bba2d7d02db2f ← I’ve made the corrections based on your feedback in this commit!

PR #520, opened 2025-01-09
feat(ecmascript): AsyncGenerator

aapoalas:
In case you want to take a look, Andreu specifically since you created the excellent async functions system that I get to use here <3


nova_vm/src/ecmascript/builtins/control_abstraction_objects/async_generator_objects.rs
```diff
@@ -2,100 +2,412 @@
...
+    pub(crate) async_generator_state: Option<AsyncGeneratorState>,
+}
+
+#[derive(Debug)]
+pub(crate) enum AsyncGeneratorState {
+    // SUSPENDED-START has `vm_or_args` set to Arguments, SUSPENDED-YIELD has it set to Vm.
+    Suspended {
+        state: SuspendedGeneratorState,
+        queue: VecDeque<AsyncGeneratorRequest>,
+    },
```
andreubotella, line 258: In the `SUSPENDED-START` state, the queue will always be empty, except inside engine code between `%AsyncGeneratorPrototype%.next` steps 8-9.a. Would it make sense to split these states?

aapoalas: Hmm, yeah might make sense.

PR #521, opened 2025-01-11
feat(ecmascript): Proxy delete

yossydev:
I’ve implemented the delete method, and overall it wasn’t too complicated.
However, I have a question: to pass the test [built-ins/Proxy/deleteProperty/trap-is-null-target-is-proxy.js](https://github.com/tc39/test262/blob/main/test/built-ins/Proxy/deleteProperty/trap-is-null-target-is-proxy.js), is it necessary to implement an additional check to determine whether the code is running in strict mode?

doc: https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-delete-p


aapoalas:
> I have a question: to pass the test [built-ins/Proxy/deleteProperty/trap-is-null-target-is-proxy.js](https://github.com/tc39/test262/blob/main/test/built-ins/Proxy/deleteProperty/trap-is-null-target-is-proxy.js), is it necessary to implement an additional check to determine whether the code is running in strict mode?

Nope: The strict mode check is done as part of [delete operator evaluation](https://tc39.es/ecma262/#sec-delete-operator-runtime-semantics-evaluation) as per the step 4.f.:

> If deleteStatus is false and ref.[[Strict]] is true, throw a TypeError exception.

ie. It's only really concerning what the delete operator (in this case the delete trap / Proxy internal method) returns and has nothing to do with what it does internally.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -822,11 +822,95 @@ impl InternalMethods for Proxy {
...
+            let scoped_handler = handler.scope(agent, gc.nogc());
+            let scoped_target = target.scope(agent, gc.nogc());
+            let scoped_property_key = property_key.scope(agent, gc.nogc());
+            let trap = get_object_method(
+                agent,
+                handler,
+                BUILTIN_STRING_MEMORY.deleteProperty.into(),
+                gc.reborrow(),
+            )?
+            .map(Function::unbind);
```
aapoalas, line 855: **praise:** Oooh, interesting trick :)

aapoalas (dismissed):
LGTM, real nice! <3

aapoalas:
Oh, need to resolve the expectations / metrics conflicts that always arise whenever we merge basically any PR. That's probably the most annoying thing about our merging system at the moment :D

yossydev:
> Oh, need to resolve the expectations / metrics conflicts that always arise whenever we merge basically any PR. That's probably the most annoying thing about our merging system at the moment :D

そうですね。できればコンフリクトを解消するこの作業はなくせたら嬉しいです。
例えば、metrics.jsonはgitignoreしてデバッグ用に使用するようにし、今websiteのリポジトリ内で同様のmetricsの計算処理を書いてもいいかもしれない👀
ただ今はsubmoduleでtest262をcloseしてくるような感じになっていると思うから、ここから変えないといけない...？？

yossydev:
I don’t think it’s that difficult either.

> There are of course still conflict possibilities with expectations.json but it’s definitely way less than the essentially guaranteed conflict with metrics.json. This just hasn’t been a huge issue when PRs have been fairly rare and most often appeared only one at a time :)

That’s true. Also, if we’re going to put effort into that, it’d be better to focus on advancing the implementation of the Nova core first for now.

PR #522, opened 2025-01-12
feat(ecmascript): Proxy ownKeys

yossydev:
doc: https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-ownpropertykeys


nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -1230,10 +1231,166 @@ impl InternalMethods for Proxy {
...
+                    "Duplicate entries found in trapResult",
+                    gc.nogc(),
+                ));
+            }
+        }
+        // 10. Let extensibleTarget be ? IsExtensible(target).
+        let extensible_target = is_extensible(agent, target, gc.reborrow())?;
+        // 11. Let targetKeys be ? target.[[OwnPropertyKeys]]().
+        let target_keys =
+            unbind_property_keys(target.internal_own_property_keys(agent, gc.reborrow())?);
```
aapoalas, line 1300: **issue:** Best do a `bind_property_keys()` after an unbind to make sure we're ready for interleaved GC.

If the borrow checker becomes angry at that, it then means that the property keys need to be scoped using the `scope_property_keys` helper function.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -1230,10 +1231,166 @@ impl InternalMethods for Proxy {
...
+        // 13. Assert: targetKeys contains no duplicate entries.
+        for i in &trap_result {
+            assert!(trap_result.contains(&i))
```
aapoalas, line 1303: **issue:** This doesn't actually check that there are no duplicates.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -1230,10 +1231,166 @@ impl InternalMethods for Proxy {
...
+        // 14. Let targetConfigurableKeys be a new empty List.
+        let mut target_configurable_keys = Vec::new();
+        // 15. Let targetNonconfigurableKeys be a new empty List.
+        let mut target_nonconfigurable_keys = Vec::new();
+        // 16. For each element key of targetKeys, do
+        for &key in target_keys.iter() {
+            // a. Let desc be ? target.[[GetOwnProperty]](key).
+            let desc = target.internal_get_own_property(agent, key.unbind(), gc.reborrow())?;
+            //  b. If desc is not undefined and desc.[[Configurable]] is false, then
+            if desc.unwrap().configurable == Some(false) {
```
aapoalas, line 1314: **issue:** "is not undefined" means essentially `desc.map_or(false, |d| d.configurable == Some(false))`.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -1230,10 +1231,166 @@ impl InternalMethods for Proxy {
...
+            } else {
+                // c. Else,
+                // i. Append key to targetConfigurableKeys.
+                target_configurable_keys.push(key);
+            }
+        }
+        // 17. If extensibleTarget is true and targetNonconfigurableKeys is empty, then
+        if extensible_target && target_nonconfigurable_keys.is_empty() {
+            // a. Return trapResult.
+            let mut property_key_list = Vec::with_capacity(trap_result.len());
```
aapoalas, line 1326: **note:** With a "PROPERTY-KEY" variant of "CreateListFromArrayLike" (eg. `create_property_key_list_from_array_like`) we should be able to return `trap_result` pretty directly here (though we probably need to use a scoped version and unscope it; that being said, we can perhaps handle the scoping directly in the "CreateListFromArrayLike" function variant).

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -1230,10 +1231,166 @@ impl InternalMethods for Proxy {
...
+            for v in trap_result {
+                property_key_list.push(unsafe { PropertyKey::from_value_unchecked(v) });
```
aapoalas, line 1328: **issue:** Not here exactly, but in the "CreateListFromArrayLike" function, you should use the proper conversion function, not the unchecked version.

The spec text requires that users return only strings or symbols from the property keys trap, but we expect indexed property keys to contain actual integers. The `from_value_unchecked` does not perform the conversion, so if a trap function in JS did `return ["0", "a"]` then using `from_value_unchecked` will lead to `PropertyKey::SmallString("0")` instead of `PropertyKey::Integer(0)`, and that will lead to odd errors.

aapoalas (changes requested):
Unfortunately this needs a bit work. Let me know if you want help with this, this is a complicated method. If you want to, we can also do pair-programming on Discord or such.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -1230,10 +1231,166 @@ impl InternalMethods for Proxy {
...
+        let trap_result_array = call_function(
+            agent,
+            trap.unbind(),
+            handler.into_value(),
+            Some(ArgumentsList(&[target.into_value()])),
+            gc.reborrow(),
+        )?;
+        // 8. Let trapResult be ? CreateListFromArrayLike(trapResultArray, property-key).
+        let trap_result =
+            create_list_from_array_like(agent, trap_result_array.unbind(), gc.reborrow())?;
```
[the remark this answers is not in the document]
yossydev: Thank you. Yes, this is exactly what I was planning to ask about.
As I thought, such a solution is necessary, right?

nova_vm/src/ecmascript/builtins/proxy.rs
yossydev: Sorry, I was planning to fix this later.

nova_vm/src/ecmascript/builtins/proxy.rs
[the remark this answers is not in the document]
aapoalas: Yeah: We either need to change the function to do have the `ALL` vs `PROPERTY-KEY` parameter, or write a completely different function. And in this case, it is pretty much guaranteed that we want a custom function because the return value changes on the parameter type, so writing a generic function that works for both cases is painful.

yossydev:
> Unfortunately this needs a bit work. Let me know if you want help with this, this is a complicated method. If you want to, we can also do pair-programming on Discord or such.


Thank you! That was really helpful!
I have plans today, so I probably won’t be able to continue working on this, but I’ll check it tomorrow!

I’d really love to do pair-programming, but I can’t speak any language other than Japanese…
I’m currently studying English, though!

aapoalas:
> > Unfortunately this needs a bit work. Let me know if you want help with this, this is a complicated method. If you want to, we can also do pair-programming on Discord or such.
> 
> Thank you! That was really helpful! I have plans today, so I probably won’t be able to continue working on this, but I’ll check it tomorrow!
> 
> I’d really love to do pair-programming, but I can’t speak any language other than Japanese… I’m currently studying English, though!

No worries, I speak fluent Japanese :) (ほんまだや、プログラミング専門用語は少し苦労するかもしれんが、できないのとではないです。)

yossydev:
> No worries, I speak fluent Japanese :) (ほんまだや、プログラミング専門用語は少し苦労するかもしれんが、できないのとではないです。)

Is that so!?
That’s amazing. I’d definitely love to do it when we both have time. I have plans today, so I can’t do it right now, though…

nova_vm/src/ecmascript/builtins/proxy.rs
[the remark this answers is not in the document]
aapoalas: Not exactly :)

Few problems: The function should not be locally here in the Proxy file but in the same place where the other create_list_from_array_like is (name of course cannot be exactly the same). It should not accept Integer at all. It should also accept SmallString. It should use a proper conversion function from strings to property keys (not an unsafe one and not a manual match-mapping).

Also it seems like you have two loops over the keys? That shouldn't be necessary at all, best as I can tell.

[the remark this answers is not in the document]
aapoalas: Oh and return value should be bound to the `GcScope<'a, '_>` lifetime.

aapoalas:
> > No worries, I speak fluent Japanese :) (ほんまだや、プログラミング専門用語は少し苦労するかもしれんが、できないのとではないです。)
> 
> Is that so!? That’s amazing. I’d definitely love to do it when we both have time. I have plans today, so I can’t do it right now, though…

I'm also entirely occupied on Mondays.

nova_vm/src/ecmascript/builtins/proxy.rs
yossydev: I wanted to use bind_property_keys, but I was getting an error with gc.reborrow. However, I found out that scope_property_keys exists. I was able to write it using that. Thank you!

640b78b8d66b8cd03e5afde70c21d1671fa18271

nova_vm/src/ecmascript/builtins/proxy.rs
yossydev: fixed by 74cc6282c93e4147a0656271fd78920fce636c14

nova_vm/src/ecmascript/builtins/proxy.rs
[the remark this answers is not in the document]
yossydev: Thank you. I don’t think the implementation of create_list_from_array_like is perfect yet, but for now, I was able to remove it with [this commit](https://github.com/trynova/nova/pull/522/commits/455aa6714b9b2608b90ce320de21c2ead10c4edd).

nova_vm/src/ecmascript/builtins/proxy.rs
yossydev: Thank you. I don’t think the implementation of create_list_from_array_like is perfect yet, but for now, I was able to remove it with [this commit](https://github.com/trynova/nova/pull/522/commits/455aa6714b9b2608b90ce320de21c2ead10c4edd).

nova_vm/src/ecmascript/builtins/proxy.rs
aapoalas: For reasons it is often necessary to first call unbind before calling bind.

nova_vm/src/ecmascript/builtins/proxy.rs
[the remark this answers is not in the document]
yossydev: > The function should not be locally here in the Proxy file but in the same place where the other create_list_from_array_like is (name of course cannot be exactly the same).

It seems it should have been placed in operations_on_objects.rs. I’ve moved it there.
To avoid naming conflicts, I renamed it to create_property_key_list_from_array_like.

> It should not accept Integer at all. It should also accept SmallString. It should use a proper conversion function from strings to property keys (not an unsafe one and not a manual match-mapping).
Also, it seems like you have two loops over the keys? That shouldn’t be necessary at all, best as I can tell.

Thank you! You’re right, Integer wasn’t necessary.
I’m thinking of editing it like this—what do you think?

```rs
pub(crate) fn create_property_key_list_from_array_like<'a>(
    agent: &mut Agent,
    obj: Value,
    mut gc: GcScope<'a, '_>,
) -> JsResult<Vec<PropertyKey<'static>>> {
    // 1. If validElementTypes is not present, set validElementTypes to all.
    // 2. If obj is not an Object, throw a TypeError exception.
    if !obj.is_object() {
            ExceptionType::TypeError,
            "Not an object",
            gc.nogc(),
        ));
    }
    // 3. Let len be ? LengthOfArrayLike(obj).
    let object = Object::try_from(obj).unwrap();
    let len = length_of_array_like(agent, object, gc.reborrow())?;
    let len = usize::try_from(len).unwrap();
    // 4. Let list be a new empty List.
    let mut list = Vec::with_capacity(len);
    // 5. Let index be 0.
    let mut index = 0;
    // 6. Repeat, while index < len,
    while index < len {
        // a. Let indexName be ! ToString(𝔽(index)).
        // b. Let next be ? Get(obj, indexName).
        let next = get(
            agent,
            object,
            PropertyKey::Integer(SmallInteger::try_from(index as u64).unwrap()),
            gc.reborrow(),
        )?;
        // c. If validElementTypes is property-key and next is not a property key, throw a TypeError exception.
        match next {
            // d. Append next to list. (Handled in match above)
            Value::String(s) => list.push(PropertyKey::String(s)),
            Value::SmallString(ss) => list.push(PropertyKey::SmallString(ss)),
            Value::Symbol(sym) => list.push(PropertyKey::Symbol(sym)),
            _ => {
                    ExceptionType::TypeError,
                    "proxy [[OwnPropertyKeys]] must return an array with only string and symbol elements",
                    gc.nogc(),
                ));
            }
        }
        // e. Set index to index + 1.
        index += 1;
    }
    // 7. Return list.
    Ok(list)
}
```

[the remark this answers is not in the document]
yossydev: I’ve pushed the changes.  acfe2604f0fce3afa2e16e4dc0fd89327c6deb08

> Oh and return value should be bound to the GcScope<'a, '_> lifetime.

When I ran `cargo clippy --fix`, it automatically fixed this.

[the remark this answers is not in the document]
aapoalas: Looks pretty good, you can use the `let Ok(object)` trick to avoid the unwrap on the object conversion.

The strings need to be properly converted to property keys using a helper method we have somewhere. JavaScript code has to return eg. `["0"]` and we'll want to convert that to a list containing `PropertyKey::Integer(0)`.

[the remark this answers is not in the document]
yossydev: > The strings need to be properly converted to property keys using a helper method we have somewhere. JavaScript code has to return eg. ["0"] and we'll want to convert that to a list containing PropertyKey::Integer(0).

I see, this is probably the [to_property_key](https://github.com/trynova/nova/blob/a358a0f12e2d64d06a201b15b893377814a844c1/nova_vm/src/ecmascript/abstract_operations/type_conversion.rs#L1162) abstract operation…?? 👀

> converts argument to a value that can be used as a property key
> ref: https://tc39.es/ecma262/#sec-topropertykey

The implementation seems to be doing something along those lines as well 👀

https://github.com/trynova/nova/blob/a358a0f12e2d64d06a201b15b893377814a844c1/nova_vm/src/ecmascript/abstract_operations/type_conversion.rs#L1219

nova_vm/src/ecmascript/abstract_operations/operations_on_objects.rs
```diff
@@ -1062,6 +1062,65 @@ pub(crate) fn create_list_from_array_like(
...
+            PropertyKey::Integer(SmallInteger::try_from(index as u64).unwrap()).unbind(),
+            gc.reborrow(),
+        )?;
+        match next {
+            Value::String(_) | Value::SmallString(_) => {
+                let string_value = String::try_from(next).unwrap();
+                let scoped_property_key =
+                    unwrap_try(to_property_key_simple(agent, string_value, gc.nogc()))
+                        .scope(agent, gc.nogc());
+                list.push(scoped_property_key.get(agent));
```
aapoalas, line 1108: **issue:** Let's just push the `scoped_property_key` into the list instead, and return a `Vec<Scoped<PropertyKey>>` from this method. That's (currently) the correct way of doing this in a GC-safe way.

nova_vm/src/ecmascript/abstract_operations/operations_on_objects.rs
```diff
@@ -1062,6 +1062,65 @@ pub(crate) fn create_list_from_array_like(
...
+            }
+            Value::Symbol(sym) => list.push(PropertyKey::Symbol(sym)),
```
aapoalas, line 1110: **note:** So here you'll also have to `.scope` the property key.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -1230,10 +1234,166 @@ impl InternalMethods for Proxy {
...
+            }
+        }
+        // 10. Let extensibleTarget be ? IsExtensible(target).
+        let extensible_target = is_extensible(agent, target, gc.reborrow())?;
+        // 11. Let targetKeys be ? target.[[OwnPropertyKeys]]().
+        let keys = target.internal_own_property_keys(agent, gc.reborrow())?;
+        let target_keys = scope_property_keys(agent, unbind_property_keys(keys), gc.nogc());
+        // 13. Assert: targetKeys contains no duplicate entries.
+        let mut seen = HashSet::new();
+        for i in &trap_result {
```
aapoalas, line 1311: **issue:** Spec says `targetKeys`, not `trapResult`.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -1230,10 +1234,166 @@ impl InternalMethods for Proxy {
...
+            gc.reborrow(),
+        )?;
+        // 8. Let trapResult be ? CreateListFromArrayLike(trapResultArray, property-key).
+        let trap_result = create_property_key_list_from_array_like(
+            agent,
+            trap_result_array.unbind(),
+            gc.reborrow(),
+        )?;
+        // 9. If trapResult contains any duplicate entries, throw a TypeError exception.
+        for (i, value) in trap_result.iter().enumerate() {
```
aapoalas, line 1292: **note:** If we return `Vec<Scoped<PropertyKey>>` from `create_property_key_list` then here we'll have to do `.get(agent)` at every turn here because `Scoped`'s are always unique, so you'd never find any duplicates here even if in reality they do exist.

nova_vm/src/ecmascript/builtins/proxy.rs
```diff
@@ -1230,10 +1234,166 @@ impl InternalMethods for Proxy {
...
+        }
+        // 10. Let extensibleTarget be ? IsExtensible(target).
+        let extensible_target = is_extensible(agent, target, gc.reborrow())?;
+        // 11. Let targetKeys be ? target.[[OwnPropertyKeys]]().
+        let keys = target.internal_own_property_keys(agent, gc.reborrow())?;
+        let target_keys = scope_property_keys(agent, unbind_property_keys(keys), gc.nogc());
+        // 13. Assert: targetKeys contains no duplicate entries.
+        let mut seen = HashSet::new();
+        for i in &trap_result {
+            assert!(seen.insert(i), "Duplicate entry found in trap_result");
```
aapoalas, line 1312: **note (nonblocking):** After `scope_property_keys`, all the `i`s in the `target_keys` vec are unique because `Scoped<T>` are always unique.

It doesn't actually matter in this case because the assert should never fail (though checking it correctly using `seen.insert(i.get(agent))` is still a good idea because we should check it anyway just to be sure we don't have a bug somewhere).

nova_vm/src/ecmascript/abstract_operations/operations_on_objects.rs
```diff
@@ -1062,6 +1062,65 @@ pub(crate) fn create_list_from_array_like(
...
+/// The abstract operation CreateListFromArrayLike takes argument obj (an ECMAScript language value)
+/// and optional argument elementTypes (a List of names of ECMAScript Language Types) and returns
+/// either a normal completion containing a List of ECMAScript language values or a throw
+/// completion. It is used to create a List value whose elements are provided by the indexed
+/// properties of obj. elementTypes contains the names of ECMAScript Language Types that are allowed
+/// for element values of the List that is created.
+pub(crate) fn create_property_key_list_from_array_like<'a>(
+    agent: &mut Agent,
+    obj: Value,
+    mut gc: GcScope,
```
aapoalas, line 1076: **issue:** You need `GcScope<'a, '_>` here.

nova_vm/src/ecmascript/builtins/proxy.rs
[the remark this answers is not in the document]
aapoalas: If possible, we could pair-program it through. The error is basically absolutely correct, but it's a probably easier to explain "face to face" and fix the issue together :)

aapoalas (changes requested):
Some issues still. If at all possible, I'd love to pair-program this through.

yossydev:
> Some issues still. If at all possible, I'd love to pair-program this through.

Thanks for the review!
Yes, I would love to have you!

aapoalas (approved):
LGTM <3 Thank you very much

PR #527, opened 2025-01-13
feat(ecmascript): Implement `JSON.stringify`

eliassjogreen:
I probably need to do some scoping/binding magic with some of my values still.


aapoalas (commented):
Some comments at least from a quick glance.

Making this GC-safe is going to be a pain with the tools that we have now. I _have to_ create a way to scope collections.

load1n9:
Should we make a move to rust nightly or figure out a way to not use `![feature(iter_intersperse)]`

aapoalas:
> Should we make a move to rust nightly or figure out a way to not use `![feature(iter_intersperse)]`

The latter: I don't want to pick up Nightly unless _absolutely_ necessary for some fundamental reason, like "we cannot continue the project (in the way we want to) if we don't do it"

load1n9:
hope this feature gets stabilized soon tbh

aapoalas:
I started working on this locally and have it about 60% done. The intersperse feature ends up being unnecessary.

PR #534, opened 2025-01-14
feat(gc): Sub-Object lifetimes

aapoalas:
Except Proxy, because it's actively being worked on and I expected it to be a bit harder than the rest.


aapoalas:
@eliassjogreen Noticed that we have missed one thing about TypedArrays: They need their custom internal methods implemented :)

eliassjogreen:
@aapoalas The `internal_get`/`internal_set` methods right?

aapoalas:
> @aapoalas The `internal_get`/`internal_set` methods right?

All that the spec defines, and maybe some others as well due to the elements / backing object split. So defineownproperty is probably needed as well even if the spec doesn't define it.

PR #539, opened 2025-01-18
feat(ecmascript): Callable

yossydev:
I have created Callable as a new implementation. I copied the contents of function.rs and added Proxy to each function.

Now, I’m thinking of rewriting is_callable as follows:

```rs
pub(crate) fn is_callable<'a, 'b>(
    argument: impl TryInto<Function<'b>>,
    gc: NoGcScope<'a, '_>,
) -> Option<Callable<'a>> {
    // 1. If argument is not an Object, return false.
    let Ok(argument) = argument.try_into() else {
        return None;
    };
    // 2. If argument has a [[Call]] internal method, return true.
    is_callable(argument, gc)
}
```

However, if I do this, it will cause compilation errors in all the places where is_callable is currently being used. Since the existing functions mostly accept Function, this behavior seems natural.

At this stage, does this implementation approach for Callable align with the intended design?

### DIFF
--- a/nova_vm/src/ecmascript/types/language/callable.rs
+++ b/nova_vm/src/ecmascript/types/language/callable.rs
@@ -253,7 +253,7 @@ impl<'a> InternalSlots<'a> for Callable<'a> {
             Callable::BuiltinPromiseResolvingFunction(d) => agent[d].object_index,
             Callable::BuiltinPromiseCollectorFunction => todo!(),
             Callable::BuiltinProxyRevokerFunction => todo!(),
-            Callable::Proxy(_) => todo!(),
+            Callable::Proxy(_) => None,
         }
     }
 
--- a/nova_vm/src/ecmascript/types/language/callable.rs
+++ b/nova_vm/src/ecmascript/types/language/callable.rs
@@ -253,7 +253,7 @@ impl<'a> InternalSlots<'a> for Callable<'a> {
             Callable::BuiltinPromiseResolvingFunction(d) => agent[d].object_index,
             Callable::BuiltinPromiseCollectorFunction => todo!(),
             Callable::BuiltinProxyRevokerFunction => todo!(),
-            Callable::Proxy(_) => None,
+            Callable::Proxy(_) => unreachable!(),
         }
     }
 
--- a/nova_vm/src/ecmascript/types/language/callable.rs
+++ b/nova_vm/src/ecmascript/types/language/callable.rs
@@ -54,7 +54,7 @@ impl std::fmt::Debug for Callable<'_> {
             }
             Callable::BuiltinPromiseCollectorFunction => todo!(),
             Callable::BuiltinProxyRevokerFunction => todo!(),
-            Callable::Proxy(d) => write!(f, "PROXY_DISCRIMINANT({:?})", d),
+            Callable::Proxy(d) => write!(f, "Proxy({:?})", d),
         }
     }
 }
--- a/nova_vm/src/ecmascript/types/language/callable.rs
+++ b/nova_vm/src/ecmascript/types/language/callable.rs
@@ -1,0 +1,755 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use super::{
+    function, value::{
+        BOUND_FUNCTION_DISCRIMINANT, BUILTIN_CONSTRUCTOR_FUNCTION_DISCRIMINANT,
+        BUILTIN_FUNCTION_DISCRIMINANT, BUILTIN_GENERATOR_FUNCTION_DISCRIMINANT,
+        BUILTIN_PROMISE_COLLECTOR_FUNCTION_DISCRIMINANT,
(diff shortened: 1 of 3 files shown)


nova_vm/src/ecmascript/types/language/callable.rs
```diff
@@ -0,0 +1,755 @@
...
+            Callable::BuiltinGeneratorFunction => todo!(),
+            Callable::BuiltinConstructorFunction(d) => {
+                write!(f, "BuiltinConstructorFunction({:?})", d)
+            }
+            Callable::BuiltinPromiseResolvingFunction(d) => {
+                write!(f, "BuiltinPromiseResolvingFunction({:?})", d)
+            }
+            Callable::BuiltinPromiseCollectorFunction => todo!(),
+            Callable::BuiltinProxyRevokerFunction => todo!(),
+            Callable::Proxy(d) => write!(f, "PROXY_DISCRIMINANT({:?})", d),
```
aapoalas, line 57: **nitpick:** Should be `Proxy({:?)`.

nova_vm/src/ecmascript/types/language/callable.rs
```diff
@@ -0,0 +1,755 @@
...
+        match self {
+            Callable::BoundFunction(d) => agent[d].object_index,
+            Callable::BuiltinFunction(d) => agent[d].object_index,
+            Callable::ECMAScriptFunction(d) => agent[d].object_index,
+            Callable::BuiltinGeneratorFunction => todo!(),
+            Callable::BuiltinConstructorFunction(d) => agent[d].object_index,
+            Callable::BuiltinPromiseResolvingFunction(d) => agent[d].object_index,
+            Callable::BuiltinPromiseCollectorFunction => todo!(),
+            Callable::BuiltinProxyRevokerFunction => todo!(),
+            Callable::Proxy(_) => todo!(),
```
aapoalas, line 256: **nitpick:** This should be unreachable as well.

aapoalas (commented):
This is exactly what I was thinking, yes.

... But now that I see it and read the spec a little more, it might be that we should just rename `Function` to `Callable` and add `Proxy` there. The spec says this about callable objects:

> an object is a member of the built-in type Object; and a function is a callable object.

and

> In addition, some objects are callable; these are referred to as functions or [function objects](https://tc39.es/ecma262/#function-object)

ie. The spec is saying that "whenever I say 'function', what I mean is any kind of actual function or a callable Proxy". And this is validated by how the spec is written. So the `Function` enum doesn't actually correspond to anything in the spec... It _does_ make sense from a human standpoint but it might not make actual sense from the code standpoint...

`is_callable` definitely should return `Callable`, and basically every place where we have `Function` as a parameter or saved in a struct should indeed have `Callable`... The only question is if there will remain any places whatsoever where the `Function` enum would remain at all?

nova_vm/src/ecmascript/types/language/callable.rs
yossydev: thank you! fixed to 2e8bba337e1d6e739f1d16309dcb3e5a03722c00

nova_vm/src/ecmascript/types/language/callable.rs
yossydev: thank you! return None 91f9ae7317c6ab7ac198ba9f31f8a8fe709af4f7

aapoalas: Just call `unreachable!()`

yossydev:
> 	… But now that I see it and read the spec a little more, it might be that we should just rename Function to Callable and add Proxy there. The spec says this about callable objects:

Yeah, I was also wondering if that might be the right approach while trying to refactor this. I was a bit unsure about when we would actually need to treat Function and Callable separately.

Should we go ahead and rewrite function.rs as callable.rs, or should we merge this PR and gradually replace the places using Function with Callable? If we decide to go with the former, I think we can go ahead and close this PR. What do you think? 👀

nova_vm/src/ecmascript/types/language/callable.rs
yossydev: Got it, I’ve made the changes! d491dbb81bc7a45a0b11963b8e2163fc74b9c46c

aapoalas:
> > … But now that I see it and read the spec a little more, it might be that we should just rename Function to Callable and add Proxy there. The spec says this about callable objects:
> 
> Yeah, I was also wondering if that might be the right approach while trying to refactor this. I was a bit unsure about when we would actually need to treat Function and Callable separately.
> 
> Should we go ahead and rewrite function.rs as callable.rs, or should we merge this PR and gradually replace the places using Function with Callable? If we decide to go with the former, I think we can go ahead and close this PR. What do you think? 👀

I'm a bit unsure. If we add Proxy to Function and rename it to Callable, then we'll get basically all we want the easy way. But! It might still be that there's a place somewhere where the spec does actually rule out callable Proxies, and if that is the case then we'll have removed the enum that would've been used to represent that. The worst thing is that we won't see that case in our code if we just rename Function to Callable.

But, if we do go ahead with a separate Callable then we're going to do a lot of work that might end up being entirely unnecessary :/

aapoalas:
A quick search through the spec suggests that Proxies are never categorically separated from callable objects, so a closing this PR and just renaming Function to Callable seems proper.

PR #541, opened 2025-01-18
feat(ecmascript): aggregate error construtor

yossydev:
I implemented it based on the current NativeError and Error.

ref: https://github.com/trynova/nova/issues/133
doc: https://tc39.es/ecma262/#sec-aggregate-error-objects


aapoalas (approved):
LGTM! There's a deeper ... not really an issue but a complication here: We have `ErrorHeapData` for storing errors. It's ... kind of like an optimised heap format for errors, created for the purpose of avoiding one extra indirection for the commonly used name and (especially) message property.

That `ErrorHeapData` does not have the `errors` property field prepared (and we likely don't want to prepare it either, since most errors objects don't have that property), so each aggregate error will necessarily allocate a backing object. That's not a great thing. But for now it's also not a horrible thing, we can live with it.

I'm kind of beginning to think that we might want to just turn the Error type into an `Error(OrdinaryObject)` instead of a separate `Error(Error)`.

PR #542, opened 2025-01-19
chore(test262): Update test262 submodule

yossydev:
The last update was in October(https://github.com/trynova/nova/pull/439), so I tried updating it locally, and it seems there were quite a few changes. Most of them were related to Temporal. It looks like it would be quite challenging to implement.


aapoalas (changes requested):
The tests that failed on CI should be looked at: One of them crashes when it is expected to timeout. That sounds like a `debug_assert` triggering when built with `cargo build` but with `cargo build --profile release` it enters an eternal loop. Maybe.

Tests where we're just too slow for a mostly good reason can be added to the skip list in `skip.json`.

yossydev:
> The tests that failed on CI should be looked at: One of them crashes when it is expected to timeout. That sounds like a `debug_assert` triggering when built with `cargo build` but with `cargo build --profile release` it enters an eternal loop. Maybe.
> 
> Tests where we're just too slow for a mostly good reason can be added to the skip list in `skip.json`.

Thank you for the review!
I see, so the difference between debug and release modes was the presence or absence of `debug_assert` triggers 👀

For now, I’ve added the other timeout cases to skip.json(3ccf4f316ba00597267dd0fefa6b5fd253a7013a), but the two failing cases in CI don’t result in errors on my machine.
I’m using the same command as CI:
`cargo b && cargo run --bin test262 -- --noprogress`
Could this be due to a difference in specs between my local environment and the CI environment? 🧐

aapoalas:
> > The tests that failed on CI should be looked at: One of them crashes when it is expected to timeout. That sounds like a `debug_assert` triggering when built with `cargo build` but with `cargo build --profile release` it enters an eternal loop. Maybe.
> > Tests where we're just too slow for a mostly good reason can be added to the skip list in `skip.json`.
> 
> Thank you for the review! I see, so the difference between debug and release modes was the presence or absence of `debug_assert` triggers 👀
> 
> For now, I’ve added the other timeout cases to skip.json([3ccf4f3](https://github.com/trynova/nova/commit/3ccf4f316ba00597267dd0fefa6b5fd253a7013a)), but the two failing cases in CI don’t result in errors on my machine. I’m using the same command as CI: `cargo b && cargo run --bin test262 -- --noprogress` Could this be due to a difference in specs between my local environment and the CI environment? 🧐

CI at least currently runs the tests using debug, just to be sure we're not hitting any of the debug asserts. Locally we often don't do that as it also makes to code massively slower. You can remove `--profile release` from your commands to (usually) get to what CI does

aapoalas:
Oh, you're already running the same command... Hmm, we _shouldn't_ have any environmental dependencies. The only real difference is or should be that the CI is generally slower.

yossydev:
@aapoalas 

> Oh, you're already running the same command... Hmm, we _shouldn't_ have any environmental dependencies. The only real difference is or should be that the CI is generally slower.

I understand why it times out because it’s slow, but I don’t quite understand why it crashes…
Since it’s in staging, I think it’s okay to add the following two failing tests to skip.json. What do you think?

- staging/sm/Array/length-truncate-nonconfigurable-sparse.js
- staging/sm/Array/length-truncate-with-indexed.js

Since it’s a crash, which isn’t ideal, I also think we should create an issue for it… 👀

aapoalas (approved):
Thank you so much <3

Issue #140, opened 2024-04-15
String constuctor builtin functions


aapoalas:

aapoalas:
> This can be closed now with [#702](https://github.com/trynova/nova/pull/702) being merged.

Thanks <3

PR #543, opened 2025-01-19
feat(ecmascript): String fromCodePoints

yossydev:
ref: https://github.com/trynova/nova/issues/140
doc: https://tc39.es/ecma262/multipage/text-processing.html#sec-string.fromcodepoint


nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +165,48 @@ impl StringConstructor {
...
-        //     a. Let nextCP be ? ToNumber(next).
-        //     b. If IsIntegralNumber(nextCP) is false, throw a RangeError exception.
-        //     c. If ℝ(nextCP) < 0 or ℝ(nextCP) > 0x10FFFF, throw a RangeError exception.
-        //     d. Set result to the string-concatenation of result and UTF16EncodeCodePoint(ℝ(nextCP)).
+        let mut result = std::string::String::new();
         // 3. Assert: If codePoints is empty, then result is the empty String.
+        if code_points.is_empty() {
+            return Ok(String::EMPTY_STRING.into_value());
+        }
+        // 2. For each element next of codePoints, do
```
aapoalas, line 184: **suggestion:** We could add a fast path for if we've been given a single integer code point. That's likely going to be the common use, and on that fast path we can use `https://doc.rust-lang.org/std/primitive.char.html#method.encode_utf8` to write the value directly into a `SmallString` without doing any heap allocation.

yossydev: I added it here! https://github.com/trynova/nova/pull/543/commits/2cbd76a2b5f1f655400bf4b9bae7350aa58793d8

Ah, I see! encode_utf8 was already being used here:
https://github.com/trynova/nova/blob/b9681f2bcd1be10866a1d1bfe24a7208fa6f6cd3/small_string/lib.rs#L183-L187

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +165,75 @@ impl StringConstructor {
...
+            // b. If IsIntegralNumber(nextCP) is false, throw a RangeError exception.
+            if !is_integral_number(agent, next_cp, gc.reborrow()) {
+                return Err(agent.throw_exception(
+                    ExceptionType::RangeError,
+                    format!("{:?} is not a valid code point", next_cp.to_real(agent)),
+                    gc.nogc(),
+                ));
+            }
+            // c. If ℝ(nextCP) < 0 or ℝ(nextCP) > 0x10FFFF, throw a RangeError exception.
+            let next_cp = to_uint32(agent, next_cp.into_value(), gc.reborrow())?;
```
aapoalas, line 224: **nitpick:** Use `to_uint32_number`.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +165,75 @@ impl StringConstructor {
...
-        _gc: GcScope,
+        code_points: ArgumentsList,
+        mut gc: GcScope,
     ) -> JsResult<Value> {
+        // 3. Assert: If codePoints is empty, then result is the empty String.
+        if code_points.is_empty() {
+            return Ok(String::EMPTY_STRING.into_value());
+        }
+        // fast path: only a single valid code unit
+        if code_points.len() == 1 {
```
aapoalas, line 183: **nitpick:** I'd just do `&& code_points.first().unwrap().is_integer()` so we eg. don't accept `String.fromCodePoint("13")` here but only `String.fromCodePoint(13)`. That way we don't need to do the `to_number` or `is_integral_number` checks here.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +165,75 @@ impl StringConstructor {
...
+            // b. If IsIntegralNumber(nextCP) is false, throw a RangeError exception.
+            if !is_integral_number(agent, next_cp, gc.reborrow()) {
+                return Err(agent.throw_exception(
+                    ExceptionType::RangeError,
+                    format!("{:?} is not a valid code point", next_cp.to_real(agent)),
+                    gc.nogc(),
+                ));
+            }
+            // c. If ℝ(nextCP) < 0 or ℝ(nextCP) > 0x10FFFF, throw a RangeError exception.
+            let next_cp = to_uint32(agent, next_cp.into_value(), gc.reborrow())?;
```
aapoalas, line 195: **issue:** Spec does not ask to call `to_uint32`; the `R(V)` is `v.to_real(agent)` for us, although since we know the value must be an integer we can just use `into_i64` instead (I think).

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
aapoalas, line 224: **issue:** Here too, don't use `to_uint32` since it's not requested.

aapoalas (changes requested):
Found some differences compared to what the spec expects.

yossydev:
@aapoalas 
Thank you for the review! I really appreciate it!
I’m not sure if my understanding is correct, but I made the fix with https://github.com/trynova/nova/pull/543/commits/bc469217fd07f92c028bc4c32cf0638e1629b876

I have two things I’d like to confirm.

First, in the fast path, I’m using into_i64, but in the non-fast path, I’m using to_uint32_number.
Did you originally intend for into_i64 to be used in the latter case as well? I initially tried using into_i64, but it caused an error in tests/test262/test/built-ins/String/fromCodePoint/number-is-out-of-range.js when -1 was included, so I decided to use to_uint32_number instead.

Second, in the fast path, if I remove to_number, I can’t use into_i64, so I left it as is.
However, is_integral_number seemed fine to remove, so I went ahead and removed it!

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +165,68 @@ impl StringConstructor {
...
+        mut gc: GcScope,
     ) -> JsResult<Value> {
+        // 3. Assert: If codePoints is empty, then result is the empty String.
+        if code_points.is_empty() {
+            return Ok(String::EMPTY_STRING.into_value());
+        }
+        // fast path: only a single valid code unit
+        if code_points.len() == 1 && code_points.first().unwrap().is_integer() {
+            // a. Let nextCP be ? ToNumber(next).
+            let next_cp = to_number(agent, code_points[0].into_value(), gc.reborrow())?.unbind();
```
aapoalas, line 185:
```suggestion
            let Value::Integer(next_cp) = code_points.first().unwrap() else {
                unreachable!();
            };
            let next_cp = next_cp.into_i64();
```

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +165,68 @@ impl StringConstructor {
...
+            // b. If IsIntegralNumber(nextCP) is false, throw a RangeError exception.
+            // c. If ℝ(nextCP) < 0 or ℝ(nextCP) > 0x10FFFF, throw a RangeError exception.
+            let next_cp = next_cp.into_i64(agent);
```
aapoalas, line 188:
```suggestion
```

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +165,68 @@ impl StringConstructor {
...
+            if next_cp > 0x10FFFF {
```
aapoalas, line 189: **issue:** You also need a lower bound check. i64 can be negative.

```suggestion
            if next_cp < 0 || next_cp > 0x10FFFF {
```

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +165,68 @@ impl StringConstructor {
...
+            // b. If IsIntegralNumber(nextCP) is false, throw a RangeError exception.
+            if !is_integral_number(agent, next_cp, gc.reborrow()) {
+                return Err(agent.throw_exception(
+                    ExceptionType::RangeError,
+                    format!("{:?} is not a valid code point", next_cp.to_real(agent)),
+                    gc.nogc(),
+                ));
+            }
+            // c. If ℝ(nextCP) < 0 or ℝ(nextCP) > 0x10FFFF, throw a RangeError exception.
+            let next_cp = to_uint32_number(agent, next_cp);
```
aapoalas, line 217: **issue:** As mentioned, the spec doesn't say to use `ToUint32` so we shouldn't use it. Here's what we should do, I believe:

```suggestion
            let next_cp = next_cp.into_i64(agent);
```

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +165,68 @@ impl StringConstructor {
...
+            if next_cp > 0x10FFFF {
```
aapoalas, line 218: (continuing the above comment) And then this should fix the issue you had with -1:

```suggestion
            if next_cp < 0 || next_cp > 0x10FFFF {
```

aapoalas (changes requested):
Still a few problems, sorry :( But after these I think this should be good and very solid <3

yossydev:
> Still a few problems, sorry :( But after these I think this should be good and very solid <3

@aapoalas 
No, thank you for reviewing it multiple times! I really appreciate it!!
I’ve fixed it here: [8e63517ac746032dc4e1692cd2ba2a132b1fa49b](https://github.com/trynova/nova/pull/543/commits/8e63517ac746032dc4e1692cd2ba2a132b1fa49b)!!

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +162,66 @@ impl StringConstructor {
...
         // 1. Let result be the empty String.
+        let mut result = std::string::String::new();
         // 2. For each element next of codePoints, do
-        //     a. Let nextCP be ? ToNumber(next).
-        //     b. If IsIntegralNumber(nextCP) is false, throw a RangeError exception.
-        //     c. If ℝ(nextCP) < 0 or ℝ(nextCP) > 0x10FFFF, throw a RangeError exception.
-        //     d. Set result to the string-concatenation of result and UTF16EncodeCodePoint(ℝ(nextCP)).
-        // 3. Assert: If codePoints is empty, then result is the empty String.
+        for next in code_points.iter() {
+            // a. Let nextCP be ? ToNumber(next).
```
aapoalas, line 206: **issue:** You shouldn't remove the `ToNumber` from this general path.
```suggestion
            // a. Let nextCP be ? ToNumber(next).
            let next_cp = to_number(agent, next, gc.reborrow())?;
```

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +162,66 @@ impl StringConstructor {
...
+            // b. If IsIntegralNumber(nextCP) is false, throw a RangeError exception.
+            // c. If ℝ(nextCP) < 0 or ℝ(nextCP) > 0x10FFFF, throw a RangeError exception.
+            let Value::Integer(next_cp) = next else {
```
aapoalas, line 209: **issue:** We don't know that `next_cp` is an integer here. You need to perform the `IsIntegralNumber` check.
```suggestion
            if !is_integral_number(agent, next_cp) {
```

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_constructor.rs
```diff
@@ -161,26 +162,66 @@ impl StringConstructor {
...
+                unreachable!();
```
aapoalas, line 210:
```suggestion
                // throw error here
```

aapoalas (changes requested):
The generic path shouldn't assume `Value::Integer`.

aapoalas (approved):
LGTM, thank you! Had one minor nitpick which has _no_ effect in the end :)

aapoalas:
BTW you seem to have updated the test262 submodule at some point. Or maybe I did that?

Anyway, it's not the _best_ thing to do it in this PR but it's fine, let's keep it :)

yossydev:
> BTW you seem to have updated the test262 submodule at some point. Or maybe I did that?
> 
> Anyway, it's not the _best_ thing to do it in this PR but it's fine, let's keep it :)

Thank you for merging! And sorry, I didn’t mean to update it, but it looks like it got included in the changes! 🙇‍♂️

PR #544, opened 2025-01-19
feat(ecmascript): Implement Annex-B string prototype methods

eliassjogreen:
I also added some feature flags for including parts of or the whole of annex-b. Made them on-by-default, but might make more sense for them to be off?


nova_vm/Cargo.toml
```diff
@@ -52,6 +53,18 @@ regexp = []
 shared-array-buffer = []
 weak-refs = []
 typescript = []
+# Enables features defined by [Annex B](https://tc39.es/ecma262/#sec-additional-ecmascript-features-for-web-browsers)
```
aapoalas, line 56: **praise:** Excellent stuff! This is exactly what I was thinking we should probably do; split the annex B parts into more component features. Excellent work with the documentation as well <3

eliassjogreen:
There are no expectations updated because we have turned off Annex B in the test runner, I ran it locally and it passed the string expectations, but I don't suppose we want to turn it on for the whole engine.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -259,6 +259,154 @@ impl Builtin for StringPrototypeIterator {
...
+impl Builtin for StringPrototypeSup {
+    const NAME: String<'static> = BUILTIN_STRING_MEMORY.sup;
+    const LENGTH: u8 = 0;
+    const BEHAVIOUR: Behaviour = Behaviour::Regular(StringPrototype::sup);
+}
+
+#[cfg(feature = "annex-b-string")]
+struct StringPrototypeTrimLeft;
+#[cfg(feature = "annex-b-string")]
+impl Builtin for StringPrototypeTrimLeft {
```
aapoalas, line 391: **nitpick (blocking, but not very strongly):** The spec's terminology for TrimLeft and TrimRight is

> The initial value of the "trimLeft" property is %String.prototype.trimStart%, defined in [22.1.3.34](https://tc39.es/ecma262/#sec-string.prototype.trimstart).

ie. it's not just another function that does the same thing but it's actually _the same function object_. You should `impl BuiltinIntrinsic for StringPrototypeTrimStart` and make `StringPrototypeTrimLeft` a reference to that. See f.ex. how `ArrayPrototypeValues` and how we use `with_property` to generate `array.prototype[Symbol.Iterator]`.

nova_vm/src/ecmascript/builtins/text_processing/string_objects/string_prototype.rs
```diff
@@ -2028,6 +2419,69 @@ fn this_string_value<'gc>(
...
 }
 
+/// ### [B.2.2.2.1 CreateHTML ( string, tag, attribute, value )](https://tc39.es/ecma262/#sec-createhtml)
+///
+/// The abstract operation CreateHTML takes arguments string (an ECMAScript
+/// language value), tag (a String), attribute (a String), and value (an
+/// ECMAScript language value) and returns either a normal completion
+/// containing a String or a throw completion.
+#[cfg(feature = "annex-b-string")]
+fn create_html<'gc>(
```
aapoalas, line 2429: **praise:** This is nice <3

aapoalas (changes requested):
Looks mostly good to me, but let's change the function identities to match what the spec expects since it should be fairly easy to do.

We could also add the feature flag to the tests build and (eventually at least) turn on Annex B testing, but indeed let's not turn it on by default for the cli build. Don't want that whole mess.

We could also make PR to https://github.com/test262-fyi/test262.fyi/blob/main/scripts/engines/nova.sh to run with the Annex B feature flags.

aapoalas (approved):
LGTM! Nice work! <3

Issue #145, opened 2024-04-15
%TypedArray%.prototype builtin functions


aapoalas:

aapoalas:
I only now realised that @yossydev you indeed finished the TypedArray prototype methods! Awesome, and thank you! <3

I'll be adding SharedArrayBuffer support here so I won't close the issue quite yet, though.

aapoalas:
With #869 this is now fully done! <3

PR #545, opened 2025-01-26
feat(ecmascript): %TypedArray%.prototype.some

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.some


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1055,13 +1057,106 @@ impl TypedArrayPrototype {
...
-        _this_value: Value,
-        _: ArgumentsList,
-        _gc: GcScope,
+        agent: &mut Agent,
+        this_value: Value,
+        arguments: ArgumentsList,
+        mut gc: GcScope,
     ) -> JsResult<Value> {
-        todo!()
+        let callback = arguments.get(0);
```
aapoalas, line 1067: **suggestion:** Call `bind` on these args already; right now it does nothing but it will make this function automatically light up when I do implement the `Value` lifetime :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1055,13 +1057,106 @@ impl TypedArrayPrototype {
...
+        // 1. Let O be the this value.
+        let o = this_value;
+        // 2. Let taRecord be ? ValidateTypedArray(O, seq-cst).
+        let ta_record = validate_typed_array(agent, o, Ordering::SeqCst, gc.nogc())?;
+        let o = TypedArray::try_from(o).unwrap();
+        // 3. Let len be TypedArrayLength(taRecord).
+        let (len, _) = match o {
+            TypedArray::Int8Array(_) => (
+                typed_array_length::<i8>(agent, &ta_record, gc.nogc()),
+                std::mem::size_of::<i8>(),
```
aapoalas, line 1078: **question:** Why the `size_of` calls when their value is getting ignored in the return?

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1055,13 +1057,106 @@ impl TypedArrayPrototype {
...
+        let this_arg = arguments.get(1);
+        // 1. Let O be the this value.
+        let o = this_value;
+        // 2. Let taRecord be ? ValidateTypedArray(O, seq-cst).
+        let ta_record = validate_typed_array(agent, o, Ordering::SeqCst, gc.nogc())?;
+        let o = TypedArray::try_from(o).unwrap();
+        // 3. Let len be TypedArrayLength(taRecord).
+        let (len, _) = match o {
+            TypedArray::Int8Array(_) => (
+                typed_array_length::<i8>(agent, &ta_record, gc.nogc()),
```
aapoalas, line 1077: **suggestion:** We've not really been consistent with this, but you could combine `i8 | u8 | U8Clamed` into one branch, same for 16 bit values, 32 bit values and 64 bit values: The TypedArrayLength will give the same result regardless of `i8` / `u8` / `U8Clamed` and the others since it only depends on the size of the type.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1055,13 +1057,106 @@ impl TypedArrayPrototype {
...
+        };
+        let callback = callback.bind(gc.nogc()).scope(agent, gc.nogc());
+        // 5. Let k be 0.
+        let mut k = 0;
+        // 6. Repeat, while k < len,
+        while k < len {
+            // a. Let Pk be ! ToString(𝔽(k)).
+            let pk = PropertyKey::from(SmallInteger::from(k as u32));
+            // b. Let kValue be ! Get(O, Pk).
+            let k_value = get(agent, o, pk, gc.reborrow())?;
```
aapoalas, line 1137: **issue:** Spec says `! Get` meaning "this API cannot throw" and in this case it also means "this API cannot call JavaScript" (since we're dealing with accessing an index from a TypedArray). So, instead of `?` you should use `unwrap()`. And because this cannot call into JavaScript, you should use `try_get` here and use `unwrap_try` since there is 0 chance of this calling into JS.

aapoalas (changes requested):
Some nitpicks / suggestions and one minor issue (which doesn't affect correctness, only performance). In general looks great to me! Thank you <3

By the way, I've been busy preparing for FOSDEM and also took a bit of a city vacation this weekend, hence the relative "radio silence" during this week :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Thank you! I made it use bind! https://github.com/trynova/nova/pull/545/commits/d7f0a4fe77c2afa80089178f2d7de3892e86ec73

aapoalas (approved):
LGTM! Had one nitpick regarding the changes though :) I'll merge this at some point.

yossydev:
> Some nitpicks / suggestions and one minor issue (which doesn't affect correctness, only performance). In general looks great to me! Thank you <3
> 
> By the way, I've been busy preparing for FOSDEM and also took a bit of a city vacation this weekend, hence the relative "radio silence" during this week :)

Sounds great! I’m looking forward to your presentation!
I’ve been busy with work lately too, so I haven’t had much time… I wish I could do more, though…

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1055,13 +1057,79 @@ impl TypedArrayPrototype {
...
+            | TypedArray::Uint8ClampedArray(_) => {
+                typed_array_length::<u8>(agent, &ta_record, gc.nogc())
+            }
+            TypedArray::Int16Array(_) | TypedArray::Uint16Array(_) => {
+                typed_array_length::<u16>(agent, &ta_record, gc.nogc())
+            }
+            TypedArray::Int32Array(_) | TypedArray::Uint32Array(_) => {
+                typed_array_length::<u32>(agent, &ta_record, gc.nogc())
+            }
+            TypedArray::BigInt64Array(_) => typed_array_length::<i64>(agent, &ta_record, gc.nogc()),
```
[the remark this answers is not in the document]
yossydev: Thank you again!
I wonder if this is correct…? https://github.com/trynova/nova/pull/545/commits/e92c14285fb6c77296aca3c7ed36ebbd7b2eb087

PR #547, opened 2025-01-29
feat(ecmascript): %TypedArray%.prototype.every

yossydev:
It’s almost the same as some! https://github.com/trynova/nova/pull/545

doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.every


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -527,13 +527,80 @@ impl TypedArrayPrototype {
...
     ) -> JsResult<Value> {
-        todo!()
+        let callback = arguments.get(0).bind(gc.nogc());
+        let this_arg = arguments.get(1).bind(gc.nogc());
+        // 1. Let O be the this value.
+        let o = this_value;
+        // 2. Let taRecord be ? ValidateTypedArray(O, seq-cst).
+        let ta_record = validate_typed_array(agent, o, Ordering::SeqCst, gc.nogc())?;
+        // 3. Let len be TypedArrayLength(taRecord).
+        let o = TypedArray::try_from(o).unwrap();
```
aapoalas, line 544: **issue:** Should call `.bind(gc.nogc())` here. That will make borrow checker unhappy down below in the loop and o will require scoping with that.

Probably the same issue slipped in for `some`: We could fix that on the side here.

aapoalas (changes requested):
One GC-related problem popped up.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Thank you! I also fixed some along with it! https://github.com/trynova/nova/pull/547/commits/c60316968f1f9736b3d3398dd1927350cf54b67a

aapoalas (approved):
LGTM, nice work <3 Thank you

PR #548, opened 2025-01-30
fix(ecmascript): `trimLeft` and `trimRight` property descriptors

eliassjogreen:
Ran test262 and found some issues in the annex-b tests. Related #549

PR #549, opened 2025-01-30
fix(ecmascript): `anchor` function length

eliassjogreen:
Ran test262 and found some issues in the annex-b tests. Related #548


aapoalas (approved):
LGTM<3

PR #550, opened 2025-01-30
feat(ecmascript): %TypedArray%.prototype.at

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.at


aapoalas (dismissed):
LGTM! Nice work <3

PR #551, opened 2025-01-30
feat(ecmascript): Implement `Float16Array` proposal

eliassjogreen:
Currently `f16` in rust requires nightly, so I'll leave the feature flag for proposals off for now.


aapoalas (changes requested):
LGTM but need to resolve some clippy.

aapoalas (approved):
<3

PR #552, opened 2025-01-30
feat(ecmascript): Implement `Math.sumPrecise` proposal

eliassjogreen:
Oops my git history got weird. Based it on #551 so let's wait for that to be merged first.


aapoalas (dismissed):
LGTM! <3

andreubotella:
I don't think this implementation is correct.

Remember that mathematical values in the spec are not `f64`, they're arbitrary-precision numbers; and that any time you do a numeric operation in `f64`, you might be losing precision. The point of `Math.sumPrecise` is to use some of the existing algorithms that use accumulators to store this lost precision. See https://gitlab.com/radfordneal/xsum and https://gitlab.com/radfordneal/xsum/-/blob/master/xsum-paper.pdf, for example.

PR #554, opened 2025-02-01
feat(ecmascript): %TypedArray%.prototype.includes

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.includes


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -721,13 +724,103 @@ impl TypedArrayPrototype {
...
+            return Ok(false.into());
+        };
+        // 5. Let n be ? ToIntegerOrInfinity(fromIndex).
+        let n = if let TryResult::Continue(n) =
+            try_to_integer_or_infinity(agent, from_index, gc.nogc())
+        {
+            n?
+        } else {
+            to_integer_or_infinity(agent, from_index, gc.reborrow())?
+        };
```
yossydev, line 777: Objects ({}) and Symbols (Symbol) will enter this case. However, since the input is expected to be an integer in most cases, using try should be fine. What do you think?

aapoalas: Using the `try` method only really makes sense if you can use it to avoid scoping some or all values. In this case because we need `o` to be scoped for the loop, it doesn't really help any.

So the "try and branch if failed" work is basically not really useful.

aapoalas (changes requested):
Looks good to me, but the try branch doesn't really help here so let's remove that.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -721,13 +724,103 @@ impl TypedArrayPrototype {
...
     ) -> JsResult<Value> {
-        todo!()
+        let search_element = arguments.get(0).bind(gc.nogc());
+        let from_index = arguments.get(1).bind(gc.nogc());
+        // 1. Let O be the this value.
+        let o = this_value;
+        // 2. Let taRecord be ? ValidateTypedArray(O, seq-cst).
+        let ta_record = validate_typed_array(agent, o, Ordering::SeqCst, gc.nogc())?;
+        // 3. Let len be TypedArrayLength(taRecord).
+        let o = TypedArray::try_from(o)
```
aapoalas, line 742: **nitpick:** `let o = ta_record.object`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Thank you, I see. I didn’t fully understand it.
I’ve removed try! 🙇‍♂️

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: thank you! I fixed! https://github.com/trynova/nova/pull/554/commits/1ab31dd1eaa34bbcbe9ac19370452dd4036bc22b

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
aapoalas: Agh! I was wrong! I was somehow thinking this is something like `find` or `forEach` where we always call user code in the loop, but of course not! We're just checking equality!

So your try-branch absolutely made sense, the only thing missing was moving the scoping to be done there on a conditional basis.

I'll push a fix to this branch so you don't have to go back-and-forth with that yourself.

PR #556, opened 2025-02-02
feat(ecmascript): %TypedArray%.prototype.indexOf

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.indexof


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -731,13 +734,113 @@ impl TypedArrayPrototype {
...
+        if len == 0 {
+            return Ok((-1).into());
+        };
+        // 5. Let n be ? ToIntegerOrInfinity(fromIndex).
+        let n = if let TryResult::Continue(n) =
+            try_to_integer_or_infinity(agent, from_index, gc.nogc())
+        {
+            n?
+        } else {
+            to_integer_or_infinity(agent, from_index, gc.reborrow())?
```
aapoalas, line 786: **issue:** Here you can perform scoping of `o` only conditionally in this else-branch and it actually makes sense since the loop in this function will never trigger GC and we thus don't need scoping of `o` for that.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -731,13 +734,113 @@ impl TypedArrayPrototype {
...
     ) -> JsResult<Value> {
-        todo!()
+        let search_element = arguments.get(0).bind(gc.nogc());
+        let from_index = arguments.get(1).bind(gc.nogc());
+        // 1. Let O be the this value.
+        let o = this_value;
+        // 2. Let taRecord be ? ValidateTypedArray(O, seq-cst).
+        let ta_record = validate_typed_array(agent, o, Ordering::SeqCst, gc.nogc())?;
+        // 3. Let len be TypedArrayLength(taRecord).
+        let o = TypedArray::try_from(o)
```
aapoalas, line 752: **nitpick:** `let o = ta_record.object;` gets rid of the unwrap path.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -731,13 +734,113 @@ impl TypedArrayPrototype {
...
+            if k < 0 {
+                0
+            } else {
+                k
+            }
+        };
+        // 11. Repeat, while k < len,
+        while k < len {
+            // a. Let kPresent be ! HasProperty(O, ! ToString(𝔽(k))).
+            let k_present = has_property(
```
aapoalas, line 819: **issue:** This should be `unwrap_try(try_has_property(...))` as the `! HasProperty` gives us a guarantee that this cannot trigger user code, ie. will never return `Break`, and also thus does not need to prepare for GC.

Note, you can press the `U` key over in https://tc39.es/ecma262 to see where user code can run.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -731,13 +734,113 @@ impl TypedArrayPrototype {
...
+        // 5. Let n be ? ToIntegerOrInfinity(fromIndex).
+        let n = if let TryResult::Continue(n) =
+            try_to_integer_or_infinity(agent, from_index, gc.nogc())
+        {
+            n?
+        } else {
+            to_integer_or_infinity(agent, from_index, gc.reborrow())?
+        };
+        // 6. Assert: If fromIndex is undefined, then n is 0.
+        if from_index.is_undefined() {
```
aapoalas, line 789: **note:** At this point, no further user code can be triggered and we're safe to perform all optimisations we want: One thing we can check is if the `search_element` is a number and if it is not, we can return `-1` immediately without searching for the value at all.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -731,13 +734,113 @@ impl TypedArrayPrototype {
...
+            let k = len + n;
+            // b. If k < 0, set k to 0.
+            if k < 0 {
+                0
+            } else {
+                k
+            }
+        };
+        // 11. Repeat, while k < len,
+        while k < len {
```
aapoalas, line 817: **thought:** We should eventually write this loop as "native code". Basically, we here know that we're looking into a direct vector of u8/i8/u16/i16/(f16)/u32/i32/f32/u64/i64/f64 and we know our search element, so we can:

a) Check that the search value _can_ even match the value. eg. `-3.14` will never match a unsigned or signed index.
b) Get a slice of the TypedArray data and then perform an `index_of` on that.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Thank you for the review!! Since is_number doesn’t support BigInt, I used is_numeric instead. Does this approach make sense? 👀

https://github.com/trynova/nova/blob/8e37b4473fc152f6623a544d978d5a10d8860cae/nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs#L954-L956

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Sorry, I couldn’t figure out how to implement this optimization. Could you provide a bit of explanation? 🙇‍♂️

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
aapoalas: Unfortunately it probably doesn't. Instead, you should split the code into (at least) two different paths, one for BigInt64 | BigUint64 and one for the others.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
aapoalas: To implement it would be best to split at least the loop part into a separate function has a `T: Viewable` generic. The Viewable trait then has a `from_value` methods that can be used to turn our search element into a T; I think we'd also need to add a method to get a `&[T]` / `&mut [T]` slice (the borrow would be derived from the Agent reference for safety) from the TypedArray. Now you can use normal Rust methods to perform index_of.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Thank you so much, again!
Would the implementation look something like this?

```ts
if from_index.is_undefined() {
    if matches!(o, TypedArray::BigInt64Array(_) | TypedArray::BigUint64Array(_)) {
        if !search_element.is_bigint() {
            return Ok((-1).into());
        }
    } else {
        if !search_element.is_numeric() {
            return Ok((-1).into());
        }
    }
}
```

aapoalas: Instead of `is_numeric` the other side should be `is_number` but yeah, otherwise this seems roughly correct.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
aapoalas: Ah: `from_le_value` / `from_be_value` is actually the wrong thing to use here; we shouldn't coerce the `search_element` to `T`; instead what we need is a separate `try_from_value` kind of API that returns `Option<Self>` I guess?

So eg. for `u8` and `U8Clamped` it would check if the value is is already an integer between 0 and 255, or a SmallFloat with value -0 I guess. If the answer is yes, it would return that u8 value wrapped in `Some`. Otherwise it would return `None`.

Now the caller can use that to turn `search_element: Value` into a `search_element: T` by calling that `T::try_from_value(search_element)`; if they get `None` they immediately return `-1` and otherwise they now do a search in the `&[T]` slice.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Thank you! I’ve fixed this part for now!
https://github.com/trynova/nova/pull/556/commits/eb967cd066115235747eee749e32d7377b767a75

aapoalas: Hmm, noticed now that the from_index being undefined is not related to returning `-1`. The `matches!(o, ...)` stuff is entirely unrelated to that and should not be inside that if-branch.

yossydev: Thank you! Sorry, I assumed it was only when it was confirmed to be undefined, so I placed the check afterward.

Is your idea something like this?

```rs
if matches!(o, TypedArray::BigInt64Array(_) | TypedArray::BigUint64Array(_)) {
        if !search_element.is_bigint() {
            return Ok((-1).into());
        }
    } else {
        if !search_element.is_numeric() {
            return Ok((-1).into());
        }
    };

if from_index.is_undefined() {
    assert_eq!(n.into_i64(), 0);
};
```

aapoalas: Should be `is_number` but otherwise looks good :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: I’m late! I made some fixes based on the pair programming session, completed the remaining part of try_from_value, and added a few more adjustments to pass Test262! https://github.com/trynova/nova/pull/556/commits/06d6e343c94cb37b9e1b939d70f53a9275fa22fd

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -899,13 +899,136 @@ impl TypedArrayPrototype {
...
+                search_typed_element::<u64>(agent, o, search_element, k, gc.nogc())
+            }
+            TypedArray::Float32Array(_) => {
+                search_typed_element::<f32>(agent, o, search_element, k, gc.nogc())
+            }
+            TypedArray::Float64Array(_) => {
+                search_typed_element::<f64>(agent, o, search_element, k, gc.nogc())
+            }
+        };
+        println!("Result {:?}", result?.map_or(-1, |v| v as i64));
```
aapoalas, line 1030: **issue:** println left.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1653,3 +1776,49 @@ pub(crate) fn require_internal_slot_typed_array<'a>(
...
+        &data_block[byte_offset..]
+    };
+    // SAFETY: All bytes in byte_slice are initialized, and all bitwise
+    // combinations of T are valid values. Alignment of T's is
+    // guaranteed by align_to itself.
+    let (head, slice, _) = unsafe { byte_slice.align_to::<T>() };
+    if !head.is_empty() {
+        return Err(agent.throw_exception_with_static_message(
+            ExceptionType::TypeError,
+            "Please no",
```
aapoalas, line 1813: **issue:** Let's change this placeholder.

This would be something like "TypedArray is not properly aligned"

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -203,6 +271,18 @@ impl Viewable for u16 {
         };
         to_uint16_number(agent, value).to_le()
     }
+
+    fn try_from_value(agent: &mut Agent, value: Value) -> Option<Self> {
+        let Ok(value) = Number::try_from(value) else {
```
aapoalas, line 276: **issue:** This implementation should be the same as u8 (just using u16), not f16/f32.

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -229,6 +309,18 @@ impl Viewable for i16 {
         };
         to_int16_number(agent, value).to_le()
     }
+
+    fn try_from_value(agent: &mut Agent, value: Value) -> Option<Self> {
+        let Ok(value) = Number::try_from(value) else {
```
aapoalas, line 314: **issue:** Here too, this should look like i8.

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -255,6 +347,18 @@ impl Viewable for u32 {
         };
         to_uint32_number(agent, value).to_le()
     }
+
+    fn try_from_value(agent: &mut Agent, value: Value) -> Option<Self> {
+        let Ok(value) = Number::try_from(value) else {
```
aapoalas, line 352: **issue:** Here too, this should look like u8 and u16.

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -281,6 +385,21 @@ impl Viewable for i32 {
         };
         to_int32_number(agent, value).to_le()
     }
+
+    fn try_from_value(agent: &mut Agent, value: Value) -> Option<Self> {
+        let Ok(value) = Number::try_from(value) else {
```
aapoalas, line 390: **issue:** Here too, this should look like i8 and i16.

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -335,6 +475,33 @@ impl Viewable for i64 {
...
+                return Some(value as i64);
+            } else {
+                return None;
+            }
+        };
+        None
+    }
 }
 #[cfg(feature = "proposal-float16array")]
 impl Viewable for f16 {
```
aapoalas, line 507: **issue:** f16's implementation is missing. It should look the same as f32/f64.

aapoalas (changes requested):
Some issues remain.

yossydev:
@aapoalas 
Thank you so much for all the reviews! I had missed some things like forgetting to remove println statements. 🙇‍♂️
I’ve made the fixes on my end, so could you take another look?

aapoalas (approved):
LGTM! There were some issues after the Value lifetime merge, I got to fixing those and also ended up taking a look at the rest of the failing indexOf tests. Those revealed some issues in both the indexOf impl and in our DataBlock implementation more deeply. I fixed those and now all should be good to go :)

Thank you <3

PR #557, opened 2025-02-03
feat(heap): String interning

eliassjogreen:
Massively increases performance for string related tasks.


eliassjogreen:
Dang, really hoped that it would speed up the timeouts enough. Made a sizeable difference locally but I guess not...

nova_vm/src/heap.rs
```diff
@@ -170,6 +170,7 @@ pub struct Heap {
     // But: Source code string data is in the string heap. We need to thus drop
     // the strings only after the source ASTs drop.
     pub strings: Vec<Option<StringHeapData>>,
+    pub string_lookup_table: AHashMap<u64, StringIndex<'static>>,
```
aapoalas, line 173: **issue / nitpick:** Use `HeapString` instead of `StringIndex`. I think we generally want to avoid the "BaseIndex" types appearing basically anywhere except inside of wrapper types.

nova_vm/src/heap.rs
```diff
@@ -263,6 +264,7 @@ impl Heap {
             #[cfg(feature = "shared-array-buffer")]
             shared_array_buffers: Vec::with_capacity(0),
             strings: Vec::with_capacity(1024),
+            string_lookup_table: AHashMap::with_capacity(1024),
```
aapoalas, line 267: **thought:** I _think_ using the hashbrown's `HashTable` type might be more correct here: We could have `HashTable<HeapString>`.

nova_vm/src/heap/heap_bits.rs
```diff
@@ -1013,6 +1013,38 @@ pub(crate) fn sweep_heap_vector_values<T: HeapMarkAndSweep + std::fmt::Debug>(
...
+    let mut iter = bits.iter();
+    vec.retain_mut(|item| {
+        let do_retain = if *iter.next().unwrap() {
+            item.sweep_values(compactions);
+            true
+        } else {
+            false
+        };
+
+        if let Some(string) = item {
```
aapoalas, line 1032: **thought:** I think we can do this better: Basically, with `HashTable` we could iterate through all entries in the table and either:
1. If the value is retained (check either through the bits directly, or based on the index and compactions), unsafely mutate it in-place based on the compactions. Why? Because we know that the string data hasn't changed with the StringIndex changing, so its hash hasn't changed, so mutating the value in-place is good and proper.
2. If the value is removed, remove it in the HashTable.

If I'm not badly mistaken, I think the [retain](https://docs.rs/hashbrown/latest/hashbrown/hash_table/struct.HashTable.html#method.retain) method is basically perfect for it.

aapoalas (changes requested):
**praise:** Excellent stuff! Had one nitpick/issue that will be easy to handle, I think, but then I have one larger "issue" that I think we could take care of: We could use a slightly "lower level" hash table type.

nova_vm/src/heap/heap_bits.rs
eliassjogreen: I think I did it properly now using the retain method in the now general `sweep_lookup_table` fn. It required a new trait: `GetBaseIndexMut` but apart from that it was a suprisingly simple thing to implement.

nova_vm/src/ecmascript/types/language/string.rs
```diff
@@ -546,8 +561,13 @@ impl Scoped<'_, String<'static>> {
 
 impl CreateHeapData<StringHeapData, String<'static>> for Heap {
     fn create(&mut self, data: StringHeapData) -> String<'static> {
+        let hash = self.string_hasher.hash_one(data.as_str());
```
aapoalas, line 564: **question:** Is keeping the string hasher actually necessary? I'd assume that it holds no meaningful internal state after `hash_one` call and likewise I'd assume that creating the hasher isn't particularly expensive (I think it should just be prepping some stack data ... Maybe?)

ie. Should we have it in the Heap struct like it is now, or would it make more sense to have it be created anew each time?

nova_vm/src/heap.rs
```diff
@@ -280,10 +286,9 @@ impl Heap {
             weak_sets: Vec::with_capacity(0),
         };
 
-        heap.strings.extend_from_slice(
-            &BUILTIN_STRINGS_LIST
-                .map(|builtin_string| Some(StringHeapData::from_static_str(builtin_string))),
-        );
+        for builtin_string in BUILTIN_STRINGS_LIST {
```
aapoalas, line 289: **note:** At some point we'll want to calculate these hashes at compile time :)

aapoalas (approved):
LGTM, great work on figuring this out! The raw HashTable API isn't necessarily the easiest one to work with so mad props for getting it right basically immediately! <3

nova_vm/src/heap.rs
eliassjogreen: Yep, agreed, but hold off on that.

nova_vm/src/ecmascript/types/language/string.rs
eliassjogreen: The hashbrown hasher [`RandomState` seems to simply initialize four constants](https://rust-lang.github.io/hashbrown/src/ahash/random_state.rs.html#89-94), so it's probably a really cheap operation. But I think it makes sense to keep it contained along side the hash table as long as were not using a constant state for the hasher.

PR #563, opened 2025-02-09
fix(ecmascript): Do not expose uninitialized data through ArrayBuffer resizing

aapoalas:
First proper, honest-to-god UB & vulnerability in the engine! Good thing no one uses this yet <3


yossydev:
I think there was also something that was skipped because it succeeds in my environment but fails in CI. Would this fix that as well?

aapoalas:
> I think there was also something that was skipped because it succeeds in my environment but fails in CI. Would this fix that as well?

Mmm, what do you mean? Do you have a list of tests that pass for you locally but fail on CI still? Note that this was already merged.

yossydev:
Ah, sorry! Never mind, forget about it!

PR #565, opened 2025-02-11
chore(ecmascript): used try and Assert Determination

yossydev:
It was the same with TypedArray, but using `assert_eq!(from_index.is_undefined(), n.into_i64() == 0);` causes some test262 cases to fail.

Even if I use `assert_eq` only when fromIndex is undefined, I believe it’s essentially doing the same thing.

I also tried using `try_to_integer_or_infinity`. However, based on my measurements using the time command on Unix, it didn’t seem to make things any faster.

So, it might be pointless… but I gave it a shot anyway.


nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -1352,9 +1352,18 @@ impl ArrayPrototype {
             return Ok(false.into());
         }
         // 4. Let n be ? ToIntegerOrInfinity(fromIndex).
-        let n = to_integer_or_infinity(agent, from_index, gc.reborrow())?;
+        let n = if let TryResult::Continue(n) =
+            try_to_integer_or_infinity(agent, from_index, gc.nogc())
```
aapoalas, line 1356: **note:** As I mentioned in either the TypedArray includes or indexOf PRs, using a `try_method` with a `TryResult::Break` branch going into a non-try method only really makes sense if it allows us to skip rooting (scoping) of values.

In this case that would mean leaving `o` unscoped and only scoping it conditionally in the `else` branch here; also the same try and conditional scoping would need to be done for the `length_of_array_like` above.

In this case those probably somewhat make sense, since it is mildly likely that getting the length of an array-like object doesn't require calling into JavaScript (though it's also fairly possible that we're staring at a getter there, like with `get %TypedArray%.prototype.length`) and we'll then get a `TryResult::Continue(len)` as our result for `try_length_of_array_like`.

And then of course it's quite likely that the from_index will return `TryResult::Continue(n)` from `try_to_integer_or_infinity` so we can avoid the scoping there as well. But the code does get a bit complicated when we have multiple try-else branches that need to conditionally scope the same thing and you want to avoid rescoping if the previous branch already did the scoping etc... :)

aapoalas (approved):
Looks good to me, though the try-methods are kind of unnecessary right now. You could either revert them back to the original, or alternatively make scoping of `o` be a conditional thing done in the else-branches.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
yossydev: Thanks for the review!
What I wanted to do in this PR was to change the assert_eq condition.
I also thought about converting it to a try-method while I was at it, but since optimization isn’t something that needs to be done immediately, I decided to remove it for now! https://github.com/trynova/nova/pull/565/commits/d6096a5bfcbf5dd10b6316f81f905d278476f69e

That said, I feel like I’m starting to understand how to use try-methods better. Thanks again!

aapoalas (approved):
Lgtm, thanks<3

PR #566, opened 2025-02-16
Add an `--expose-internals` CLI flag

andreubotella:
The test262 runner uses the Nova CLI, but it also needs access to functions exposing either some of the engine's internals, or things that are not accessible from JS code but could be from a runtime built on top of the engine. An example of this is `detachArrayBuffer`, which is currently exposed on the global.

Rather than exposing these functions on the global, this patch instead adds an `--expose-internals` CLI flag, which adds a `__nova__` object to the global which would contain those functions. So the function to detach an ArrayBuffer would now be accessible from `__nova__.detachArrayBuffer`.

As part of this, this patch also adds a public
`OrdinaryObject::create` static method that allows creating objects by embedders, which was not previously possible. This object is identical to that which would be created by evaluating `{}` in the current realm.


nova_vm/src/ecmascript/types/language/object.rs
```diff
@@ -396,6 +397,15 @@ impl<'a> OrdinaryObject<'a> {
     pub(crate) const fn get_index(self) -> usize {
         self.0.into_index()
     }
+
+    pub fn create(agent: &mut Agent, gc: NoGcScope<'a, '_>) -> Self {
```
aapoalas, line 401: **suggestion:** Should we maybe name this `create_empty_object`?

aapoalas (approved):
Looks good to me, thank you! <3

nova_vm/src/ecmascript/types/language/object.rs
andreubotella: Renamed.

PR #567, opened 2025-02-16
Support realms in test262

andreubotella:
If the `--expose-internals` flag is enabled, this patch adds a `__nova__.createRealm()` function that creates a new realm and returns its global. This is then used by the test262 runner to support testing cross-realm interactions for the first time.

----------

This PR must be merged after #566.


tests/nova-harness.js
```diff
@@ -16,8 +16,22 @@
 // to `$262.global` (needed for old tests, since `globalThis` was only added in
 // ES2020).
 
-globalThis.$262 = {
-    global: globalThis,
-    detachArrayBuffer: __nova__.detachArrayBuffer,
-};
-delete globalThis.__nova__;
\ No newline at end of file
+// This function must be completely independent of the current realm, and it
```
aapoalas, line 19: **praise:** This is really good documentation <3

aapoalas (approved):
LGTM! I will eventually want to refactor the realms system, probably splitting into a "main realm" and "other realms" so that the most common use case gets to keep one the current level of indirection from `&Agent` to heap data. But this doesn't really block or hinder that work in any way. Great stuff!

PR #571, opened 2025-02-24
feat(ecmascript): %TypedArray%.prototype.findIndex

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.findindex


aapoalas (approved):
LGTM! Thank you!

For performance reasons we'll probably want to split the `find_via_predicate` from the Array's more generic version at some point, but that's not important right now.

PR #572, opened 2025-02-24
chore: update rust toolchain version by cargo 1.85

yossydev:
CI is running Rust 1.85, but the rust-toolchain file in Nova was set to 1.81.
As a result, the following error occurs in CI.

```
error: failed to parse manifest at `/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/munge_macro-0.4.2/Cargo.toml`
Caused by:
  feature `edition2024` is required
  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.81.0 (2dbb1af80 2024-08-20)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
Error: Process completed with exit code 101.
```

In this PR, I unified the Cargo version used in Nova to 1.85 and ran `cargo clippy --fix` along with some manual fixes.
Alternatively, aligning the CI version with the rust-toolchain version could also be a solution.

### DIFF
--- a/nova_vm/src/ecmascript/builders/builtin_function_builder.rs
+++ b/nova_vm/src/ecmascript/builders/builtin_function_builder.rs
@@ -280,7 +280,7 @@ impl<'agent, P, L, B, Pr> BuiltinFunctionBuilder<'agent, P, L, NoName, B, Pr> {
     }
 }
 
-impl<'agent, P, L, B, Pr> BuiltinFunctionBuilder<'agent, P, L, CreatorName, B, Pr> {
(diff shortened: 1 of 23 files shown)


eliassjogreen:
Nice! This will be a prerequisite for if we want to write custom nova specific lints using [dylint](https://github.com/trailofbits/dylint).

aapoalas (approved):
LGTM, thank you <3

Issue #573, opened 2025-02-25
Linting


eliassjogreen:
As the Nova project grows it is important that we maintain convention and especially conform to certain rules when it comes to the GC (as described in the [CONTRIBUTING](https://github.com/trynova/nova/blob/main/CONTRIBUTING.md) document). To ease this I suggest we implement some project specific lints using [dylint](https://github.com/trailofbits/dylint) just for nova and projects embedding nova.

## Rules

These are some rules I suggest we implement off the top of my head:

- [x] Warn when a function could take a `NoGcScope` instead of a `GcScope`
- [x] Make sure `Agent` is the first argument of fn definitions
- [x] Make sure `GcScope`/`NoGcScope` is the last argument of fn definitions
- [x] Make sure `GcScope`/`NoGcScope` are only passed by value
- [x] Check formatting of spec doc comments
- [ ] No unbinding of values into local variables
- [ ] No duplicate scoping of values
- [ ] Calling reborrow only at the call site

PR #574, opened 2025-02-25
feat(linter): Create `nova_lint` and implement three basic rules

eliassjogreen:
This is an initial take on #573 using dylint. I implemented three simple `agent_comes_first`, `gc_scope_comes_last` and `gc_scope_is_only_passed_by_value` rules just as a proof-of-concept and it's was actually rather fun and simple to implement!

It seems dylint requires rust nightly to build and test, which is why it's excluded from the main workspace and instead lives as it's entirely own crate with it's own `rust-toolchain`. I tried to get it to work alongside nova without nightly but nightly seems to be required to expose the internals of rust and allow us to use them to build our own clippy rules.

### DIFF
--- a/nova_lint/src/gc_scope_comes_last.rs
+++ b/nova_lint/src/gc_scope_comes_last.rs
@@ -1,8 +1,9 @@
-use clippy_utils::{diagnostics::span_lint_and_help, match_def_path};
+use clippy_utils::diagnostics::span_lint_and_help;
 use rustc_hir::{def_id::LocalDefId, intravisit::FnKind, Body, FnDecl};
 use rustc_lint::{LateContext, LateLintPass};
-use rustc_middle::ty::{Ty, TyKind};
 use rustc_span::Span;
+
+use crate::{is_gc_scope_ty, is_no_gc_scope_ty};
 
 dylint_linting::declare_late_lint! {
     /// ### What it does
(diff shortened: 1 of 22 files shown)


nova_lint/src/gc_scope_comes_last.rs
```diff
@@ -0,0 +1,93 @@
...
+use rustc_span::Span;
+
+dylint_linting::declare_late_lint! {
+    /// ### What it does
+    ///
+    /// Checks that the gc scope is the last parameter of a function.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// The gc scope is expected to be the last parameter of a function
```
aapoalas, line 14: **note:** The exact reason for why it's last is that ... "the moment" when a GcScope parameter is passed to a call (or has `reborrow()` called on it), it invalidates all values that refer to it. If GcScope was the second parameter (as it originally was) then it would invalidate all GC-bound parameters before they are passed into the call or get to call `unbind()`.

eg.
```rs
let data = data.bind(gc.nogc());
call(agent, gc.reborrow(), data.unbind());
```
This wouldn't work beause `gc.reborrow()` invalidates `data` immediately, meaning that when `data.unbind()` is being called the `data` is already invalidated and illegal to use, leading to a borrow checker error.

nova_lint/src/gc_scope_comes_last.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// according to the Nova engines conventions.
+    ///
+    /// ### Example
+    ///
+    /// ```rust
+    /// fn bar(gc: &GcScope<'_, '_>, other: &Other) {}
```
aapoalas, line 20: **note:** GcScope should never be passed as ref, always by value.

nova_lint/ui/gc_scope_comes_last.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    fn test_one_param(&self, _foo: ()) {
+        unimplemented!()
+    }
+
+    fn test_self_and_owned_gc_scope_only(&self, gc_scope: GcScope<'_, '_>) {
+        unimplemented!()
+    }
+
+    fn test_self_and_something_after_gc_scope(&self, gc_scope: &GcScope<'_, '_>, foo: ()) {
```
aapoalas, line 69: **note:** This sort of `&GcScope` call should probably be illegal. `&mut GcScope` kind of works (it's just less efficient than `GcScope` by value). `&GcScope` is mostly equivalent to `NoGcScope` by value.

nova_lint/src/gc_scope_comes_last.rs
eliassjogreen: Makes sense! I'll update the lint to not peel the references then. Adding another lint which checks that we only pass `GcScope` by value should be trivial.

nova_lint/src/gc_scope_comes_last.rs
```diff
@@ -1,29 +1,40 @@
...
+    /// The gc scope parameter should be passed as the last parameter of a
+    /// function because it invalidates all values which refer to it, take
+    /// for example the following code:
+    /// 
+    /// ```rust
+    /// let data = data.bind(gc.nogc());
+    /// call(agent, gc.reborrow(), data.unbind());
+    /// ```
+    /// 
+    /// This wouldn't work beause `gc.reborrow()` invalidates data immediately,
```
aapoalas, line 24: **nitpick:** Use backticks around "data" here and below in "the data is already"; those refer to the `data` variable.

aapoalas:
Really cool stuff! <3 <3 <3 <3

PR #585, opened 2025-03-06
feat(ecmascript): %TypedArray%.prototype.lastIndexOf

yossydev:
doc: https://tc39.es/ecma262/#sec-%typedarray%.prototype.lastindexof


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1852,6 +2033,7 @@ fn search_typed_element<T: Viewable + std::fmt::Debug>(
     search_element: Value,
     k: usize,
     len: usize,
+    ascending: bool,
```
yossydev, line 2036: Since the internal processing is almost the same, how about adding an option like ascending in find_via_predicate?

aapoalas: Make it a `const ASCENDING: bool` generic.

aapoalas (changes requested):
LGTM except for that let's make the method take the ascending/descending as const generic. Nice work! <3

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: thank you! I fixed it! https://github.com/trynova/nova/pull/585/commits/6092afa028efb6af0d13f834cfd1bab69c928a64

PR #588, opened 2025-03-10
feat(ecmascript): implement Iterator.prototype.forEach

r58Playz:
Implements `Iterator.prototype.forEach`.

I was originally planning to do more of the `Iterator` functions but couldn't figure out how to implement Abstract Closures. Is there an existing place in the code that uses them?


nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
```diff
@@ -34,15 +50,90 @@ impl IteratorPrototype {
...
+                nogc,
+            );
+            // b. Return ? IteratorClose(iterated, error).
+            return iterator_close(agent, o.unbind(), Err(error), gc);
+        };
+        let scoped_procedure = Value::from(procedure).scope(agent, gc.nogc());
+
+        // 5. Set iterated to ? GetIteratorDirect(O).
+        // if i rebind `iterated` here borrow checker complains
+        let iterated = get_iterator_direct(agent, o.unbind(), gc.reborrow()).unbind()?;
```
r58Playz, line 90: Binding `iterated` here results in both the reborrow calls in the loop saying "cannot borrow `gc` as mutable because it is also borrowed as immutable`. I can't scope the iterator either so I just left it as unbound which seems to fix it. This is probably wrong, though.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
```diff
@@ -34,15 +50,90 @@ impl IteratorPrototype {
...
+        let Some(iterated) = iterated else {
+            // what should happen here?
+            todo!();
```
r58Playz, line 93: The spec isn't clear about what should happen here. Should it just be a `TypeError` like the above for `procedure`?

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
aapoalas: Yeah, it'll eventually lead to UAF once we do GC.

Now that the `done` field is gone from the iterator record, it's equal to just two bound values. So what I've done elsewhere is to take out the two bound values and scoped them separately, then I recreate the IteratorRecord at every callsite. It's not pretty, but it works and is probably even fairly optimal.

aapoalas:
> I was originally planning to do more of the `Iterator` functions but couldn't figure out how to implement Abstract Closures. Is there an existing place in the code that uses them?

So far, we've always managed to avoid needing to actually create AC's. Some fairly common ACs did get their own special flavour of function (Promise resolving functions), but most have just been rolled into the engine code generally. Eg. The way AsyncGenerator ACs are done is that they appear as parts of the re-entry code paths that are taken by resuming async generators.

So, the answer is: It's on a case-by-case basis. If we can roll the ACs into being part of eg. yield paths then we'll do that. If the ACs don't capture anything (or we can imitate the capturing through this argument somehow?) then we can use just normal built-in function objects. If neither of these is possible, then we'll have to look into actually implementing or supporting ACs somehow.

aapoalas (changes requested):
Hello and welcome! Thank you for the method implementation, looks good to me! The unbind thing we can fix fairly simply by recreating the IteratorRecord at callsites. Once that's done we can merge this no problem <3

Great job!

nova_vm/src/ecmascript/abstract_operations/operations_on_iterator_objects.rs
```diff
@@ -123,7 +161,7 @@ pub(crate) fn get_iterator_from_method<'a>(
     }))
 }
 
-/// ### [7.4.3 GetIterator ( obj, kind )](https://tc39.es/ecma262/#sec-getiterator)
+/// ### [7.4.4 GetIterator ( obj, kind )](https://tc39.es/ecma262/#sec-getiterator)
```
aapoalas, line 164: **praise:** Thanks for fixing the numbering <3

aapoalas (dismissed):
LGTM! There's a conflict with metrics (the most annoying thing with merging any changes here, sorry about it :( ) but aside from that I'm happy to merge this. Thank you, great work, and welcome to the project! <3

PR #589, opened 2025-03-10
ci: run build and test on all three platforms

kermanx:
Context: https://github.com/trynova/nova/pull/586#issuecomment-2708921873, and let's see if running tests on Windows is actually much slower.

This PR completes the TODO comment in ci.yml.


aapoalas:
Hmm... It did indeed complete in a reasonable time, though some tests fail (which is in itself really damn interesting).

PR #590, opened 2025-03-10
feat(ecmascript): %TypedArray%.prototype.findLastIndex

yossydev:
You mentioned it could be optimized before, but I just went ahead and implemented it for now.

doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.findlastindex


aapoalas (commented):
LGTM! Nice <3

The optimisation would be making a `find_via_predicate` version here in `typed_array_intrinsic_object.rs` that is generic on `T: Viewable` ... maybe. Not sure if it would actually give a meaningful performance increase now that I think about it.

aapoalas (approved):

PR #591, opened 2025-03-13
feat(ecmascript): Implement Iterator functions that don't need a helper

r58Playz:
Implements `Iterator` object, `Iterator.prototype.every`, `Iterator.prototype.find`, `Iterator.prototype.reduce`, `Iterator.prototype.some`, and `Iterator.prototype.toArray`.

Refactors `set_default_global_bindings` to use a macro.


nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
```diff
@@ -51,6 +93,203 @@ impl IteratorPrototype {
...
+        // 4. If IsCallable(predicate) is false, then
+        let Some(predicate) = is_callable(predicate, nogc).unbind().bind(nogc) else {
+            // a. Let error be ThrowCompletion(a newly created TypeError object).
+            let error = agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "'predicate' is not callable",
+                nogc,
+            );
+            // b. Return ? IteratorClose(iterated, error).
+            return iterator_close(agent, o.unbind(), Err(error), gc);
```
aapoalas, line 225: **nitpick:** I previously added an `iterator_close_with_error` method that can be used here instead.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
```diff
@@ -51,6 +93,203 @@ impl IteratorPrototype {
...
+            };
+            let value = iterator_step_value(agent, record, gc.reborrow())
+                .unbind()
+                .bind(gc.nogc())?;
+
+            // b. If value is done, return undefined.
+            let Some(value) = value else {
+                return Ok(Value::Undefined);
+            };
+            let scoped_value = value.scope(agent, gc.nogc());
```
aapoalas, line 263: **suggestion:** Here it would be preferable to avoid the scope on each loop; you should instead prepare a `let scoped_value = Value::Undefined.scope_static();` outside of the loop and use `scoped_value.replace(value)` here inside the loop. I wrote a bit more about this in [a new blog post / guide](https://trynova.dev/blog/guide-to-nova-gc), see under the title "Changing values".

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
```diff
@@ -74,7 +313,7 @@ impl IteratorPrototype {
 
         // 3. Let iterated be the Iterator Record { [[Iterator]]: O, [[NextMethod]]: undefined, [[Done]]: false }.
         // 4. If IsCallable(procedure) is false, then
-        let Some(procedure) = is_callable(procedure, nogc).unbind().bind(gc.nogc()) else {
+        let Some(procedure) = is_callable(procedure, nogc).unbind().bind(nogc) else {
```
aapoalas, line 316: **praise:** Thanks for fixing these.

**note:** I don't know if you've noticed, but there is a bit of a tradeoff between using a `let nogc = gc.nogc()` and individual `gc.nogc()` calls: The former gets rid of repeated calls, while the latter will give easier-to-debug borrow checker errors. What I've been doing is that when I initially write a function, I'll use individual `gc.nogc()` calls, and then later I'll convert it over to `let nogc = gc.nogc()` once I've figured out all the issues.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
```diff
@@ -142,16 +381,310 @@ impl IteratorPrototype {
...
+        }) = iterated
+        else {
+            return Err(throw_not_callable(agent, gc.into_nogc()));
+        };
+        let iterator = iterator.scope(agent, gc.nogc());
+        let next_method = next_method.scope(agent, gc.nogc());
+
+        let (accumulator, mut counter) = match scoped_initial_value.get(agent).bind(gc.nogc()) {
+            // 6. If initialValue is not present, then
+            Value::Null | Value::Undefined => {
```
aapoalas, line 438: **issue:** This is unfortunately not what "is not present" means. "not present" means that no parameter was given. eg. in `reduce(cb, undefined)` the initial value is present, while in `reduce(cb)` it is not.

So, instead you'll need to check the length of `arguments` to determine presence :(

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
```diff
@@ -142,16 +381,310 @@ impl IteratorPrototype {
...
+            .bind(gc.nogc())?;
+        let Some(IteratorRecord {
+            iterator,
+            next_method,
+            ..
+        }) = iterated
+        else {
+            return Err(throw_not_callable(agent, gc.into_nogc()));
+        };
+        let iterator = iterator.scope(agent, gc.nogc());
```
aapoalas, line 563: **nitpick (blocking):** With `get_iterator_direct` we know that `iterated.iterator === o`, so `scoped_o` already contains the `iterator` Object, you can just do `let iterator = scoped_o` instead of scoping the object a second time.

Note: This applies to each of the methods.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
```diff
@@ -142,16 +381,310 @@ impl IteratorPrototype {
...
+
+            // d. IfAbruptCloseIterator(result, iterated).
+            let record = IteratorRecord {
+                iterator: iterator.get(agent),
+                next_method: next_method.get(agent),
+            };
+            let result = if_abrupt_close_iterator!(agent, result, record, gc);
+
+            // e. Set accumulator to result.
+            accumulator = result.scope(agent, gc.nogc());
```
aapoalas, line 508: **issue:** Here too you should just `accumulator.replace(result)`.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
```diff
@@ -142,16 +381,310 @@ impl IteratorPrototype {
...
+            ..
+        }) = iterated
+        else {
+            return Err(throw_not_callable(agent, gc.into_nogc()));
+        };
+        let iterator = iterator.scope(agent, gc.nogc());
+        let next_method = next_method.scope(agent, gc.nogc());
+
+        // 4. Let items be a new empty List.
+        let mut items: Vec<Value> = Vec::new();
```
aapoalas, line 649: **issue:** This needs to be a `Vec<Scoped<Value>>`, otherwise it's UAF in the making.

nova_vm/src/ecmascript/execution/realm.rs
```diff
@@ -332,745 +332,258 @@ pub(crate) fn set_default_global_bindings<'a>(
     let global = agent[realm_id].global_object.scope(agent, gc.nogc());
 
     // 2. For each property of the Global Object specified in clause 19, do
-    // a. Let name be the String value of the property name.
-    // b. Let desc be the fully populated data Property Descriptor for the
-    //    property, containing the specified attributes for the property. For
-    //    properties listed in 19.2, 19.3, or 19.4 the value of the [[Value]]
-    //    attribute is the corresponding intrinsic object from realmRec.
-    // c. Perform ? DefinePropertyOrThrow(global, name, desc).
+    macro_rules! define_property {
```
aapoalas, line 335: **praise:** Oh wow! Great simplification, thank you! <3

aapoalas (changes requested):
Wow, amazing work! Thank you!

I found some places where we can improve, nothing major though. Mostly related to avoiding duplicate scoping and such.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
r58Playz: The borrow checker doesn't like the `'static` lifetime on `scoped_value` if I try to move it out of the loop. Each call to `reborrow` says "cannot borrow gc as mutable because it is also borrowed as immutable". The rebinding of `value` says "argument requires that 'gc must outlive 'static".

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
r58Playz: I've removed scoped_o entirely as I think it makes it clearer that the iterator is being used.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
r58Playz: Same here, the borrow checker complains. Each call to `reborrow` says "cannot borrow gc as mutable because it is also borrowed as immutable". The rebind in the `if_abrupt_close_iterator` call says "argument requires that 'gc must outlive 'static".

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_constructor.rs
```diff
@@ -42,6 +42,19 @@ impl IteratorConstructor {
...
+        if new_target
+            == agent
+                .running_execution_context()
+                .function
+                .unwrap()
+                .into_object()
+        {
+            return Err(agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "Iterator is abstract",
```
aapoalas, line 54: **nitpick:** Let's use the same error message as Firefox does: `Iterator constructor can't be used directly`.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
aapoalas: Ah, sorry, I had forgotten that due to lifetime complexities, the value parameter to `replace` requires `.unbind()` when being passed in.

I pushed a fix to use replace in these two places.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/iteration/iterator_prototype.rs
aapoalas: Nice <3

aapoalas (approved):
I fixed the lifetime issues (just more required unbinding that I'd forgotten about :D ) and the one nit I had about the error message.

Will merge once tests say everything is okay. Great work, thank you very much! <3

PR #592, opened 2025-03-14
feat(ecmascript): array and object assignment pattern

kermanx:
This PR adds support for array and object assignment pattern.

TODO: Object assignment pattern's rest part. (I am not sure how to design it. The existing implementation of object binding pattern requires many special instructions, but are they also required in Object assignment pattern?)


nova_vm/src/engine/bytecode/bytecode_compiler/assignment.rs
```diff
@@ -0,0 +1,354 @@
...
+                        if is_reference(&self.right) {
+                            ctx.add_instruction(Instruction::GetValue);
+                        }
+                        ctx.add_instruction(Instruction::LoadCopy);
+                        self.left.compile(ctx);
+                        ctx.add_instruction(Instruction::Store);
+                    }
+                    AssignmentOperator::LogicalNullish | AssignmentOperator::LogicalOr => {}
+                    _ => {
+                        // TODO: throw
```
aapoalas, line 35: **question:** Is it even possible to end up here? I'm expecting oxc to give parsing or syntax errors for these. (And actually I can't figure out how to get `{} &&= expr` or `[] &&= expr` kind of stuff working either; are those valid either?)

nova_vm/src/engine/bytecode/bytecode_compiler/assignment.rs
```diff
@@ -0,0 +1,354 @@
...
+            }
+            ast::AssignmentTarget::TSAsExpression(_)
+            | ast::AssignmentTarget::TSSatisfiesExpression(_)
+            | ast::AssignmentTarget::TSNonNullExpression(_)
+            | ast::AssignmentTarget::TSTypeAssertion(_)
+            | ast::AssignmentTarget::TSInstantiationExpression(_) => unreachable!(),
+        };
+
+        if self.operator == AssignmentOperator::Assign {
+            ctx.add_instruction(Instruction::PushReference);
```
aapoalas, line 52: **suggestion (nitpick):** I _think_ we can do this a bit more optimally for when `self.right` is not a reference or at least when it is a literal: We can skip the PushReference and PopReference, conceptually leaving `self.left` in the reference slot while `self.right` is just brought into the result slot.

nova_vm/src/engine/bytecode/bytecode_compiler/assignment.rs
```diff
@@ -0,0 +1,354 @@
...
+            // 8. Perform ? PutValue(lref, r).
+            ctx.add_instruction(Instruction::PopReference);
+            ctx.add_instruction(Instruction::PutValue);
+            // 9. Return r.
+            ctx.add_instruction(Instruction::Store);
+        }
+    }
+}
+
+impl CompileEvaluation for ast::AssignmentTarget<'_> {
```
aapoalas, line 175: **praise:** Good job extracting this out <3

nova_vm/src/engine/bytecode/bytecode_compiler/assignment.rs
```diff
@@ -0,0 +1,354 @@
...
+}
+
+impl CompileEvaluation for ast::ObjectAssignmentTarget<'_> {
+    fn compile(&self, ctx: &mut CompileContext) {
+        ctx.add_instruction(Instruction::ToObject);
+        if self.properties.len() > 1 || self.rest.is_some() {
+            ctx.add_instruction(Instruction::LoadCopy);
+        }
+        for (index, property) in self.properties.iter().enumerate() {
+            match property {
```
aapoalas, line 235: **suggestion:** This could maybe be extracted into its own CompileEvaluation impl, just to clean up the inside of the loop a little.

aapoalas (approved):
Wow, amazing work, figuring out how the bytecode compiler works and especially how on earth all the assignment expressions work, those complicated devils!

How did you like the bytecode work? I've always found it pretty fun and amazing, how the small instructions come together to form something great. I look forward to one day maybe making an optimising compiler out of this all :)

I'm perfectly happy to merge this as-is; I did leave one or two nitpicks and suggestions, if you want to take those up. Also, if you want to still finish up some of the remaining todo!'s then that's fine as well. Let me know how you want to go forward with this, and thank you <3

nova_vm/src/engine/bytecode/bytecode_compiler/assignment.rs
kermanx: I agree with you. Only `=` should work with object/array assignment pattern. I would like to open an issue in Oxc for the missing SyntaxError later.

kermanx:
Thank you for reviewing~


> Also, if you want to still finish up some of the remaining todo!'s then that's fine as well. Let me know how you want to go forward with this, and thank you <3

At the moment I mostly choose what to do at random. In the future I would like to do things that improve performance :)

nova_vm/src/engine/bytecode/bytecode_compiler/assignment.rs
kermanx: I opened a issue in Oxc but suddenly found that Oxc actually reports it: https://playground.oxc.rs/#eNo9j71OxEAMhF/l5AqkFIBEE4kOXQuClsbZOKeVNnZkew9OIe+OcyFX7frvm5kZErRwNy+H35cD3n8xNCDQzqCV18cu7PgDrWulBkpmh3bAYlFYkon2iV3GTspeuSLbIDr+Ly8NTKhGuiKxFPn+IK/Kb9Ut93SsnDxLCG7nk1Lsnukdldi2biBW9Ssi/jeFjemoJwprQPb08PgcIbJJQaf+lVJBxRUfpN1Nkp5OdE1IjF2hT6maaMTppjZmzkPe9ZKwq5RjWF+PzqSdWKTfgMvyB489dQY=. So I closed that issue.

PR #593, opened 2025-03-15
feat(ecmascript): %TypedArray%.prototype.reverse

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.reverse


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1780,13 +1780,88 @@ impl TypedArrayPrototype {
...
+            | TypedArray::BigUint64Array(_)
+            | TypedArray::Float64Array(_) => typed_array_length::<u64>(agent, &ta_record, gc),
+        } as i64;
+        let scoped_o = o.scope(agent, gc);
+        // 4. Let middle be floor(len / 2).
+        let middle = len / 2;
+        // 5. Let lower be 0.
+        let mut lower: i64 = 0;
+        // 6. Repeat, while lower ≠ middle,
+        while lower != middle {
```
aapoalas, line 1820: **issue:** While this is absolutely correct, we can and should do much better: We should use the same `as_slice` logic we used in `search_typed_element` and perform a native Rust [`reverse`](https://doc.rust-lang.org/std/primitive.slice.html#method.reverse) on the `&mut [T: Viewable]`.

aapoalas (changes requested):
Let's try to do this in (roughly) the most performant way possible :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2150,3 +2206,49 @@ fn search_typed_element<T: Viewable + std::fmt::Debug, const ASCENDING: bool>(
...
+fn reverse_typed_array<T: Viewable + Copy + std::fmt::Debug>(
+    agent: &mut Agent,
+    ta: TypedArray,
+    len: usize,
+    gc: NoGcScope,
+) -> JsResult<()> {
+    let array_buffer = ta.get_viewed_array_buffer(agent, gc);
+    let byte_offset = ta.byte_offset(agent);
+    let byte_length = ta.byte_length(agent);
+    let byte_slice = array_buffer.as_slice(agent);
```
aapoalas, line 2219: **issue:** Use `as_mut_slice` to get mutable access to the AB directly.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2150,3 +2206,49 @@ fn search_typed_element<T: Viewable + std::fmt::Debug, const ASCENDING: bool>(
...
+        return Err(agent.throw_exception_with_static_message(
+            ExceptionType::TypeError,
+            "TypedArray is not properly aligned",
+            gc,
+        ));
+    }
+
+    // Fast path: If there’s only one element, reversing it won’t change anything, so return immediately.
+    let len = len.min(slice.len());
+    if len <= 1 {
```
aapoalas, line 2243: **nitpick:** This is probably not really necessary.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2150,3 +2206,49 @@ fn search_typed_element<T: Viewable + std::fmt::Debug, const ASCENDING: bool>(
...
+        return Ok(());
+    }
+
+    let mut temp = slice[..len].to_vec();
```
aapoalas, line 2247: **issue:** You can just use https://doc.rust-lang.org/std/primitive.slice.html#method.reverse on the mutable slice directly. No need for the `to_vec` and `set_offset_by_byte` stuff.

aapoalas (changes requested):
We can go deeper! :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
aapoalas: Close, but we can still go deeper! :)

yossydev:
@aapoalas 
Thanks! I made the slice itself mutable—so this was the correct approach after all! https://github.com/trynova/nova/pull/593/commits/f4fd1b3f251c4013af32826822c0630d3a59e5be

aapoalas (approved):
LGTM! Nice! <3

PR #601, opened 2025-03-19
feat(ecmascript): TypedArrayCreateFromConstructor and %TypedArray%.of

yossydev:
doc:
- https://tc39.es/ecma262/multipage/indexed-collections.html#sec-typedarraycreatefromconstructor
- https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.of


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1160,120 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
...
+/// and returns either a normal completion containing a TypedArray or a throw completion.
+/// It is used to specify the creation of a new TypedArray using a constructor function.
+pub(crate) fn typed_array_create_from_constructor<'a>(
+    agent: &mut Agent,
+    constructor: Function,
+    arguments: ArgumentsList,
+    mut gc: GcScope,
+) -> JsResult<TypedArray<'a>> {
+    // 1. Let newTypedArray be ? Construct(constructor, argumentList).
+    let arguments = arguments.bind(gc.nogc());
```
aapoalas, line 1176: **nitpick:** Also bind `constructor` just to show it doesn't get used after the `construct` call.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1160,120 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
...
+    let new_typed_array = construct(
+        agent,
+        constructor,
+        Some(arguments.unbind()),
+        None,
+        gc.reborrow(),
+    )?;
+    // 2. Let taRecord be ? ValidateTypedArray(newTypedArray, seq-cst).
```
aapoalas, line 1184: **nitpick:** You could temporarily unbind the `new_typed_array` here and use `let gc = gc.into_nogc();` to avoid all the `gc.nogc()` calls later and just use `gc` everywhere.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1160,120 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
...
+    let ta_record = validate_typed_array(
+        agent,
+        new_typed_array.into_value().unbind(),
+        Ordering::SeqCst,
+        gc.nogc(),
+    )?;
+    let o = ta_record.object;
+    let scoped_o = o.scope(agent, gc.nogc());
+    // 3. If the number of elements in argumentList is 1 and argumentList[0] is a Number, then
+    if arguments.unbind().len() == 1 && arguments[0].unbind().is_number() {
```
yossydev, line 1194: @aapoalas 
 The changes in [nova/pull/599](https://github.com/trynova/nova/pull/599) seem to be causing a borrow checker error here. It looks like the ownership of arguments is being moved to construct, which is likely the cause of the issue.

Would it be necessary to add Clone to ArgumentList in order to resolve this...??

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1160,120 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
...
+            TypedArray::BigUint64Array(_) => {
+                typed_array_length::<u64>(agent, &ta_record, gc.nogc())
+            }
+            #[cfg(feature = "proposal-float16array")]
+            TypedArray::Float16Array(_) => typed_array_length::<f16>(agent, &ta_record, gc.nogc()),
+            TypedArray::Float32Array(_) => typed_array_length::<f32>(agent, &ta_record, gc.nogc()),
+            TypedArray::Float64Array(_) => typed_array_length::<f64>(agent, &ta_record, gc.nogc()),
+        } as i64;
+        // c. If length < ℝ(argumentList[0]), throw a TypeError exception.
+        let arguments_list = arguments[0]
```
aapoalas, line 1263: **issue:** Huh, `arguments` should've been bound to `gc.nogc()` and accessing it here should be a compile error... The above `is_number` check is technically okay since that check only depends on the number-ness of the Value and not its actual contents (so GC doesn't matter), but this here will indeed cause a problem since if `arguments[0]` points to a heap number, GC triggered by the `Construct` call above will have invalidated that heap number and this will panic or find a different value than it originally pointed to.

But! We can solve this in a fairly neat fashion: the whole work on `argumentsList[0]` can be done before the `Construct` call. After all, we are in this if-branch only if `arguments[0]` is a number, and if it is a number then we can convert it into an `i64` pretty much freely (NaN and Infinities cause some headache here but it's probably solvable). Something like this:

```rust
let first_arg = if arguments.len() == 1 && arguments[0].is_number() {
  Some(Number::try_from(arguments[0]).unwrap()).to_i64(agent))
} else {
  None
};
```

So we get an `Option<i64>` which is `Some` if we should enter this if branch, and nothing in calculating this `Option<i64>` value can call into JS, trigger GC, or allocate onto the heap. All perfectly safe and valid stuff to move around to wherever we want.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -112,13 +115,53 @@ impl TypedArrayIntrinsicObject {
...
+                gc.nogc(),
+            ));
+        };
+        // 4. Let newObj be ? TypedArrayCreateFromConstructor(C, « 𝔽(len) »).
+        let len = len.to_i64().unwrap();
+        let len_value = Value::try_from(len).unwrap();
+        let new_obj = typed_array_create_from_constructor(
+            agent,
+            c,
+            ArgumentsList(&[len_value]),
```
aapoalas, line 143: **issue:** Since we actually know statically the type of arguments list we call with, it seems much better to create a `typed_array_create_from_constructor_with_length` variant that doesn't take `ArgumentsList` at all but instead takes `length: Option<i64>` directly, which you call here with `Some(len)`. The original function you created can then be turned into a function that just does the `Option<i64>` extraction before calling into our `typed_array_create_from_constructor_with_length` variant.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -112,13 +115,53 @@ impl TypedArrayIntrinsicObject {
...
+            gc.reborrow(),
+        )?;
+        // 5. Let k be 0.
+        // 6. Repeat, while k < len,
+        for (k, &k_value) in arguments.into_iter().enumerate() {
```
aapoalas, line 148: **note:** If the `TypedArrayCreateFromConstructor` call above, or one of the `Set` calls inside this loop trigger GC then the `arguments` becomes use-after-free. I've code now in main to fix this issue mostly ergonomically for the `TypedArrayCreateFromConstructor` call but I need to add some more support before iterating over the `arguments` can be done in a reasonable way.

For now this is fine.

tests/expectations.json
```diff
@@ -9709,20 +9704,9 @@
   "built-ins/TypedArrayConstructors/internals/Set/key-is-minus-zero.js": "FAIL",
   "built-ins/TypedArrayConstructors/internals/Set/key-is-not-integer.js": "FAIL",
   "built-ins/TypedArrayConstructors/internals/Set/tonumber-value-throws.js": "FAIL",
-  "built-ins/TypedArrayConstructors/of/BigInt/argument-number-value-throws.js": "FAIL",
-  "built-ins/TypedArrayConstructors/of/BigInt/custom-ctor-returns-other-instance.js": "FAIL",
-  "built-ins/TypedArrayConstructors/of/BigInt/custom-ctor.js": "FAIL",
   "built-ins/TypedArrayConstructors/of/BigInt/new-instance-empty.js": "FAIL",
```
aapoalas, line 9707: **question:** Do you know why these three tests fail? Based on the name only, it sounds like they should maybe pass.

aapoalas (changes requested):
Thank you, and great work!

Found some issues around GC, and we can split the method up a bit to make it cleaner in a sense. (Oh, we could also theoretically do a fast-path for when all the parameters given are numbers or bigints, depending on the TypedArray kind; in those cases we could probably mostly skip the whole `Set` internal method. But that's involved and somewhat error prone, so let's leave that for another day! :D )

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
aapoalas: Oh, but the problem can be solved by pre-computing the `Option<i64>` that I suggested in a review comment somewhere here.

tests/expectations.json
yossydev:
```
thread 'main' panicked at nova_vm/src/ecmascript/builtins/structured_data/json_object.rs:181:9:
not yet implemented
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It looks like JSON.stringify is marked as todo!

aapoalas: Roger!

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
yossydev: Thank you!

As you suggested, defining this at the beginning:

```rs
let first_arg = if arguments.len() == 1 && arguments[0].is_number() {
  Some(Number::try_from(arguments[0]).unwrap().to_i64(agent))
} else {
  None
};
```

worked perfectly…!! 🚀

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
yossydev: Thank you!

As you suggested, defining this at the beginning:

```rs
let first_arg = if arguments.len() == 1 && arguments[0].is_number() {
  Some(Number::try_from(arguments[0]).unwrap().to_i64(agent))
} else {
  None
};
```

worked perfectly…!! 🚀

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Thank you!
I think I might have fixed it with the following! https://github.com/trynova/nova/pull/601/commits/cb49c959495870e0b9064e2e304d693be24060cc

When defining newTypedArray, I’m unwrapping length before passing it to ArgumentsList, but I feel like that might not be a good idea… What do you think?

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
yossydev: thank you! I fixed it! https://github.com/trynova/nova/pull/601/commits/f988f37db43157472f836d8450d324dbb6f136e8

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
yossydev: thank you comments for me!
I fixed this review by https://github.com/trynova/nova/pull/601/commits/6fe16159383f57395a75cfea9a6485e8ca1d97b3

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1158,126 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
...
+        first_arg,
+        gc.reborrow(),
+    )
+}
+
+pub(crate) fn typed_array_create_from_constructor_with_length<'a>(
+    agent: &mut Agent,
+    constructor: Function,
+    length: Option<i64>,
+    mut gc: GcScope,
```
aapoalas, line 1191: **issue:** Missing `GcScope<'a, '_>` to bind the `'a` lifetime to GcScope's GC token borrow.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1158,126 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
...
+) -> JsResult<TypedArray<'a>> {
+    let constructor = constructor.bind(gc.nogc());
+    // 1. Let newTypedArray be ? Construct(constructor, argumentList).
+    let new_typed_array = construct(
+        agent,
+        constructor.unbind(),
+        Some(ArgumentsList::from_mut_slice(&mut [Value::try_from(
+            length.unwrap(),
```
aapoalas, line 1199: **issue:** Yeah, this unwrap isn't going to stand the test of time and JavaScript. Right now it works because we always call with `Some(i64)` but later calls coming from other methods may call in with `None`.

This should instead be ... Oh... eff... Annoying. We need to pass in the whole argumentsList :/ A quick look at the spec reveals that the two possible call paths are that either we call into here with `<< F(len) >>` or with `<< buffer, F(offset), Option<F(len)> >>`... So I guess we actually need to split the method into three parts:

1. `pub(crate) fn typed_array_create_from_constructor_with_length` as it is now but with the length as just `i64` (or maybe `u64` really, but that's a side quest that I need to get into one of these days). And it needs to perform the `Construct` call before calling into `typed_array_create_from_constructor_internal` with `Some(length)` as a parameter.
2. `pub(crate) fn typed_array_create_from_constructor_with_buffer` which is a copy of `typed_array_create_from_constructor_with_length` but takes as parameters an `ArrayBuffer`, an `i64` byte offset and an `Option<i64>` length. It needs to perform the `Construct` call with the `ArrayBuffer` and byte offset as parameters, and if the length is Some then that as a third parameter (note: this requires an if else branch to create a slice of two or three Values depending on the length Option). Then it calls into `typed_array_create_from_constructor_internal` with
3. `fn typed_array_create_from_constructor_internal` (note, no pub so private) which includes steps 2 onwards.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1158,126 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
...
+        None,
+        gc.reborrow(),
+    )?;
+    // 2. Let taRecord be ? ValidateTypedArray(newTypedArray, seq-cst).
+    let new_typed_array = new_typed_array.unbind();
+    let gc = gc.into_nogc();
+    let ta_record =
+        validate_typed_array(agent, new_typed_array.into_value(), Ordering::SeqCst, gc)?;
+    let o = ta_record.object;
+    let scoped_o = o.scope(agent, gc);
```
aapoalas, line 1211: **issue:** There is no need to perform scoping here: We're in NoGcScope land after the `gc.into_nogc()` call and thus the borrow checker should statically guarantee that `o` does not invalidate before the end of this function.

aapoalas (changes requested):
Damn, sorry I lead you astray a bit.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
yossydev: Thank you! Indeed, it wasn’t necessary!
I fixed it! https://github.com/trynova/nova/pull/601/commits/a4535d2ca7770ca7322b173623857d5c664e7037

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
yossydev: Thank you!

> or with << buffer, F(offset), Option<F(len)> >>…

Indeed, that possibility seems likely in cases like [%TypedArray%.prototype.subarray](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%25typedarray%25.prototype.subarray) 👀

I’ve revised it again! Does it now align with the intended behavior…? https://github.com/trynova/nova/pull/601/commits/bd3f32c7d5d33c275ea4531a483749cd145ffaf2

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -68,7 +71,7 @@ struct TypedArrayOf;
 impl Builtin for TypedArrayOf {
     const BEHAVIOUR: Behaviour = Behaviour::Regular(TypedArrayIntrinsicObject::of);
     const LENGTH: u8 = 0;
-    const NAME: String<'static> = BUILTIN_STRING_MEMORY.fromCodePoint;
```
aapoalas, line 71: **praise:** Oh wow, good catch! Damn!

aapoalas (changes requested):
Still a bit smoothing of edges to do :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1158,146 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
...
+pub(crate) fn typed_array_create_from_constructor_with_buffer<'a>(
+    agent: &mut Agent,
+    constructor: Function,
+    array_buffer: ArrayBuffer,
+    byte_offset: i64,
+    length: Option<i64>,
+    mut gc: GcScope<'a, '_>,
+) -> JsResult<TypedArray<'a>> {
+    let constructor = constructor.bind(gc.nogc());
+    let mut args = Vec::with_capacity(3);
```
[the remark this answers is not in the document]
yossydev: I actually tried doing that at first, but ran into a type error because the number of elements returned in the if and else branches was different:

```rs
error[E0308]: `if` and `else` have incompatible types
    --> nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs:1298:9
     |
1291 |         let args = if let Some(len) = length {
     |  __________________-
1292 | | /         &mut [
1293 | | |             array_buffer.into_value(),
1294 | | |             Value::try_from(byte_offset).unwrap(),
1295 | | |             Value::try_from(len).unwrap(),
1296 | | |         ]
     | | |_________- expected because of this
1297 | |       } else {
1298 | | /         &mut [
1299 | | |             array_buffer.into_value(),
1300 | | |             Value::try_from(byte_offset).unwrap(),
1301 | | |         ]
     | | |_________^ expected an array with a size of 3, found one with a size of 2
1302 | |       };
     | |_______- `if` and `else` have incompatible types
     |
     = note: expected mutable reference `&mut [language::value::Value<'_>; 3]`
                found mutable reference `&mut [language::value::Value<'_>; 2]`
```

So I ended up using a Vec, but yeah, heap allocation isn’t ideal.

From what I’ve seen, SmallVec looks like a good fit here.
What do you think? 👀
Or is there an even better way to handle this?

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1158,146 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
...
+/// and returns either a normal completion containing a TypedArray or a throw completion.
+/// It is used to specify the creation of a new TypedArray using a constructor function.
+fn typed_array_create_from_constructor_internal<'a>(
+    agent: &mut Agent,
+    new_typed_array: Object<'_>,
+    length: Option<i64>,
+    gc: NoGcScope<'a, '_>,
+) -> JsResult<TypedArray<'a>> {
+    // 2. Let taRecord be ? ValidateTypedArray(newTypedArray, seq-cst).
+    let new_typed_array = new_typed_array.unbind();
```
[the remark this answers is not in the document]
yossydev: thank you! I fixed and add comment!
https://github.com/trynova/nova/pull/601/commits/598723a9fc97ee1a6140882cb0b199ec2b237ade

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1158,146 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
     // 9. Return unused.
     Ok(())
 }
+
+/// ### [23.2.4.2 TypedArrayCreateFromConstructor ( constructor, argumentList )](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-typedarraycreatefromconstructor)
```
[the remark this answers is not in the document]
yossydev: thank you! I fixed and add comment!
https://github.com/trynova/nova/pull/601/commits/598723a9fc97ee1a6140882cb0b199ec2b237ade

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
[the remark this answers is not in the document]
yossydev: thank you! I fixed! https://github.com/trynova/nova/pull/601/commits/a4fddc8c0054f0f979ca2ab6223926a6de355630

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1158,3 +1158,146 @@ pub(crate) fn allocate_typed_array_buffer<T: Viewable>(
...
+        None,
+        gc.reborrow(),
+    )?;
+    typed_array_create_from_constructor_internal(
+        agent,
+        new_typed_array.unbind(),
+        Some(length),
+        gc.nogc(),
+    )
+    .unbind()
```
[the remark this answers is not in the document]
yossydev: thank you! Ah, I see—so I should’ve just used into_nogc.
https://github.com/trynova/nova/pull/601/commits/46c46cb93a0c9709a1cfed43697f7c9121fd91ac

load1n9 (dismissed):
LGTM :) but remove the index.js file before merging

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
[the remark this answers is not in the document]
aapoalas: Yeah; apparently it required a manual type annotation but `let args: &mut [Value] = if {} else {};` makes it work no problem.

aapoalas (approved):

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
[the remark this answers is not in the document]
yossydev: Ahhh, it was a problem because it was being inferred…
Thanks for fixing it!

PR #603, opened 2025-03-26
feat(ecmascript): Array.prototype.toSpliced

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.tospliced
ref: https://github.com/trynova/nova/issues/144

I think Rust has a native method called splice, so using that might make it faster.
https://doc.rust-lang.org/std/vec/struct.Splice.html

However, I couldn’t come up with a good implementation approach, so for now I just made something that works.


aapoalas (approved):
LGTM! I had such a minor nit to pick that I'll just ignore that and merge this as is!

Implementing a "fast path" for this method _would_ be possible, but it's definitely pretty hard. The `Splice` wouldn't really come into it, as we'd optimally want to avoid all Vecs in their entirety.

The way it'd work is something like this:
1. We check that we have integer or undefined start and end values, and a dense & trivial Array as `this_value`. Also probably need to check that the integers go within the Array's length now that I think about it.
2. Because we've checked the above conditions, we can now say that the source is an Array of N densely packed Values, and we'll be copying them to M densely packed Values where `M = N - removal + addition`. So, we create a new Array of length `M`.
3. Now, we need to use some nice APIs that we don't necessarily have yet to get a `&mut` slice of the new Array's `M` items (or the backing storage of them anyway), and a `&` slice of the `this_value` Array. Then we perform the copies from here to there, removing and adding as needed.

PR #604, opened 2025-03-26
feat(ecmascript): %TypedArray%.from

yossydev:
doc: https://github.com/trynova/nova/issues/145
ref: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.from


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -97,13 +101,181 @@ impl TypedArrayIntrinsicObject {
...
-        todo!();
+        let this_value = this_value.bind(gc.nogc());
+        let source = arguments.get(0).bind(gc.nogc());
+        let mapper = arguments.get(1).bind(gc.nogc());
+        let this_arg = arguments.get(2).bind(gc.nogc());
+
+        // 1. Let C be the this value.
+        let c = this_value;
+        // 2. If IsConstructor(C) is false, throw a TypeError exception.
+        let Some(c) = is_constructor(agent, c) else {
```
aapoalas, line 119: **thought:** Here we'd have options for doing a fast path. The first thing would be to have a quick-check kind of thing for if `this_value` is indeed one of the `TypedArray` constructors. Then, if there is no mapper and the source is an dense & trivial Array with the default iterator, or a TypedArray with the default iterator, then we could read their data directly and do stuff based on that... Well, except that Array still wouldn't work because it might require conversion calls...

Yeah, another TA would work (if it has the default iterator) in which case it would basically be mostly equivalent to the `TA.prototype.set` method. An Array would only work if it is dense, trivial, has the default iterator, and only contains numbers or bigints depending on the `this_value` TA type.

aapoalas (dismissed):
LGTM! Stellar work, you're really cranking these out now! <3

PR #605, opened 2025-03-27
feat(ecmascript): %TypedArray%.prototype.copyWithin

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.copywithin


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -616,13 +618,286 @@ impl TypedArrayPrototype {
...
+        let start = arguments.get(1).bind(gc.nogc());
+        let end = if arguments.len() >= 3 {
+            Some(arguments.get(2).bind(gc.nogc()))
+        } else {
+            None
+        };
+        // 1. Let O be the this value.
+        let o = this_value;
+
+        let end = end.map(|e| e.scope(agent, gc.nogc()));
```
aapoalas, line 641: **nitpick:** This should work without the map as well, the `Scopable` trait is implemented for `Option<T: Scopable>` as well.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -616,13 +618,286 @@ impl TypedArrayPrototype {
...
+            | TypedArray::BigUint64Array(_)
+            | TypedArray::Float64Array(_) => {
+                typed_array_length::<u64>(agent, &ta_record, gc.nogc())
+            }
+        }
+        .to_i64()
+        .unwrap();
+        // 4. Let relativeTarget be ? ToIntegerOrInfinity(target).
+        let relative_target =
+            to_integer_or_infinity(agent, unsafe { target.take(agent) }, gc.reborrow())?;
```
aapoalas, line 676: **issue:** Missing safety comment. (Even though it's always the same comment :) )

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -616,13 +618,286 @@ impl TypedArrayPrototype {
...
+        let o = this_value;
+
+        let end = end.map(|e| e.scope(agent, gc.nogc()));
+        let start = start.scope(agent, gc.nogc());
+        let target = target.scope(agent, gc.nogc());
+
+        // 2. Let taRecord be ? ValidateTypedArray(O, seq-cst).
+        let mut ta_record = validate_typed_array(agent, o, Ordering::SeqCst, gc.nogc())?;
+        let o = ta_record.object;
+        let scoped_o = o.scope(agent, gc.nogc());
```
aapoalas, line 648: **note:** Because o is scoped after all the others, it ends up at the "top" of the scoping stack. That means that all of the `take` calls below will fail to "pop" items off the scoping stack since they'd be effectively invalidating `scoped_o`.

If you move scoping of end, start, and target to happen only after this scope call then they'll get to pop stuff off the stack.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -616,13 +618,286 @@ impl TypedArrayPrototype {
...
+            } else if relative_end.is_negative() {
+                // 14. Else if relativeEnd < 0, let endIndex be max(len + relativeEnd, 0).
+                (len + relative_end.into_i64()).max(0)
+            } else {
+                // 15. Else, let endIndex be min(relativeEnd, len).
+                relative_end.into_i64().min(len)
+            }
+        };
+        // 16. Let count be min(endIndex - startIndex, len - targetIndex).
+        let count = (end_index - start_index).min(len - target_index);
```
aapoalas, line 718: **issue:** Everything after this step is nogc: You can use `gc.into_nogc()` here to signify that, do `let o = scoped_o.take(agent).bind(gc);` to get the TypedArray once and it will be valid until the end of this call.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -616,13 +618,286 @@ impl TypedArrayPrototype {
...
+        // 17. If count > 0, then
+        if count > 0 {
+            // a. NOTE: The copying must be performed in a manner that preserves the bit-level encoding of the source data.
+            // b. Let buffer be O.[[ViewedArrayBuffer]].
+            let buffer = scoped_o
```
aapoalas, line 723: **issue:** Probably this entire if contents could and should be a `Viewable` generic function.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -616,13 +618,286 @@ impl TypedArrayPrototype {
...
+                        from_byte_index + count_bytes - 1,
+                        to_byte_index + count_bytes - 1,
+                    )
+                } else {
+                    // m. Else,
+                    // i. Let direction be 1.
+                    (1, from_byte_index, to_byte_index)
+                };
+            // n. Repeat, while countBytes > 0,
+            while count_bytes > 0 {
```
aapoalas, line 862: **issue:** In the `Viewable` generic function, this entire loop should then be just getting the TypedArray mutable slice like we've already done a few times elsewhere, and use Rust's own `copy_within`: https://doc.rust-lang.org/stable/std/primitive.slice.html#method.copy_within

aapoalas (changes requested):
Looks really good from the basic implementation wise, but we should use TypedArray's special circumstances to get faster code :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Thanks for your comment! I did struggle a bit when handling resize, but it seems to be working now—would you mind taking a look?
As always, I’m amazed at how much faster it runs before/after—it’s seriously impressive! 🥳 https://github.com/trynova/nova/pull/605/commits/cef65216baea18980df95416d116e09aab78acf5

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Something like this should work, right? 👀

```rs
let end = map.scope(agent, gc.nogc());
```

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2450,3 +2680,66 @@ fn reverse_typed_array<T: Viewable + Copy + std::fmt::Debug>(
...
+            "TypedArray is not properly aligned",
+            gc,
+        ));
+    }
+    let slice = &mut slice[..len];
+    let start_bound = start_index as usize;
+    let target_index = target_index as usize;
+    let before_len = before_len as usize;
+    if before_len != slice.len() {
+        let end_bound = (len - target_index).max(0).min(before_len - target_index);
```
aapoalas, line 2737: **nitpick:** IIRC there's a `clamp` function that can do max+min directly.

aapoalas (approved):
LGTM! Great work! <3

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
aapoalas: I thought it should but apparently I'm misremembering and I couldn't implement `Scopable for Option<T: Scopable>`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
aapoalas: Yeah, the "naive loop" is basically filled to the gills with all kinds of extra checks, balances, and conversions to and from this value to that and back again.

A proper `copy_within` just moves some bytes around.

PR #606, opened 2025-03-28
chore(ecmascript): remove TypedArray toSpliced

yossydev:
It looks like toSpliced and splice don’t exist on TypedArray!
There was no splice, but toSpliced did exist, so I removed it!

ref: https://github.com/tc39/proposal-change-array-by-copy/issues/88


aapoalas (approved):
Oops :D

PR #607, opened 2025-03-29
feat(ecmascript): %TypedArray%.prototype.fill

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.fill

(I haven’t looked at it properly yet, but) I suppose this could also be optimized by using [Rust’s standard fill](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.fill), right? 👀


aapoalas:
> (I haven’t looked at it properly yet, but) I suppose this could also be optimized by using [Rust’s standard fill](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.fill), right? 👀

Yup, exactly! :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -893,12 +893,462 @@ impl TypedArrayPrototype {
     }
 
     fn fill<'gc>(
-        _agent: &mut Agent,
-        _this_value: Value,
-        _: ArgumentsList,
-        _gc: GcScope<'gc, '_>,
+        agent: &mut Agent,
```
aapoalas, line 896: **nitpick:** Missing the function documentation string: I use it to quickly ascertain how many arguments are needed etc.

tests/expectations.json
```diff
@@ -8754,53 +8754,7 @@
...
-  "built-ins/TypedArray/prototype/fill/absent-indices-computed-from-initial-length.js": "CRASH",
-  "built-ins/TypedArray/prototype/fill/coerced-end-detach.js": "CRASH",
-  "built-ins/TypedArray/prototype/fill/coerced-indexes.js": "CRASH",
-  "built-ins/TypedArray/prototype/fill/coerced-start-detach.js": "CRASH",
-  "built-ins/TypedArray/prototype/fill/coerced-value-detach.js": "CRASH",
-  "built-ins/TypedArray/prototype/fill/coerced-value-start-end-resize.js": "CRASH",
-  "built-ins/TypedArray/prototype/fill/detached-buffer.js": "CRASH",
-  "built-ins/TypedArray/prototype/fill/fill-values-conversion-once.js": "CRASH",
-  "built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js": "CRASH",
   "built-ins/TypedArray/prototype/fill/fill-values-conversion-operations.js": "CRASH",
```
yossydev, line 8757: In this test, it seems that when 2147483647 is passed to fill, it needs to become 2147483648, but I’m wondering if the issue might be with the use of v.into_value()… 👀

https://github.com/trynova/nova/blob/115c527d70ed368b12ee2d90598e1aead25f72c5/nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs#L1171-L1447

yossydev (commented):
The CI is failing, but I didn’t have any issues locally…

tests/expectations.json
aapoalas: Alright; us using SmallInteger both for safe integers and small bigints has bit us in the ass here. We're converting into a number but we use the SmallInteger conversion which effectively tries to convert into a small bigint. That conversion succeeds, even though a safe integer conversion wouldn't.

aapoalas:
I refactored the code a little, moving most of the work into the helper function and cleaning up repeated parts.

I'm starting to think something like this is what we should actually do for _all_ the prototype methods; we need to hide away the `TypedArrayIndex` from the public API anyway, so something like `TypedArrayBase<T: Viewable>` might be in the cards, which would then implement the prototype methods for each `T: Viewable`.

yossydev:
Thank you! Yeah, the code gets pretty long and it’s tough to follow sometimes…👀

so something like TypedArrayBase<T: Viewable> might be in the cards, which would then implement the prototype methods for each T: Viewable.

Sounds good! I think that would really help clean up the code in each file — right now, we’re writing everything from i8 to f64 inline, so each method gets pretty long.
I’ll try experimenting with that approach in a few other places 👀

tests/expectations.json
yossydev: So that’s what it was all about…

aapoalas: Alright, I spawned #613 based on this.

aapoalas:
> Thank you! Yeah, the code gets pretty long and it’s tough to follow sometimes…👀
> 
> so something like TypedArrayBase<T: Viewable> might be in the cards, which would then implement the prototype methods for each T: Viewable.
> 
> Sounds good! I think that would really help clean up the code in each file — right now, we’re writing everything from i8 to f64 inline, so each method gets pretty long. I’ll try experimenting with that approach in a few other places 👀

Sounds good to me! And the code here looks good to me, so let's get it merged. Thank you as always, and great work <3

Sorry it took me so wrong to get around to reviewing this and the other code, my evenings are pretty busy with some upcoming choir performances.

aapoalas (approved):

PR #609, opened 2025-03-30
feat(ecmascript): %TypedArray%.prototype.reduce

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.reduce

I want to check whether `iter.reduce` can be used. 
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.reduce

↑ Since JavaScript’s reduce takes a callback, if I want to use Rust’s reduce, I think I need to pass that callback as-is.
But since the internal logic of the callback is unknown, it might be quite tricky… right?


aapoalas:
Any JavaScript method that takes a callback basically cannot be represented as Rust's reduce because JS's callbacks can mutate the memory they iterate over, whereas Rust's must not.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2210,13 +2210,107 @@ impl TypedArrayPrototype {
...
-        _this_value: Value,
-        _: ArgumentsList,
-        _gc: GcScope<'gc, '_>,
+        agent: &mut Agent,
+        this_value: Value,
+        arguments: ArgumentsList,
+        mut gc: GcScope<'gc, '_>,
     ) -> JsResult<Value<'gc>> {
-        todo!()
+        let callback = arguments.get(0).bind(gc.nogc()).scope(agent, gc.nogc());
```
aapoalas, line 2222: **nitpick:** You could hold off on scoping callback and initial value until before the loop, where you scope `o` as well.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2210,13 +2210,107 @@ impl TypedArrayPrototype {
...
+        } as i64;
+        // 4. If IsCallable(callback) is false, throw a TypeError exception.
+        let Some(stack_callback_fn) = is_callable(callback.get(agent), gc.nogc()) else {
+            return Err(agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "Callback is not callable",
+                gc.into_nogc(),
+            ));
+        };
+        let scoped_callback = unsafe { callback.replace_self(agent, stack_callback_fn.unbind()) };
```
aapoalas, line 2261: **nitpick (blocking):** Missing safety comment.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2210,13 +2210,107 @@ impl TypedArrayPrototype {
...
+        //    a. Set accumulator to initialValue.
+        let no_initial_value = initial_value.is_none();
+        let mut accumulator = initial_value.unwrap_or(Value::Undefined.scope_static(gc.nogc()));
+        let scoped_o = o.scope(agent, gc.nogc());
+        // 9. Else,
+        if no_initial_value {
+            // a. Let Pk be ! ToString(𝔽(k)).
+            let pk = PropertyKey::try_from(k).unwrap();
+            // b. Set accumulator to ! Get(O, Pk).
+            let result = unwrap_try(try_get(agent, scoped_o.get(agent), pk, gc.nogc()));
```
aapoalas, line 2283: **praise:** Nicely handled! <3

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2210,13 +2210,107 @@ impl TypedArrayPrototype {
...
+                Value::Undefined,
+                Some(ArgumentsList::from_mut_slice(&mut [
+                    accumulator.get(agent),
+                    k_value.unbind(),
+                    Number::from(k_int).into_value(),
+                    scoped_o.get(agent).into_value(),
+                ])),
+                gc.reborrow(),
+            )?;
+            unsafe { accumulator.replace(agent, result.unbind()) };
```
aapoalas, line 2308: **nitpick (blocking):** Missing safety comment.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2210,13 +2210,107 @@ impl TypedArrayPrototype {
...
+        // 6. Let k be 0.
+        let mut k = 0;
+        // 7. Let accumulator be undefined.
+        // 8. If initialValue is present, then
+        //    a. Set accumulator to initialValue.
+        let no_initial_value = initial_value.is_none();
+        let mut accumulator = initial_value.unwrap_or(Value::Undefined.scope_static(gc.nogc()));
+        let scoped_o = o.scope(agent, gc.nogc());
+        // 9. Else,
+        if no_initial_value {
```
aapoalas, line 2279: **thought:** Not in this PR, but eventually we'll probably want to implement an optimisation for TypedArray loops where we scope the backing ArrayBuffer and store it together with the backing pointer + Viewable slice length locally in a special struct. We then implement ... probably an iterator method on the struct that checks that the AB hasn't been detached and  that the index we're accessing is still within the buffer (the AB hasn't been resized to be smaller), then it'll read the next Viewable through the backing pointer and give that as `Some(Viewable)`, otherwise it'll give `None`.

I expect that'd make the loops even more performant :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2210,13 +2210,107 @@ impl TypedArrayPrototype {
...
+                gc.nogc(),
+            ));
+        };
+        // 6. Let k be 0.
+        let mut k = 0;
+        // 7. Let accumulator be undefined.
+        // 8. If initialValue is present, then
+        //    a. Set accumulator to initialValue.
+        let no_initial_value = initial_value.is_none();
+        let mut accumulator = initial_value.unwrap_or(Value::Undefined.scope_static(gc.nogc()));
```
aapoalas, line 2276: **nitpick:** This could be done something like:
```rust
let mut accumulator = let Some(init) = initial_value {
  init.scope(agent, gc.nogc())
} else {
  // 9. Else
  // ...
  result.scope(agent, gc.nogc())
};
```

aapoalas (changes requested):
Some minor codestyle / safety comment issues, but generally looks good to me! Stellar work <3

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2324,13 +2324,113 @@ impl TypedArrayPrototype {
...
+            let pk = PropertyKey::try_from(k).unwrap();
+            // b. Set accumulator to ! Get(O, Pk).
+            let result = unwrap_try(try_get(agent, o, pk, gc.nogc()));
+            // c. Set k to k + 1.
+            k += 1;
+            result.scope(agent, gc.nogc())
+        };
+        // SAFETY: scoped_callback is not shared.
+        let scoped_callback = unsafe {
+            callback
```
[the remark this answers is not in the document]
yossydev: Ah, thanks! You’re right, the unsafe wasn’t necessary!
I’ve fixed it! https://github.com/trynova/nova/pull/609/commits/f991f2e3a91b523122266d6bb9f51163d7f8a7c1

aapoalas (approved):
LGTM, great work <3

PR #610, opened 2025-03-31
feat(ecmascript): %TypedArray%.prototype.sort

yossydev:
doc: 
- https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.sort
- https://tc39.es/ecma262/multipage/indexed-collections.html#sec-comparetypedarrayelements


nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -4401,6 +4417,72 @@ fn compare_array_elements(
...
+        if v.is_nan(agent) {
+            Ok(Ordering::Equal)
+        } else if v.is_sign_positive(agent) {
+            Ok(Ordering::Greater)
+        } else if v.is_sign_negative(agent) {
+            Ok(Ordering::Less)
+        } else {
+            Ok(Ordering::Equal)
+        }
+    } else if let (Value::Integer(x), Value::Integer(y)) = (x, y) {
```
aapoalas, line 4458: **thought:** We'd probably want to have a separate method that performs the fast paths directly on `&mut [T: Viewable]` when comparator is None. In that call we'd skip the `sort_indexed_properties` call entirely and just `sort` the slice directly.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -4296,7 +4296,23 @@ fn sort_indexed_properties<'scope, const SKIP_HOLES: bool, const TYPED_ARRAY: bo
     // performing any further calls to SortCompare and return that Completion
     // Record.
     if TYPED_ARRAY {
-        items.sort_by(|_a, _b| compare_typed_array_elements());
+        let mut error: Option<JsError> = None;
+        items.sort_by(|a, b| {
+            let result =
+                compare_typed_array_elements(agent, a, b, comparator.clone(), gc.reborrow());
```
aapoalas, line 4302: **thought:** For a TypedArray we could actually generate `items` a lot smoother by getting the typed array slice `&[T: Viewable]` and then using the `Viewable::into_le/be_value` API to get a `Numeric`, scope that and collect into a `Vec<Scoped<Numeric>>`.

That would require an entirely separate `sort_typed_array_properties` method though.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2388,13 +2388,90 @@ impl TypedArrayPrototype {
...
+        let comparator = arguments.get(0);
+        // 1. If comparator is not undefined and IsCallable(comparator) is false, throw a TypeError exception.
+        let comparator = if comparator.is_undefined() {
+            None
+        } else if let Some(comparator) = is_callable(comparator, gc.nogc()) {
+            Some(comparator.scope(agent, gc.nogc()))
+        } else {
+            return Err(agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "",
```
aapoalas, line 2412: **issue:** Missing error message.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2388,13 +2388,90 @@ impl TypedArrayPrototype {
...
-        _this_value: Value,
-        _: ArgumentsList,
-        _gc: GcScope<'gc, '_>,
+        agent: &mut Agent,
+        this_value: Value,
+        arguments: ArgumentsList,
+        mut gc: GcScope<'gc, '_>,
     ) -> JsResult<Value<'gc>> {
-        todo!();
+        let comparator = arguments.get(0);
```
aapoalas, line 2403: **nitpick:** Could bind `this_value` and `comparator` here. That way already from the first lines of the function we can see that binding/scoping of arguments has been done, and we don't need to read further down to check if everything is a-okay.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2388,13 +2388,90 @@ impl TypedArrayPrototype {
...
+                ExceptionType::TypeError,
+                "",
+                gc.nogc(),
+            ));
+        };
+        // 2. Let obj be the this value.
+        let obj = this_value;
+        // 3. Let taRecord be ? ValidateTypedArray(obj, seq-cst).
+        let ta_record = validate_typed_array(agent, obj, Ordering::SeqCst, gc.nogc())?;
+        let o = ta_record.object;
```
aapoalas, line 2420: **nitpick:** Shadow the `obj` name to match the specification text.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
aapoalas: Actually, better: We'd rather get a `let items: Vec<T: Viewable>` clone of the original slice. Then we can use `items.sort_by(comparator(a.into_le/be_value(), b.into_le/be_value()))` for the sorting, ie. the numeric values are never actually scoped by us, they're created on-demand for the comparator call. Once sorting is over, then we re-copy `items` into the source `&mut [T: Viewable]`.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
```diff
@@ -4296,7 +4296,23 @@ fn sort_indexed_properties<'scope, const SKIP_HOLES: bool, const TYPED_ARRAY: bo
...
+            if error.is_some() {
```
aapoalas, line 4303: **issue:** This should be before the `compared_typed_array_elements` call: The two "call-throws" tests are failing because of this misordering; if the comparator throws an error then we should stop calling it immediately.

aapoalas (changes requested):
One trivial issue, one more substantial but still simple issue; generally code wise looks good to me! Thank you!

There is a more substantial rewrite possible here, though: If there is no comparator then we could perform a normal Rust `sort_by` on the `&mut [T: Viewable]` directly (`sort_by` instead of a direct `sort` because floating point values require a special sorting function to deal with NaNs). Even if there is a comparator, we could avoid all the scoping etc. by cloning the TA data into a `Vec<T: Viewable>` and performing `sort_typed_array_properties<T: Viewable>` on it.

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
yossydev: Sure! Here’s the English translation of your message:

⸻

Thanks for the review!
I’m currently working on a fix here:
https://github.com/trynova/nova/pull/610/commits/8a3e36b1ec0972c37ab9e248ab7b5ea16abc5fdf

But I’m running into the following error due to double borrowing of agent:

```rs
error[E0500]: closure requires unique access to `*agent` but it is already borrowed
    --> nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs:3309:19
     |
3285 |     let byte_slice = array_buffer.as_mut_slice(agent);
     |                                                ----- borrow occurs here
...
3309 |     items.sort_by(|a, b| {
     |                   ^^^^^^ closure construction occurs here
...
3313 |         let a_value = a.into_le_value(agent, gc.nogc()).into_value();
     |                                       ----- second borrow occurs due to use of `*agent` in closure
...
3349 |     slice.copy_from_slice(&items);
     |     ----- first borrow later used here
```

It’s failing at both a.into_le_value(...) and the final copy_from_slice(...) due to borrowing of agent.
I’m unsure how to resolve this—maybe my whole approach is incorrect?

yossydev:
I’ve fixed everything except for the comment at https://github.com/trynova/nova/pull/610#discussion_r2033320967!

nova_vm/src/ecmascript/builtins/indexed_collections/array_objects/array_prototype.rs
yossydev: Thanks for the comment! It really helped!!
I’m not sure if it’s perfect, but I tried fixing it in [this commit](https://github.com/trynova/nova/pull/610/commits/bdf546b4e3b7ea4bf903c0446816c0f9343ba0b8).
Could you please take another look? 🙇‍♂️

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3230,3 +3421,178 @@ fn fill_typed_array<'a, T: Viewable>(
...
+            "TypedArray is not properly aligned",
+            gc,
+        ));
+    }
+    let slice = &mut slice[..len];
+    slice.sort_by(|a, b| a.partial_cmp(b).unwrap());
+    Ok(())
+}
+
+fn sort_total_cmp_typed_array<T: Viewable + std::fmt::Debug + TotalOrder + 'static>(
```
aapoalas, line 3460: **issue:** The `TotalOrder` trait does not implement the correct ordering required by ECMAScript.

ECMAScript requires that all NaN's are sorted to the end of the buffer, TotalOrder will sort negative NaNs to the front and positive NaNs to the end. Effectively, we'll need to do the ordering somewhat manually.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3230,3 +3421,178 @@ fn fill_typed_array<'a, T: Viewable>(
...
+        }
+    });
+    if let Some(error) = error {
+        return Err(error);
+    }
+    let array_buffer = ta.get_viewed_array_buffer(agent, gc.nogc());
+    let byte_offset = ta.byte_offset(agent);
+    let byte_length = ta.byte_length(agent);
+    let byte_slice = array_buffer.as_mut_slice(agent);
+    if byte_slice.is_empty() || len == 0 {
```
aapoalas, line 3573: **nitpick:** `len` cannot have changed to be 0 now.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3230,3 +3421,178 @@ fn fill_typed_array<'a, T: Viewable>(
...
+        let end_index = byte_offset + byte_length;
+        if end_index > byte_slice.len() {
+            return Ok(());
+        }
+        &mut byte_slice[byte_offset..end_index]
+    } else {
+        &mut byte_slice[byte_offset..]
+    };
+    let (head, slice, _) = unsafe { byte_slice.align_to_mut::<T>() };
+    if !head.is_empty() {
```
aapoalas, line 3586: **nitpick:** This also cannot be true, as the `byte_slice`'s start pointer is at this point already known to be the same as it was before. If it wasn't unaligned before, it's not unaligned now.

aapoalas (changes requested):
The native sorting is starting to look pretty great! Some problems did slip in though, let's fix those.

Also, I expect you'll have a bit of a merge conflict to fix, sorry about that :(

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: I confirmed that in Rust, total_cmp will place -f64::NAN at the beginning if it exists.
(Reference: [Rust Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=63ce2edc4717c9b481fa6e776c0b1052))
However, when I actually tried this in Nova, the sort worked correctly.
I assume this is because JavaScript’s NaN does not distinguish between positive and negative values?

```js
const arr = new Float64Array([
  Number.NaN,
  Number.POSITIVE_INFINITY,
  -3.5,
  0,
  -Number.NaN,
  Number.NEGATIVE_INFINITY,
  7,
  2,
]);
arr.sort(); // [-inf, -3.5, 0.0, 2.0, 7.0, inf, NaN, NaN]
```

So while the current implementation does produce the same behavior,
do you think it should still be corrected?
Or if I’m missing something in my understanding, please let me know 👀

yossydev:
@aapoalas 
Thanks for the review! I believe I’ve checked and fixed all the points you mentioned, so if it’s okay, could you take another look? 🙇‍♂️

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3230,3 +3421,178 @@ fn fill_typed_array<'a, T: Viewable>(
     // 20. Return O.
     Ok(ta)
 }
+
+fn sort_partial_cmp_typed_array<T: Viewable + std::fmt::Debug + PartialOrd>(
```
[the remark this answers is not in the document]
yossydev: I found out from the CI error that f16 is not included in either Ord or TotalOrd.
For now, I’ve changed the constraint to PartialOrd, but I’m wondering… 🤔
Would it be better to implement a separate sorting function specifically for f16?

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3230,3 +3421,178 @@ fn fill_typed_array<'a, T: Viewable>(
...
+        ));
+    }
+    let slice = &mut slice[..len];
+    let mut items: Vec<T> = slice.to_vec();
+    let mut error: Option<JsError> = None;
+    items.sort_by(|a, b| {
+        if error.is_some() {
+            return std::cmp::Ordering::Equal;
+        }
+        let a_val = a.into_le_value(agent, gc.nogc()).into_value().unbind();
```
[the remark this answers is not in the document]
eliassjogreen: Sidenote: A custom lint rule to detect this should be fairly easy. I might give it a shot this holiday/weekend.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
aapoalas: Yup, it does need to be fixed. In JS all NaNs are just "some NaN" and usually engines will at least partially normalise/canonicalise them so that they're all mostly the same value (if read bitwise as a u64, see eg. `f64::to_bits()`). We currently fully normalise all NaNs so assigning a NaN will into a Float64Array will always produce the same bit value.

But that's not the only way to put various NaNs into a Float64Array. You can access the array's buffer as a BigUin64Array using `new BigUintArray(arr.buffer)` and then write the u64 bit values into that memory. With this you can put any and all possible NaN values into a Float64Array and at that point the sort function will start misbehaving.

yossydev: Thank you!! Yeah, you’re right — when I wrote the bit patterns directly, the Negative NaN got messed up!

```js
const nanPatterns = [
  0x7ff8000000000001n,
  0x3ff0000000000000n,
  0x7ff0000000000000n,
  0xfff0000000000000n,
  0x0000000000000000n,
  0xfff8000000000000n,
  0xbff0000000000000n,
];

const arr = new Float64Array(nanPatterns.length);
nanPatterns.forEach((bits, i) => {
  const u64 = new BigUint64Array(arr.buffer);
  u64[i] = bits;
});

arr.sort();
```

https://github.com/trynova/nova/pull/610/commits/f67cdcd950b1232d937fc48438de641bf774a9ea ← I modified it to not use total_cmp, so I think it’s probably fixed now!

aapoalas: Hmm... Rather than using conversion to values, I think it'd be neater to just do a custom `trait CompareTypedArrayElements` or such, which has one method 'fn cmp_elements(a: Self, b: Self) -> Ordering`, and then implement that for f16, f32, and f64.

All the integers can just use the normal total ordering of `Ord` for sorting, but for floats we'd then have a separate method that takes `T: Viewable + CompareTypedArrayElements`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2834,13 +2837,154 @@ impl TypedArrayPrototype {
...
+        };
+        // 2. Let obj be the this value.
+        let obj = this_value;
+        // 3. Let taRecord be ? ValidateTypedArray(obj, seq-cst).
+        let ta_record = validate_typed_array(agent, obj, Ordering::SeqCst, nogc)
+            .unbind()?
+            .bind(nogc);
+        let obj = ta_record.object;
+        // 4. Let len be TypedArrayLength(taRecord).
+        let len = match obj {
```
aapoalas, line 2875: **thought:** We could actually move the length checking inside of the sort functions. That would avoid one extra match.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,200 @@ fn fill_typed_array<'a, T: Viewable>(
...
+}
+
+fn sort_comparator_typed_array<'a, T: Viewable + Copy + std::fmt::Debug>(
+    agent: &mut Agent,
+    ta: TypedArray,
+    len: usize,
+    comparator: Scoped<'_, Function<'static>>,
+    mut gc: GcScope<'a, '_>,
+) -> JsResult<'a, ()> {
+    let array_buffer = ta.get_viewed_array_buffer(agent, gc.nogc());
```
aapoalas, line 3631: **issue:** You've forgotten to bind `ta` here, which makes it GC-unsafe. One you do bind it, the borrow checker will yell at you about the final `ta` usage on line 3697. To get that fixed, you'd need to scope `ta` but that would be kind of redundant since you've already scoped `ta` in the caller of this function; better thus to just take as parameter a `Scoped<TypedArray>`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,200 @@ fn fill_typed_array<'a, T: Viewable>(
...
+    let copy_len = items.len().min(len);
+    slice[..copy_len].copy_from_slice(&items[..copy_len]);
+    Ok(())
+}
+
+fn sort_comparator_typed_array<'a, T: Viewable + Copy + std::fmt::Debug>(
+    agent: &mut Agent,
+    ta: TypedArray,
+    len: usize,
+    comparator: Scoped<'_, Function<'static>>,
```
aapoalas, line 3628: **nitpick:** This can be simplified as `Scoped<Function>`, turns out.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,200 @@ fn fill_typed_array<'a, T: Viewable>(
...
+            Value::Undefined,
+            Some(ArgumentsList::from_mut_slice(&mut [
+                a_val.unbind(),
+                b_val.unbind(),
+            ])),
+            gc.reborrow(),
+        )
+        .unbind()
+        .map(|v| v.to_number(agent, gc.reborrow()))
+        .and_then(|r| r);
```
aapoalas, line 3676: **thought:** This seems like an unnecessary `and_then`. Does this actually do something? `call_function` should return `JsResult<Value>`, right? And then mapping that to `v.to_number` should return `JsResult<Number>`, I'd assume...? Or does it actually become `JsResult<JsResult<Number>>`, and then the `and_then` flattens that? I think it might work the same if instead of `map(closure).and_then(id)` you just did `and_then(closure)`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,200 @@ fn fill_typed_array<'a, T: Viewable>(
...
+        } else if num.is_sign_negative(agent) {
+            std::cmp::Ordering::Less
+        } else {
+            std::cmp::Ordering::Equal
+        }
+    });
+    if let Some(error) = error {
+        return Err(error);
+    }
+    let array_buffer = ta.get_viewed_array_buffer(agent, gc.nogc());
```
aapoalas, line 3697: **note:** Here `ta` is use-after-free.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,200 @@ fn fill_typed_array<'a, T: Viewable>(
...
+    agent: &mut Agent,
+    ta: TypedArray,
+    len: usize,
+    comparator: Scoped<'_, Function<'static>>,
+    mut gc: GcScope<'a, '_>,
+) -> JsResult<'a, ()> {
+    let array_buffer = ta.get_viewed_array_buffer(agent, gc.nogc());
+    let byte_offset = ta.byte_offset(agent);
+    let byte_length = ta.byte_length(agent);
+    let byte_slice = array_buffer.as_mut_slice(agent);
```
aapoalas, line 3634: **nitpick:** All of the slice work starting here could be just plain &ref instead of &mut, so `as_slice`, `&byte_slice[..]`, `align_to::<T>()` etc, since we're only making a copy of the data here before the loop.

aapoalas (approved):
I still had some thoughts and nitpicks here and there, but this is looking really good now! Excellent work on this, and thank you <3

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2834,13 +2837,154 @@ impl TypedArrayPrototype {
...
+                }
+                TypedArray::BigInt64Array(_) => {
+                    sort_partial_cmp_typed_array::<i64>(agent, scoped_obj.get(agent), len, nogc)?
+                }
+                TypedArray::BigUint64Array(_) => {
+                    sort_partial_cmp_typed_array::<u64>(agent, scoped_obj.get(agent), len, nogc)?
+                }
+                #[cfg(feature = "proposal-float16array")]
+                TypedArray::Float16Array(_) => {
+                    sort_partial_cmp_typed_array::<f16>(agent, scoped_obj.get(agent), len, nogc)?
```
aapoalas, line 2975: **issue:** This should be using the same comparison function as f32 and f64.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2834,13 +2837,154 @@ impl TypedArrayPrototype {
...
+                    sort_partial_cmp_typed_array::<i32>(agent, scoped_obj.get(agent), len, nogc)?
+                }
+                TypedArray::Uint32Array(_) => {
+                    sort_partial_cmp_typed_array::<u32>(agent, scoped_obj.get(agent), len, nogc)?
+                }
+                TypedArray::BigInt64Array(_) => {
+                    sort_partial_cmp_typed_array::<i64>(agent, scoped_obj.get(agent), len, nogc)?
+                }
+                TypedArray::BigUint64Array(_) => {
+                    sort_partial_cmp_typed_array::<u64>(agent, scoped_obj.get(agent), len, nogc)?
```
aapoalas, line 2971: **issue:** These function names are still the wrong way around, all integers should be compared using the normal "total order" of `Ord`, and floats should be compared using the custom `ECMAScriptOrd` trait (or whatever feels like a good name).

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,200 @@ fn fill_typed_array<'a, T: Viewable>(
     // 20. Return O.
     Ok(ta)
 }
+
+fn sort_partial_cmp_typed_array<'a, T: Viewable + std::fmt::Debug + PartialOrd>(
```
aapoalas, line 3522: **issue:** Use `Ord` instead of `PartialOrd`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,200 @@ fn fill_typed_array<'a, T: Viewable>(
...
+            "TypedArray is not properly aligned",
+            gc,
+        ));
+    }
+    let slice = &mut slice[..len];
+    slice.sort_by(|a, b| a.partial_cmp(b).unwrap());
+    Ok(())
+}
+
+fn sort_total_cmp_typed_array<'a, T: Viewable + std::fmt::Debug + PartialOrd + 'static>(
```
aapoalas, line 3557: **issue:** Use a custom `ECMAScriptOrd` or something as the trait bound instead of `Ord`, and implement that trait for f16, f32, and f64.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,200 @@ fn fill_typed_array<'a, T: Viewable>(
...
+        return Err(agent.throw_exception_with_static_message(
+            ExceptionType::TypeError,
+            "TypedArray is not properly aligned",
+            gc,
+        ));
+    }
+    let slice = &mut slice[..len];
+    let mut items: Vec<T> = slice.to_vec();
+    items.sort_by(|a: &T, b: &T| {
+        let left = a.into_le_value(agent, gc).into_value();
```
aapoalas, line 3590: **issue:** Let's rely on the `ECMAScriptOrd` trait instead of using `into_le/be_value` calls, so as to avoid unnecessary conversions.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,200 @@ fn fill_typed_array<'a, T: Viewable>(
...
+        ));
+    }
+    let slice = &mut slice[..len];
+    let mut items: Vec<T> = slice.to_vec();
+    let mut error: Option<JsError> = None;
+    items.sort_by(|a, b| {
+        if error.is_some() {
+            return std::cmp::Ordering::Equal;
+        }
+        let a_val = a.into_le_value(agent, gc.nogc()).into_value();
```
aapoalas, line 3662: **nitpick (sort of blocking):** We shouldn't assume an LE machine over a BE machine, but should instead check what the default endianness of the machine architecture is.

We honestly just need an `into_ne_value` on the `Viewable` that internally is just an `if cfg(little_endian) { self.into_le_value } else { self.into_be_value }`, so that we don't need to repeat the same if-cfg checks everywhere.

aapoalas (changes requested):
Oops, there were some issues and I was kind of re-reviewing before you'd had the chance to fix my previous comments as well. Sorry :(

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: > Or does it actually become JsResult<JsResult<Number>>?

You’re exactly right — it does become `JsResult<JsResult<Number>>,` which is why I was using and_then!

>  I think it might work the same if instead of map(closure).and_then(id) you just did and_then(closure)

Thanks! Looks like I can remove the `map` and just use `and_then`!

yossydev:
Thank you for the review! Sorry for the delay!
I’ve made the changes based on all the feedback you gave me! (Apologies if I missed anything 🙇‍♂️)
https://github.com/trynova/nova/pull/610/commits/ee7ba110741f7da6a2b23c14deb40daa3eeea705

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -167,6 +168,14 @@ impl Viewable for u8 {
         Number::from(self.to_le()).into_numeric()
     }
 
+    fn into_ne_value<'a>(self, agent: &mut Agent, gc: NoGcScope<'a, '_>) -> Numeric<'a> {
```
aapoalas, line 171: **nitpick:** This can be done as a trait default impl, saving a lot of repetition. I'll fix this.

aapoalas (changes requested):
I'll fix a few things quick snap, but aside from those this looks good to go now! Nice work, and thanks for sticking it out despite both of us being busy :) <3

aapoalas (approved):

PR #660, opened 2025-04-10
feat(ecmascript): %TypedArray%.prototype.filter

yossydev:
Based on what I’m seeing from the failing parts of test262, it seems like I might also need to make changes to some of the existing methods. (Maybe?)

doc: 

- https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.filter
- https://tc39.es/ecma262/multipage/indexed-collections.html#typedarray-species-create
- https://tc39.es/ecma262/multipage/abstract-operations.html#sec-speciesconstructor


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1246,13 +1246,170 @@ impl TypedArrayPrototype {
...
+        // 5. Let kept be a new empty List.
+        let mut kept = Vec::with_capacity(len.try_into().unwrap());
+        // 6. Let captured be 0.
+        let mut captured = 0;
+        // 7. Let k be 0.
+        let mut k = 0;
+        // 8. Repeat, while k < len,
+        while k < len {
+            // a. Let Pk be ! ToString(𝔽(k)).
+            let pk = PropertyKey::from(SmallInteger::from(k as u32));
```
aapoalas, line 1310: issue: `k` can be larger than u32, but never larger than i53 which.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1246,13 +1246,170 @@ impl TypedArrayPrototype {
...
+            // b. Let kValue be ! Get(O, Pk).
+            let k_value = unwrap_try(try_get(agent, scoped_o.get(agent), pk, gc.nogc()))
```
aapoalas, line 1312: **thought:** Here again we'd prefer to have pretty much this entire method (after the callable check) be specialized by `T: Viewable`, so we could have `kept` be a `Vec<T>` instead of `Vec<Scoped<Value>>`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1246,13 +1246,170 @@ impl TypedArrayPrototype {
...
+                ])),
+                gc.reborrow(),
+            )?;
+            let selected = to_boolean(agent, call);
+            // d. If selected is true, then
+            if selected {
+                //  i. Append kValue to kept.
+                kept.push(k_value.get(agent).unbind());
+                //  ii. Set captured to captured + 1.
+                captured += 1;
```
aapoalas, line 1332: **nitpick:** You don't need to keep count manually, captured is always equal to `kept.len()`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1246,13 +1246,170 @@ impl TypedArrayPrototype {
...
+            }
+            // e. Set k to k + 1.
+            k += 1;
+        }
+        // 9. Let A be ? TypedArraySpeciesCreate(O, « 𝔽(captured) »).
+        let o = scoped_o.get(agent);
+        let a = match o {
+            TypedArray::Int8Array(_) => {
+                typed_array_species_create_with_length::<i8>(agent, o, captured, gc.reborrow())?
+                    .unbind()
```
aapoalas, line 1342: **issue:** Don't forget to rebind.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1246,13 +1246,170 @@ impl TypedArrayPrototype {
...
+            }
+            TypedArray::Float64Array(_) => {
+                typed_array_species_create_with_length::<f64>(agent, o, captured, gc.reborrow())?
+                    .unbind()
+            }
+        };
+        // 10. Let n be 0.
+        // 11. For each element e of kept, do
+        let scoped_a = a.scope(agent, gc.nogc());
```
aapoalas, line 1397: **issue:** Scoping here is unnnecessary.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1246,13 +1246,170 @@ impl TypedArrayPrototype {
...
+            // a. Perform ! Set(A, ! ToString(𝔽(n)), e, true).
+            // b. Set n to n + 1.
+            unwrap_try(try_set(
+                agent,
+                scoped_a.get(agent).into_object(),
+                n.try_into().unwrap(),
+                *e,
+                true,
+                gc.nogc(),
+            ))?;
```
aapoalas, line 1408: **issue:** Spec uses `!` to indicate that this will never throw; you can safely unwrap here.

aapoalas (changes requested):
Found some issues that need to be fixed before merging.

Generally looks good, though this method would also benefit from being specialised on `T: Viewable`; that way we could avoid much of the "generic" stuff and extra heap allocations etc. (There's actually also a fun future optimisation where we don't put the kept items in a `Vec<Viewable>` but instead allocate a `DataBlock` directly and assign into it, then finally resize it to the desired size and create the backing ArrayBuffer with that DataBlock as its buffer. But we don't have the APIs for that and it's all in all a much more involved change, so not today :)

yossydev:
@aapoalas 
Thanks for the review!
I believe I’ve addressed all the points you mentioned, so if you have a moment, could you please take another look when you get a chance? 🙇‍♂️

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3230,3 +3372,85 @@ fn fill_typed_array<'a, T: Viewable>(
...
+    let callback = callback.scope(agent, gc.nogc());
+    let this_arg = this_arg.scope(agent, gc.nogc());
+    let scoped_o = o.scope(agent, gc.nogc());
+    // 5. Let kept be a new empty List.
+    // 6. Let captured be 0.
+    let mut kept: Vec<T> = Vec::with_capacity(len.try_into().unwrap());
+    // 7. Let k be 0.
+    let mut k = 0;
+    // 8. Repeat, while k < len,
+    while k < len {
```
aapoalas, line 3396: **nitpick:** You could just use `for k in 0..len`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3230,3 +3372,85 @@ fn fill_typed_array<'a, T: Viewable>(
...
+        // a. Let Pk be ! ToString(𝔽(k)).
+        let pk = k.try_into().unwrap();
+        // b. Let kValue be ! Get(O, Pk).
+        let k_value =
```
aapoalas, line 3400: **issue:** My idea here was to get the actual `&[T]` slice of the `o` and use normal Rust `slice.get(k)` to get `k_value`; if the result is `None` it means that we've gone out of bounds and should pass `Value::Undefined` to the callback as the value. If it's `Some(T)` then we'll use `T::into_le/be_value` and pass the result to the callback as the value. Then, when we get `selected` we'll just push `Some(T).unwrap_or(T::default())` to `kept`.

This way we avoid converting `T` to `Value` and back, and we avoid needing to go through the generic `Get` function as well.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3230,3 +3372,85 @@ fn fill_typed_array<'a, T: Viewable>(
...
+                scoped_o.get(agent).into_value().unbind(),
+            ])),
+            gc.reborrow(),
+        )?;
+        let selected = to_boolean(agent, call);
+        // d. If selected is true, then
+        if selected {
+            //  i. Append kValue to kept.
+            let value = T::try_from_value(agent, k_value.get(agent));
+            let Some(value) = value else {
```
aapoalas, line 3419: **note:** This whole error branch here then becomes unnecessary; also the error message here isn't quite correct :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3230,3 +3372,85 @@ fn fill_typed_array<'a, T: Viewable>(
...
+    }
+    // 9. Let A be ? TypedArraySpeciesCreate(O, « 𝔽(captured) »).
+    let len = kept.len() as i64;
+    let o = scoped_o.get(agent);
+    let a = typed_array_species_create_with_length::<T>(agent, o, len, gc.reborrow())?
+        .unbind()
+        .bind(gc.nogc());
+    // 10. Let n be 0.
+    // 11. For each element e of kept, do
+    for (n, e) in kept.iter().enumerate() {
```
aapoalas, line 3440: **issue:** Here we'll want to get a `&mut [T]` from `a` and do a copy-from-slice.

aapoalas (changes requested):
Looks good on the tin, but we can go a bit closer to the metal and do more direct memory copies :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1388,3 +1392,192 @@ pub(crate) fn typed_array_create_same_type<'a>(
...
+        TypedArray::BigUint64Array(_) => TypeId::of::<T>() == TypeId::of::<u64>(),
+        #[cfg(feature = "proposal-float16array")]
+        TypedArray::Float16Array(_) => TypeId::of::<T>() == TypeId::of::<f16>(),
+        TypedArray::Float32Array(_) => TypeId::of::<T>() == TypeId::of::<f32>(),
+        TypedArray::Float64Array(_) => TypeId::of::<T>() == TypeId::of::<f64>(),
+    };
+    if !type_id_matches {
+        return Err(agent.throw_exception_with_static_message(
+            ExceptionType::TypeError,
+            "TypedArray out of bounds",
```
aapoalas, line 1481: **issue:** Error message here is wrong: this error happens in this sort of situation:
```rust
class MySillyTypedArray extends BigUint64Array {
  constructor(...args) {
    super(...args);
    if (something(args)) {
      return new Uint8Array(1024);
    }
  }
}
```
ie. a custom TypedArray-subclass that (sometimes) returns a _different_ TA than what it extends. In code if we got the actual `this` object and were calling filter on it, then we'd find that our `exemplar` would be `TypedArray::BigUint64Array` but when we call its constructor, if it went through the `if (something)` then we'd get a `TypedArray::Uint8Array` as a our result.

This would be an issue because later we'd try to assign 64-bit values from our source TypedArray into our target TypedArray that can only hold 8-bit values etc. This would be a very easy path to undefined behaviour and memory unsafety, buffer overflows and all that.

So, this error message should be something like "wrong TypedArray flavour" or whatever. Going with what Firefox gives here is again a good idea.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1388,3 +1392,192 @@ pub(crate) fn typed_array_create_same_type<'a>(
...
+pub(crate) fn typed_array_species_create_with_buffer<'a, T: Viewable + 'static>(
+    agent: &mut Agent,
+    exemplar: TypedArray,
+    array_buffer: ArrayBuffer,
+    byte_offset: i64,
+    length: Option<i64>,
+    mut gc: GcScope<'a, '_>,
+) -> JsResult<'a, TypedArray<'a>> {
+    // 1. Let defaultConstructor be the intrinsic object associated with the constructor name exemplar.[[TypedArrayName]] in Table 73.
+    let default_constructor = {
```
aapoalas, line 1498: **thought:** This could be extracted into a separate function, actually, so you wouldn't need to duplicate it in both calls.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1388,3 +1392,192 @@ pub(crate) fn typed_array_create_same_type<'a>(
...
+        array_buffer,
+        byte_offset,
+        length,
+        gc.reborrow(),
+    )
+    .unbind()?
+    .bind(gc.nogc());
+    // 4. Assert: result has [[TypedArrayName]] and [[ContentType]] internal slots.
+    // 5. If result.[[ContentType]] is not exemplar.[[ContentType]], throw a TypeError exception.
+    let type_id_matches = match result {
```
aapoalas, line 1559: **thought:** This too could be separated into a helper function.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -1284,13 +1284,157 @@ impl TypedArrayPrototype {
...
+                typed_array_length::<u32>(agent, &ta_record, gc.nogc())
+            }
+            TypedArray::BigInt64Array(_)
+            | TypedArray::BigUint64Array(_)
+            | TypedArray::Float64Array(_) => {
+                typed_array_length::<u64>(agent, &ta_record, gc.nogc())
+            }
+        } as i64;
+        // 4. If IsCallable(callback) is false, throw a TypeError exception.
+        let Some(callback) = is_callable(callback, gc.nogc()) else {
```
aapoalas, line 1331: **note:** You could move this step to happen before step 3; `TypedArrayLength` cannot be observed nor can `IsCallable`, so they can be freely reorganised.

With that reorganisation, you could move the `typed_array_length` call inside of `filter_typed_array` and avoid the second match call.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,103 @@ fn fill_typed_array<'a, T: Viewable>(
...
+    // 5. Let kept be a new empty List.
+    // 6. Let captured be 0.
+    let mut kept: Vec<T> = Vec::with_capacity(len.try_into().unwrap());
+    // 7. Let k be 0.
+    // 8. Repeat, while k < len,
+    // b. Let kValue be ! Get(O, Pk).
+    let ta = scoped_o.get(agent);
+    for k in 0..len {
+        let array_buffer = ta.get_viewed_array_buffer(agent, gc.nogc());
+        let byte_offset = ta.byte_offset(agent);
```
aapoalas, line 3545: **nitpick:** `byte_offset` and `byte_length` can never change, so these could be moved out of the function.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,103 @@ fn fill_typed_array<'a, T: Viewable>(
...
+    let scoped_o = o.scope(agent, gc.nogc());
+    // 5. Let kept be a new empty List.
+    // 6. Let captured be 0.
+    let mut kept: Vec<T> = Vec::with_capacity(len.try_into().unwrap());
+    // 7. Let k be 0.
+    // 8. Repeat, while k < len,
+    // b. Let kValue be ! Get(O, Pk).
+    let ta = scoped_o.get(agent);
+    for k in 0..len {
+        let array_buffer = ta.get_viewed_array_buffer(agent, gc.nogc());
```
aapoalas, line 3544: **suggestion:** You could scope the array_buffer outside of the loop and `.get()` it from the `Scoped<ArrayBuffer>` in each loop, since the TypedArray cannot change which ArrayBuffer it views.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,103 @@ fn fill_typed_array<'a, T: Viewable>(
...
+    let callback = callback.scope(agent, gc.nogc());
+    let this_arg = this_arg.scope(agent, gc.nogc());
+    let scoped_o = o.scope(agent, gc.nogc());
+    // 5. Let kept be a new empty List.
+    // 6. Let captured be 0.
+    let mut kept: Vec<T> = Vec::with_capacity(len.try_into().unwrap());
+    // 7. Let k be 0.
+    // 8. Repeat, while k < len,
+    // b. Let kValue be ! Get(O, Pk).
+    let ta = scoped_o.get(agent);
```
aapoalas, line 3542: **issue:** This is not GC-safe. `scoped_o` is GC-safe, but the value returned from `.get(agent)` must be bound if it is assigned to a variable. And once you do bind `scoped_o.get(agent).bind(gc.nogc())` then the borrow checker will become very angry indeed.

But! This will mostly go away when you move `byte_offset` and `byte_length` outside of the loop, and scope `array_buffer`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,103 @@ fn fill_typed_array<'a, T: Viewable>(
...
+        if !head.is_empty() {
+            return Err(agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "TypedArray is not properly aligned",
+                gc.into_nogc(),
+            ));
+        }
+        let index: usize = k.try_into().unwrap();
+        let value = slice.get(index).copied();
+        let k_value = value.map_or_else(
```
aapoalas, line 3568: **nitpick:** `map_or` allows you to get create the `Value::Undefined` value without a closure.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,103 @@ fn fill_typed_array<'a, T: Viewable>(
...
+            || Value::Undefined,
+            |v| v.into_le_value(agent, gc.nogc()).into_value(),
+        );
+        let call = call_function(
```
aapoalas, line 3572: **nitpick:** `call` sounds like this is a function or something. Rather use `result`? Or just `selected`?

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,103 @@ fn fill_typed_array<'a, T: Viewable>(
...
+        )
+        .unbind()?
+        .bind(gc.nogc());
+        let selected = to_boolean(agent, call);
+        if selected {
+            kept.push(value.unwrap_or(T::default()));
+        }
+    }
+    // 9. Let A be ? TypedArraySpeciesCreate(O, « 𝔽(captured) »).
+    let len = kept.len() as i64;
```
aapoalas, line 3591: **nitpick:** Use the `captured` variable name from spec; `let captured = kept.len()`.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3374,3 +3518,103 @@ fn fill_typed_array<'a, T: Viewable>(
...
+    let o = scoped_o.get(agent);
```
aapoalas, line 3592: **issue:** Missing bind.

nova_vm/src/ecmascript/types/spec/data_block.rs
```diff
@@ -549,6 +589,9 @@ impl Viewable for f16 {
             None
         }
     }
+    fn default() -> Self {
```
aapoalas, line 592: **praise:** I would've gotten this incorrect! I had though the `undefined -> FloatXArray` conversion would produce 0 and not NaN. Nice work checking and getting this correct <3

aapoalas (changes requested):
Still some issues, sorry about this :(

yossydev:
@aapoalas 
Sorry for the delay!
I think I’ve made the corrections based on your review!
When you have a moment, could you please take another look? 🙇‍♂️ https://github.com/trynova/nova/pull/660/commits/ac8756a3a0f2cdd2a70e6ff6d25a319498e784d0

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3788,3 +3946,104 @@ fn sort_comparator_typed_array<'a, T: Viewable + Copy + std::fmt::Debug>(
...
+    // 5. Let kept be a new empty List.
+    // 6. Let captured be 0.
+    let mut kept: Vec<T> = Vec::with_capacity(len.try_into().unwrap());
+    // 7. Let k be 0.
+    // 8. Repeat, while k < len,
+    // b. Let kValue be ! Get(O, Pk).
+    let ta = scoped_o.get(agent);
+    let byte_offset = ta.byte_offset(agent);
+    let byte_length = ta.byte_length(agent);
+    let array_buffer = ta
```
aapoalas, line 3973: Let's do this:
```suggestion
    let local_array_buffer = ta.get_viewed_array_buffer(agent, gc.nogc());
    let array_buffer = local_array_buffer.scope(agent, gc.nogc());
    let properly_aligned = unsafe { local_array_buffer.as_slice(agent).align_to::<T>().0.is_empty() };
    if !properly_aligned {
      return Err(agent.throw_exception_with_static_message(
        ExceptionType::TypeError,
        "TypedArray is not properly aligned",
        gc.into_nogc(),
      ));
    }
```

With this we can skip checking the alignment on every loop iteration.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3788,3 +3946,104 @@ fn sort_comparator_typed_array<'a, T: Viewable + Copy + std::fmt::Debug>(
...
+        )
+        .unbind()?
+        .bind(gc.nogc());
+        let selected = to_boolean(agent, result);
+        if selected {
+            kept.push(value.unwrap_or(T::default()));
+        }
+    }
+    // 9. Let A be ? TypedArraySpeciesCreate(O, « 𝔽(captured) »).
+    let captured = kept.len() as i64;
```
aapoalas, line 4020: **nitpick:** Maybe don't do `as i64` here but instead only in the `typed_array_species_create_with_length` call; saves one `as` conversion :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3788,3 +3946,104 @@ fn sort_comparator_typed_array<'a, T: Viewable + Copy + std::fmt::Debug>(
...
+        if end_index > byte_slice.len() {
+            return Ok(a.unbind());
+        }
+        &mut byte_slice[byte_offset..end_index]
+    } else {
+        &mut byte_slice[byte_offset..]
+    };
+    let (_, slice, _) = unsafe { byte_slice.align_to_mut::<T>() };
+    let len = (captured as usize).min(slice.len());
+    let copy_len = slice.len().min(len);
```
aapoalas, line 4045: **nitpick:** This is always equal to `len` since this is just `slice.len().min(captured.min(slice.len()))`.

aapoalas (changes requested):
Some issues got left over.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/abstract_operations.rs
```diff
@@ -1388,3 +1392,144 @@ pub(crate) fn typed_array_create_same_type<'a>(
...
+        agent,
+        constructor.unbind(),
+        length,
+        gc.reborrow(),
+    )
+    .unbind()?
+    .bind(gc.nogc());
+    // 4. Assert: result has [[TypedArrayName]] and [[ContentType]] internal slots.
+    // 5. If result.[[ContentType]] is not exemplar.[[ContentType]], throw a TypeError exception.
+    let is_type_match = is_expected_typed_array_type::<T>(result);
```
aapoalas, line 1481: **issue (for me to fix):** Oh crud, I was wrong here! Your original implementation was pretty much spot on; the types do not have to match exactly, only the BigInt / Number stuff needs to!

I'm so sorry, I made you do a refactoring because of my misread. I'll do the rework to put it back to rights.

yossydev: No problem at all! Thanks for the fix!!

yossydev:
@aapoalas 
Thank you so much again! I’m really sorry for all the little things I missed! 🙇‍♂️
I’ve made the fixes again, so if you have a moment, I’d appreciate it if you could take another look. 🙇‍♂️

aapoalas (approved):
LGTM! Good job with this, despite it taking a bit of time :) <3

PR #663, opened 2025-04-13
feat(ecmascript): %TypedArray%.prototype.with

yossydev:
doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.with


load1n9 (dismissed):
LGTM! Nice job with the garbage collector stuff

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3230,3 +3327,90 @@ fn fill_typed_array<'a, T: Viewable>(
...
+    // 10. Let A be ? TypedArrayCreateSameType(O, « 𝔽(len) »).
+    let a = typed_array_create_same_type(agent, o, len, gc.reborrow())
+        .unbind()
+        .bind(gc.nogc())?
+        .scope(agent, gc.nogc());
+    // 11. Let k be 0.
+    let mut k = 0;
+    let scoped_o = o.scope(agent, gc.nogc());
+    // 12. Repeat, while k < len,
+    while k < len {
```
aapoalas, line 3390: **issue:** The way we want to write this loop is to instead get a `&mut [T]` of `a`, and a `&[T]` of `o` (of length `len` or smaller for both; `o` might have been shrunk or detached, and `a` might've been detached before we come to this line) and then do a copy-from-slice from `o` to `a`, possibly with `split_at_[mut]_checked` to split at `actual_index` to get a "pre-slice" that can be copy-from-slice'd directly, and a "post-slice" that contains our "actual index" which we `split_first_mut` (and unwrap) to get the actual_index slot to write `numeric_value` into, and then a "rest" slice that we can again copy-from-slice directly.

The tough part here will be that trying to get a mutable slice of `a` while we have an immutable slice of `o` is not going to work directly: Rust does not allow that as it would cause aliasing of mutable memory/state if `o` and `a` point to the same ArrayBuffer. But luckily in this case it is completely impossible for that to happen: based on my reading, `TypedArrayCreateSameType` cannot actually call user code and thus it will always call the normal TypedArray constructor for `T`. That actually also means that we could optimise this code and others like it quite a bit more :) But let's leave that for tomorrow.

Anyway, getting a mutable slice to `a` and `o` at the same time would be done correctly by
1. Setting up helper methods for getting a mutable/immutable slice from a `T: Viewable` TypedArray in the first place: that slice generating code is starting to proliferate too much.
2. Set up a helper method that gives two mutable slices for two TypedArrays that are both `T: Viewable` (the same type); that method would then check that the two TA's point to different buffers and panic if they don't. Then it'll just get slices from the two ArrayBuffers, one at a time, converting the first one `into_raw_parts` before getting the second, and then reconsistutes the first with `from_raw_parts`: the reconsitution is unsafe but we've checked that the two ABs are not the same so they're distinct memory areas and we can thus safely "split" the `&mut Agent` lifetime into these two `&mut [u8]` slice's lifetimes. Then we just make sure both are alignable as `&mut [T]` and return to the caller.

aapoalas (changes requested):
Some minor fixes to be made, and then there's a big point of making the data copying more efficient. That'll need some work.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: Thanks for the review!! And sorry for the delay!

Setting up helper methods for getting a mutable/immutable slice from a T: Viewable TypedArray in the first place: that slice generating code is starting to proliferate too much.

Got it, that makes sense.

Then it’ll just get slices from the two ArrayBuffers, one at a time, converting the first one into_raw_parts before getting the second

When I looked it up, it seemed like into_raw_parts isn’t a stable API yet, so just to confirm — does this mean we should create our own API with the same functionality…? 👀

and then reconsistutes the first with from_raw_parts:

It seems like from_raw_parts already exists for slices, so I’m planning to use that 👀 (https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html)

aapoalas: In place of `slice.into_raw_parts()` you can use `slice.as_mut_ptr()` and `slice.len()` to get the "raw parts" individually.

yossydev: > Set up a helper method that gives two mutable slices for two TypedArrays that are both T: Viewable (the same type); that method would then check that the two TA's point to different buffers and panic if they don't. Then it'll just get slices from the two ArrayBuffers, one at a time, converting the first one into_raw_parts before getting the second, and then reconsistutes the first with from_raw_parts: the reconsitution is unsafe but we've checked that the two ABs are not the same so they're distinct memory areas and we can thus safely "split" the &mut Agent lifetime into these two &mut [u8] slice's lifetimes. Then we just make sure both are alignable as &mut [T] and return to the caller.

Sorry for the delay — before setting up the shared helper, I went ahead and implemented part 2 first! https://github.com/trynova/nova/pull/663/commits/e5db4174a175514367667a6b39ac17dc7a6c49ec

yossydev: Alright, I’ve created a helper function to get both an immutable and a mutable slice, and I’ve updated the existing code to use it!
(For the places that aren’t using Viewable yet, I thought it’d be better to fix them together when we eventually update those.)

Would you mind reviewing it for me? 🙇‍♂️

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -4194,3 +4164,95 @@ fn copy_between_same_type_typed_arrays<T: Viewable>(kept: &[T], byte_slice: &mut
...
+    } else {
+        to_integer_or_infinity(agent, index.unbind(), gc.reborrow())
+            .unbind()?
+            .bind(gc.nogc())
+            .into_i64()
+    };
+    let numeric_value = scoped_value
+        .get(agent)
+        .into_value()
+        .to_numeric(agent, gc.reborrow())
```
aapoalas, line 4215: **issue:** The spec says to perform either `ToBigInt` or `ToNumber`, based on `T::IS_BIGINT`. `ToNumeric` is not equal to that.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -4194,3 +4164,95 @@ fn copy_between_same_type_typed_arrays<T: Viewable>(kept: &[T], byte_slice: &mut
...
+    let len = typed_array_length::<T>(agent, &ta_record, gc.nogc()) as i64;
+    // 5. If relativeIndex ≥ 0, let actualIndex be relativeIndex.
+    let actual_index = if relative_index >= 0 {
+        relative_index
+    } else {
+        // 6. Else, let actualIndex be len + relativeIndex.
+        len + relative_index
+    };
+    // 9. If IsValidIntegerIndex(O, 𝔽(actualIndex)) is false, throw a RangeError exception.
+    if is_valid_integer_index_generic(agent, o, actual_index, gc.nogc()).is_none() {
```
aapoalas, line 4230: **nitpick:** We have `is_valid_integer_index::<T: Viewable>` which would be preferable here since we already have a `T: Viewable` generic type parameter.

aapoalas (changes requested):
Some issues remain. The code overall looks really good with the helper functions <3

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3307,13 +3308,111 @@ impl TypedArrayPrototype {
...
+                gc,
+            ),
+            TypedArray::Float64Array(_) => with_typed_array::<f64>(
+                agent,
+                ta_record.unbind(),
+                index.unbind(),
+                value.unbind(),
+                gc,
+            ),
+        };
```
eliassjogreen, line 3413: This can be simplified to (#689):
```suggestion
        let o = with_typed_array_viewable!(
          ta_record.object,
          with_typed_array::<T>(
            agent,
            ta_record.unbind(),
            index.unbind(),
            value.unbind(),
            gc,
          )
        );
```

eliassjogreen (changes requested):

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
yossydev: You’re awesome, thank you so much!!!

yossydev:
It seems like the results differ between `--profile release` and `--profile dev-fast`… 👀

```
✘ cargo build --profile dev-fast && cargo run --bin test262 --profile dev-fast -- -u
    Finished `dev-fast` profile [optimized + debuginfo] target(s) in 22.57s
    Finished `dev-fast` profile [optimized + debuginfo] target(s) in 0.06s
     Running `target/dev-fast/test262 -u`
Updating the expectations file with 1 unexpected test results.

❯ cargo build --profile release && cargo run --bin test262 --profile release -- -u
    Finished `release` profile [optimized] target(s) in 0.20s
    Finished `release` profile [optimized] target(s) in 0.06s
     Running `target/release/test262 -u`
Updating the expectations file with 1 unexpected test results.
```

By the way, the test in question is timing out:

```
❯ cargo build && cargo run --bin test262 eval-test built-ins/Array/prototype/push/S15.4.4.7_A3.js
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.19s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/test262 eval-test built-ins/Array/prototype/push/S15.4.4.7_A3.js`
Loose mode run:
Running: "${HOME}/target/debug/nova_cli" "eval" "--expose-internals" "--no-strict" "--nogc" "${HOME}/tests/nova-harness.js" "${HOME}/tests/test262/harness/assert.js" "${HOME}/tests/test262/harness/sta.js" "${HOME}/tests/test262/test/built-ins/Array/prototype/push/S15.4.4.7_A3.js"

Test result: Timeout
```

Is it okay to skip this…?

tests/skip.json
```diff
@@ -10,6 +10,7 @@
     "built-ins/Array/length/15.4.5.1-3.d-3.js",
     "built-ins/Array/length/S15.4.2.2_A2.1_T1.js",
     "built-ins/Array/length/S15.4.5.2_A3_T4.js",
+    "built-ins/Array/prototype/push/S15.4.4.7_A3.js",
```
yossydev, line 13: In my environment, this part seems to cause a difference. For now, I’ve made it skip.

https://github.com/trynova/nova/pull/663#issuecomment-2863313374

aapoalas: Ah yeah, that's one of the "Array as hash map" tests, which we don't really support. Skip is absolutely good and correct here <3

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -2957,3 +2848,105 @@ fn copy_between_same_type_typed_arrays<T: Viewable>(kept: &[T], byte_slice: &mut
...
+    let scoped_o = o.scope(agent, gc.nogc());
+    let scoped_value = value.scope(agent, gc.nogc());
+    // 3. Let len be TypedArrayLength(taRecord).
+    let len = typed_array_length::<T>(agent, &ta_record, gc.nogc()) as i64;
+    // 4. Let relativeIndex be ? ToIntegerOrInfinity(index).
+    // 5. If relativeIndex ≥ 0, let actualIndex be relativeIndex.
+    let relative_index = if let Value::Integer(index) = index {
+        index.into_i64()
+    } else {
+        to_integer_or_infinity(agent, index.unbind(), gc.reborrow())
```
aapoalas, line 2896: **note:** You could skip scoping `value` in the `if` branch and only do it in this `else` branch.

aapoalas (dismissed):
LGTM! Real nice work, this really cleaned up a lot of the TA prototype methods related code which is great to see <3

eliassjogreen (approved):
Looks good!

PR #668, opened 2025-04-23
feat(engine): compile for-in/of bodies for non-assignment destructures

DonIsaac:
## What This PR does

Updates `for_in_of_body_evaluation` to handle compilation of non-assignment destructure bindings. This reduces the number of crashing test262 cases by 100.


nova_vm/src/engine/bytecode/bytecode_compiler/for_in_of_statement.rs
```diff
@@ -187,9 +187,23 @@ fn for_in_of_body_evaluation(
                 } else {
                     // 2. Else,
                     // a. Assert: lhsKind is VAR-BINDING.
+                    assert_eq!(lhs_kind, LeftHandSideKind::VarBinding);
                     // b. Assert: lhs is a ForBinding.
                     // c. Let status be Completion(BindingInitialization of lhs with arguments nextValue and undefined).
-                    todo!();
+                    match lhs {
```
aapoalas, line 193: **nitpick:** You could use a `let ForStatementLeft::VariableDeclaration(decl) = lhs else { unreachable!() }` to de-indent the important stuff by one step :)

aapoalas (dismissed):
LGTM! You got an absolutely superb +/- ratio here, +19/-99 is a level we haven't seen in a good while I think :)

Also, somewhat long time no see, thanks for dropping by again! <3

aapoalas (approved):

PR #669, opened 2025-04-23
feat(ecmascript): `BigInt.prototype.toString`

eliassjogreen:


nova_vm/src/ecmascript/abstract_operations/type_conversion.rs
```diff
@@ -453,6 +453,32 @@ impl IntegerOrInfinity {
     }
 }
 
+impl PartialEq<i64> for IntegerOrInfinity {
```
aapoalas, line 456: **praise:** Ooh, nice idea!

nova_vm/src/ecmascript/types/language/bigint.rs
```diff
@@ -613,6 +613,27 @@ impl<'a> BigInt<'a> {
...
+        gc: NoGcScope<'gc, '_>,
+    ) -> String<'gc> {
+        String::from_string(
+            agent,
+            match x {
+                BigInt::SmallBigInt(x) => {
+                    // NOTE: The rust standard library provides no way to convert
+                    // numbers into strings with a radix which is why I am borrowing
+                    // num_bigints `to_str_radix` implementation.
+                    x.into_i64().to_bigint().unwrap().to_str_radix(radix)
```
aapoalas, line 630: **question:** How do we do to_string_radix for integer numbers...? I think we'd prefer to use the same logic here, so as to avoid the bigint heap allocation.

aapoalas (changes requested):
LGTM, except for the expectations needing to be updated. Nice one, this has been a bit of a thorn in my side for a good long while! :)

nova_vm/src/ecmascript/types/language/bigint.rs
aapoalas: Oh, I see. Yeah, roger that.

aapoalas (approved):
LGTM <3

PR #670, opened 2025-04-23
feat(ecmascript): `Number.prototype.toString`

eliassjogreen:


nova_vm/src/ecmascript/builtins/numbers_and_dates/number_objects/number_prototype.rs
```diff
@@ -433,21 +433,44 @@ impl NumberPrototype {
         (flt.len() as i32) - 1
     }
 
+    /// ### [21.1.3.6 Number.prototype.toString ( [ radix ] )](https://tc39.es/ecma262/#sec-number.prototype.tostring)
```
aapoalas, line 436: **praise:** Thank you for adding the notes and steps <3

nova_vm/src/ecmascript/types/language/number.rs
```diff
@@ -1315,6 +1317,24 @@ impl<'a> Number<'a> {
...
+        x: Self,
+        radix: u32,
+        gc: NoGcScope<'gc, '_>,
+    ) -> String<'gc> {
+        String::from_string(
+            agent,
+            match x {
+                Number::Number(x) => agent[x].to_radix_str(radix as u8).unwrap(),
+                Number::SmallF64(x) => x.into_f64().to_radix_str(radix as u8).unwrap(),
+                Number::Integer(x) => x.into_i64().to_bigint().unwrap().to_str_radix(radix),
```
aapoalas, line 1332: **note:** It's probably better performance-wise to go from i64 to f64 (we're guaranteed that it's a lossless conversion as well) and use the ecmascript-radix library.

aapoalas (dismissed):
LGTM! <3

aapoalas:
Turns out ecmascript-radix doesn't work very well :) -1 gets written out as "-00" for some radii.

nova_vm/src/ecmascript/types/language/number.rs
eliassjogreen: We cant use the ecmascript-radix library for the integer conversions actually

aapoalas: Oh?

eliassjogreen:
Yep

eliassjogreen:
Got some merge conflicts locally

eliassjogreen:
But yeah. The crate didn't work for ints nor for a lot of radii.

aapoalas:
> But yeah. The crate didn't work for ints nor for a lot of radixes. Hack fix was to still use the big-int `to_str_radix` for integers which seems to pass all tests for radixes `2-37`.

We should probably contribute tests and fixes.

aapoalas (approved):

eliassjogreen:
Either that or we borrow boa's [implementation](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/number/mod.rs#L539) which in turn is adapted from v8's. Although, I can't say I like the look of the 2200 byte heap allocation but if it's okay for v8 I guess it's okay for any engine.

aapoalas:
Nevermind, it actually uses a 2200 char allocation mistakenly, leading to a 8800 byte allocation.

eliassjogreen:
Seems the v8 one is popular, even spidermonkey seems to use it:
https://github.com/servo/mozjs/blob/main/mozjs-sys/mozjs/mfbt/double-conversion/double-conversion/double-to-string.cc#L174

aapoalas:
Hmm... I've been refactoring / fixing the radix-ecmascript implementation and found a V8 bug, apparently. Compare `(1.7976931348623157e308).toString(3)` to `BigInt(1.7976931348623157e308).toString(3)`, and then do the same in Firefox.

Chrome disagrees with itself on how to stringify this float compared to the same number as an integer. Boa's version also has that same bug, but Firefox doesn't. Gotta look at how they implement ... some part there?

aapoalas (approved):
LGTM!

PR #679, opened 2025-04-30
chore: bump oxc -> 0.67

load1n9:


load1n9:
new version implements `ast::PropertyKey::V8IntrinsicExpression` which i put aside with a todo comment once we determine how it should be handled

aapoalas (approved):
LGTM, thanks!

I'll implement the intrinsic expressions some day soon.

PR #681, opened 2025-05-02
fix(vm): Handle empty `for` test expression

eliassjogreen:
This fixes among other things the `Octane/box2d.js` benchmark


aapoalas (approved):
Nice! <3

PR #683, opened 2025-05-02
chore: auto-impl `IntoValue` for `Into<Value>`

DonIsaac:
## What This PR Does

Adds an automatic trait implementation for `IntoValue<'a>` for types that satisfy `Into<Value<'a>>`. Doing this removes a lot of redundant code.

We should probably make follow-up PRs that do something similar for `IntoObject` and `IntoPrimitive`


aapoalas:
Excellent, thank you <3 Absolutely, doing the same for IntoObject and IntoPrimitive would be much appreciated. 

This is the kind of stuff I keep thinking about but never get around to :D

PR #684, opened 2025-05-02
feat(vm): Labelled statements

aapoalas:
Also some fixes found around the work.


nova_vm/src/engine/bytecode/bytecode_compiler.rs
```diff
@@ -48,7 +53,26 @@ pub(crate) enum NamedEvaluationParameter {
     ReferenceStack,
 }
 
-pub(crate) struct CompileContext<'agent, 'gc, 'scope> {
+pub(crate) struct JumpTarget {
```
DonIsaac, line 56:
```suggestion
#[derive(Debug, Default)]
pub(crate) struct JumpTarget {
```

nova_vm/src/engine/bytecode/bytecode_compiler.rs
```diff
@@ -48,7 +53,26 @@ pub(crate) enum NamedEvaluationParameter {
...
+    /// Depth of the lexical of the jump target.
```
DonIsaac, line 57: what does a jump target's lexical depth mean?

nova_vm/src/engine/bytecode/bytecode_compiler.rs
```diff
@@ -48,7 +53,26 @@ pub(crate) enum NamedEvaluationParameter {
...
+    pub(super) fn new(depth: u32) -> Self {
+        Self {
+            depth,
+            continues: vec![],
+            breaks: vec![],
+        }
+    }
+}
+
+pub(crate) struct CompileContext<'agent, 'script, 'gc, 'scope> {
```
DonIsaac, line 75: There's getting to be a lot of lifetimes here, some docs could be helpful

aapoalas: Good point; I'll add some.

nova_vm/src/engine/bytecode/bytecode_compiler.rs
aapoalas: I'll add more comments here, but: it's basically memory of how many declarative environments we had entered when the jump target (loop or switch statement) was created. When a break or continue statement is found, it compares the current number of declarative environments entered (this is tracked in the context) against the jump target memory; this gives us the number of how many environments need to be exited when breaking / continuing.

nova_vm/src/engine/bytecode/bytecode_compiler.rs
aapoalas: Hmm, this is never debug printed so that'd be just a bit of unnecessary macro expansion, and the default implementation on this should never be used (it'd give lexical depth of 0 which may not be right).

I'll leave these out.

PR #685, opened 2025-05-02
build: run test262 with new dev-fast profile

DonIsaac:
## What This PR Does
Adds a `dev-fast` build profile and uses it to run Test262 in CI.

Using this profile trades slightly longer compile times for much faster run times. It takes ~10s longer to build dev builds while taking half the time of release builds, but runs tests 300x faster than dev while only being marginally slower than release builds.

Below are build/test times collected (quite non-scientifically) from my local machine (m1 mac pro, 16gb).
* Each build is "clean" (`cargo clean && cargo build --profile <profile>`).
* Test262 times are in "real" time (`/usr/bin/time -h cargo run --bin test262 --profile <profile>`)

|         | dev      | dev-fast | release |
|---------|----------|----------|---------|
| build   | 27.07s   | 39.83s   | 1m 06s  |
| test262 | 1m32.19s | 33.70s   | 33.58s  |

## Other Changes
- Move spell checking to separate job
- split build + unit testing and test262 into separate jobs. The latter uses a more expensive profile, and its faster to use dev profile for the former.
- add caching for lint and test262 jobs
- use `cargo binstall` to install dylink. Avoids re-compiling it on each run.
- new pushes to a PR branch will cancel in-progress jobs
- test failures will print stack traces (both unit and test262)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,9 @@ jobs:
           shared-key: warm
           save-if: ${{ github.ref == 'main' }}
       - name: Install Dylint
-        run: cargo install cargo-dylint dylint-link
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-dylint dylint-link
       - name: Check formatting
         run: cargo fmt --check
       - name: Clippy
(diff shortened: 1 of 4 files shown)


DonIsaac:
Once I add a step to cache results from `cargo build`, this will get even faster

aapoalas:
Ooh, this is really nice! Great idea! Do you want to add results caching in this PR, or do should I merge directly?

DonIsaac:
I'll add caching to the next one

DonIsaac:
@aapoalas ready; merge whenever

PR #686, opened 2025-05-02
perf(bytecode): elide arguments object instructions when its never used

DonIsaac:
## What This PR Does
- Makes oxc's semantic analysis information available when compiling scripts
- Use that info to not generate instructions that create a function's `arguments` object when that function's body never references it


nova_vm/src/ecmascript/scripts_and_modules/source_code.rs
```diff
@@ -123,13 +126,15 @@ impl<'a> SourceCode<'a> {
             // TODO: Include error messages in the exception.
             return Err(errors);
         }
+        let semantic = unsafe { core::mem::transmute::<Semantic, Semantic<'static>>(semantic) };
```
aapoalas, line 129: **issue:** Safety comment needed. You can just extend the below `unsafe {}` block transmuting `Program` to encompass this as well, and modify the safety comment to cover both `Program` and `Semantic`.

nova_vm/src/ecmascript/scripts_and_modules/source_code.rs
```diff
@@ -123,13 +126,15 @@ impl<'a> SourceCode<'a> {
...
         // SAFETY: Caller guarantees that they will drop the Program before
         // SourceCode can be garbage collected.
         let program = unsafe { core::mem::transmute::<Program, Program<'static>>(program) };
-        let source_code = agent.heap.create(SourceCodeHeapData {
-            source: source.unbind(),
+        let source_code = agent.heap.create(SourceCodeHeapData::new(
             allocator,
-        });
+            source.unbind(),
```
aapoalas, line 135: **nitpick:** You can probably get rid of this unbind by tweaking the `new` function.

Best possible scenario is that we have `SourceCodeHeapData::new(source: HeapString<'a>) -> SourceCodeHeapData<'a>` and then the `CreateHeapData` for `SourceCodeHeapData<'a>` would return `SourceCode<'a>`. That way the GC lifetime automatically gets threaded through here.

nova_vm/src/ecmascript/scripts_and_modules/source_code.rs
```diff
@@ -150,9 +170,33 @@ pub struct SourceCodeHeapData<'a> {
     /// string was small-string optimised and on the stack, then those
     /// references would necessarily and definitely be invalid.
     source: HeapString<'a>,
+    /// Semantic analysis results obtained when parsing a source file.
+    ///
+    /// This only exists until this source code gets compiled. Since it's only
```
aapoalas, line 175: **issue:** As mentioned in the one-off comment, parts of the source code get compiled at different times; each function instantiation (first time anyway) performs a new compilation. So, you cannot take the semantic out of here, at least not until you're sure that all the functions have been compiled.

But even then, maybe some day in the future we'll want to have tiered compilation where having the semantics available when we recompile a function is necessary. So, I think it's better to just leave the `Semantic` here.

nova_vm/src/ecmascript/scripts_and_modules/source_code.rs
```diff
@@ -170,6 +214,10 @@ impl Drop for SourceCodeHeapData<'_> {
         // SAFETY: All references to this SourceCode should have been dropped
         // before we drop this.
         drop(unsafe { Box::from_raw(self.allocator.as_mut()) });
+        // drop(
```
aapoalas, line 217: **note:** The drop order should be "sema first, allocator second".

nova_vm/src/ecmascript/scripts_and_modules/source_code.rs
```diff
@@ -249,18 +307,12 @@ unsafe impl Bindable for SourceCodeHeapData<'_> {
 
 impl HeapMarkAndSweep for SourceCodeHeapData<'static> {
     fn mark_values(&self, queues: &mut WorkQueues) {
-        let Self {
-            source,
-            allocator: _,
-        } = self;
+        let Self { source, .. } = self;
```
aapoalas, line 310: **issue:** Do _not_ use `..` in the GC HeapMarkAndSweep trait. The explicit skips are intended; when we add fields to the struct, the `..` makes it less likely that we'll remember to properly handle their GC.

eg. I've repeatedly forgot to add tracing of new fields back when I didn't yet have this rule of `no .. in GC`, which would then lead to weird bugs.

aapoalas (changes requested):
As mentioned, the semantics should remain usable multiple times over.

Otherwise, this is a great addition that will definitely be helpful in the future! Us not taking advantage of the semantic analysis oxc performs is a big mistake :)

aapoalas:
I did a refactoring pass on this because I wanted to merge this but found a couple of issues with using the Semantic's unresolved functions.

First, just checking the body isn't enough: we need to also check the formal parameters, but that was easy to solve.

A problem that is not easy to solve is that we don't get "arguments" as unresolved here:
```js
var arguments = undefined;
function foo() {
  return arguments.length;
}
```
So the check says that the `foo` function doesn't need the arguments object, but it does.

Another issue is that in this function we again get the wrong result from the check:
```js
function __func(){
    return typeof arguments;
    var arguments = "foo";
};
```
`arguments` is not unresolved, so the object doesn't get created.

A third issue is that `eval` can also mess with arguments, and of course this check doesn't guard for `eval`. So, that also needs checking.

It honestly seems like it'd be easier to not use the Semantic object at all, and instead just walk the AST and check references in it manually. That would also mean that `Program` could be handled like previously. Right now I had to move it into the `SourceCode` because `Semantic` holds a reference to `Program`.

PR #687, opened 2025-05-03
fix(engine): peek last instruction should never return non-instruction bytes

DonIsaac:
`ctx.peek_last_instruction` could return data from an index or jump label if the last instruction pushed had arguments. It now always returns an instruction.


aapoalas (approved):
Nice <3

DonIsaac:
@aapoalas I'm getting different results for Test262 locally than in CI. Could you update expectations/metrics files for me please?

aapoalas (approved):
LGTM! Nice work <3

I fixed the tests (re-added some skips to make them more reliable), and did a few minor code passes as well. Most importantly, I reduced the memory usage of the current instruction counter 4-fold by using a u32 and dealing with the empty case in a different place.

PR #689, opened 2025-05-03
chore: `with_typed_array_viewable` macro to reduce duplicate code

eliassjogreen:
Removes a *lot* of duplicate code within the code for typed arrays by introducing a `with_typed_array_viewable` which automatically matches the typed array variant to the corresponding `Viewable` type.


aapoalas (approved):
Oooh lovely, thank you so much <3

PR #692, opened 2025-05-07
feat(ecmascript): WeakRef

aapoalas:
Nice and easy little piece.


eliassjogreen (approved):
Didn't take much to implement weak refs 😄
Looks good

aapoalas:
> Didn't take much to implement weak refs 😄 Looks good

Yup, WeakRefs and WeakSets are pretty easy, I think. FinalizationRegistry and WeakMap should be where shit hits the fan with ephemerons, and those are then ... well, in a way they're easy but they're an absolute pain for the GC implementation since it requires coordination / going back-and-forth from the ephemeron to the key and back, before you can trace to the value.

PR #695, opened 2025-05-11
feat(ecmascript): %TypedArray%.prototype.map

yossydev:
- doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.map
- ref: https://github.com/trynova/nova/issues/145


tests/expectations.json
```diff
@@ -8573,86 +8572,7 @@
...
-  "built-ins/TypedArray/prototype/map/callbackfn-this.js": "FAIL",
-  "built-ins/TypedArray/prototype/map/detached-buffer.js": "FAIL",
-  "built-ins/TypedArray/prototype/map/invoked-as-func.js": "FAIL",
-  "built-ins/TypedArray/prototype/map/invoked-as-method.js": "FAIL",
-  "built-ins/TypedArray/prototype/map/resizable-buffer-grow-mid-iteration.js": "FAIL",
-  "built-ins/TypedArray/prototype/map/resizable-buffer-shrink-mid-iteration.js": "FAIL",
-  "built-ins/TypedArray/prototype/map/resizable-buffer.js": "FAIL",
-  "built-ins/TypedArray/prototype/map/return-abrupt-from-this-out-of-bounds.js": "FAIL",
-  "built-ins/TypedArray/prototype/map/return-new-typedarray-conversion-operation-consistent-nan.js": "FAIL",
   "built-ins/TypedArray/prototype/map/return-new-typedarray-conversion-operation.js": "FAIL",
```
yossydev, line 8575: It seems to be failing during comparison.

```
thread 'main' panicked at nova_vm/src/ecmascript/types/language/number.rs:542:17:
assertion failed: y.into_f64().to_bits() == (-0.0f64).to_bits() ||
    (x.into_i64() as f64) != y.into_f64()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

load1n9 (changes requested):
LGTM but fix conflicts

yossydev:
> LGTM but fix conflicts

thank you! I fixed it!

aapoalas (approved):
LGTM! I'll take a look at the remaining fail; it's probably some deeper thing. I'll also move the one callback check outside the function to "clean up" the specialised code.

I wish there was some way we could optimise the mapping code better but because it intersects with JS, there isn't much ways to do so :( Of course we could do the typed slice acquisition stuff and that would actually be somewhat better... But not by a whole massive lot.

tests/expectations.json
aapoalas: Alright, I fixed this by separating SmallBigInt from SmallInteger entirely.

yossydev:
> I wish there was some way we could optimise the mapping code better but because it intersects with JS, there isn't much ways to do so :( Of course we could do the typed slice acquisition stuff and that would actually be somewhat better... But not by a whole massive lot.

Yeah, someday I’d love to do it…
But it looks like a tough one.

PR #700, opened 2025-05-16
feat(ecmascript): Implement `decodeURI` and `decodeURIComponent`

eliassjogreen:


tests/expectations.json
```diff
@@ -9213,97 +9213,8 @@
...
-  "built-ins/decodeURI/S15.1.3.1_A1.4_T1.js": "FAIL",
-  "built-ins/decodeURI/S15.1.3.1_A1.5_T1.js": "FAIL",
-  "built-ins/decodeURI/S15.1.3.1_A1.6_T1.js": "FAIL",
-  "built-ins/decodeURI/S15.1.3.1_A1.7_T1.js": "FAIL",
-  "built-ins/decodeURI/S15.1.3.1_A1.8_T1.js": "FAIL",
-  "built-ins/decodeURI/S15.1.3.1_A1.8_T2.js": "FAIL",
-  "built-ins/decodeURI/S15.1.3.1_A1.9_T1.js": "FAIL",
-  "built-ins/decodeURI/S15.1.3.1_A1.9_T2.js": "FAIL",
-  "built-ins/decodeURI/S15.1.3.1_A1.9_T3.js": "FAIL",
   "built-ins/decodeURI/S15.1.3.1_A2.1_T1.js": "FAIL",
```
eliassjogreen, line 9216: The remaining two tests fail because of https://github.com/trynova/nova/blob/main/small_string/lib.rs#L57

nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -1085,38 +1085,141 @@ impl GlobalObject {
...
+        arguments: ArgumentsList,
+        mut gc: GcScope<'gc, '_>,
     ) -> JsResult<'gc, Value<'gc>> {
-        Err(agent.todo("decodeURI", gc.into_nogc()))
+        let encoded_uri = arguments.get(0).bind(gc.nogc());
+
+        // 1. Let uriString be ? ToString(encodedURI).
+        let uri_string = to_string(agent, encoded_uri.unbind(), gc.reborrow())
+            .unbind()?
+            .scope(agent, gc.nogc());
```
aapoalas, line 1108: **issue:** There's no reason to scope here since you just `.get()` immediately.

You can view this as being akin to doing a `vec.push(uri_string)` immediately followed by a `*vec.last().unwrap()`, where the `vec` is some vector that is somehow known to the garbage collector. (This is actually nearly exactly what happens :D ).

nova_vm/src/ecmascript/builtins/global_object.rs
```diff
@@ -1160,3 +1264,209 @@ impl GlobalObject {
...
+/// characters in preserveEscapeSet.
+///
+/// Adapted from Boa JS engine. Source https://github.com/boa-dev/boa/blob/ced222fdbabacc695f8f081c5b009afc9be6b8d0/core/engine/src/builtins/uri/mod.rs#L366
+///
+/// Copyright (c) 2019 Jason Williams
+fn decode<'gc, F>(
+    agent: &mut Agent,
+    string: String,
+    reserved_set: F,
+    gc: GcScope<'gc, '_>,
```
aapoalas, line 1283: **issue:** You could just take a `NoGcScope` here directly, I believe.

tests/expectations.json
aapoalas: I'll take a look.

aapoalas (changes requested):
Some changes needed, nothing major, looks really good in general! I'll take a look at the failing tests, see what is up with that, and I'll also fix the comments while I'm at it myself.

Thank you! <3

tests/expectations.json
aapoalas: Alright, it was a fairly simple case of using the `string.len(agent)` function, ie. getting the UTF-8 length, but then using `string.utf16_char(agent, index)` function to get a char which expects a UTF-16 index.

I refactored the code a little to use more UTF-8 instead of unnecessarily converting into UTF-16 quite so much. Probably some of the code is now more incorrect than it was, but none of the tests fail so I guess it's good.

aapoalas (approved):

PR #701, opened 2025-05-17
feat(ecmascript): %TypedArray%.prototype.toSorted

yossydev:
- doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.tosorted
- ref: https://github.com/trynova/nova/issues/145


tests/expectations.json
```diff
@@ -8981,11 +8981,7 @@
   "built-ins/TypedArray/prototype/toLocaleString/this-is-not-typedarray-instance.js": "FAIL",
   "built-ins/TypedArray/prototype/toLocaleString/user-provided-tolocalestring-grow.js": "FAIL",
   "built-ins/TypedArray/prototype/toLocaleString/user-provided-tolocalestring-shrink.js": "FAIL",
-  "built-ins/TypedArray/prototype/toSorted/comparefn-not-a-function.js": "FAIL",
   "built-ins/TypedArray/prototype/toSorted/ignores-species.js": "FAIL",
```
yossydev, line 8984: This test is failing because of an assert that checks whether the source and destination ArrayBuffers point to different memory locations. In other words, the test fails because they are actually pointing to the same buffer, which shouldn’t happen.
You can see the failing assertion here:
https://github.com/trynova/nova/blob/bd489c4cc5afc5b86df56be0d94ae62233b576ad/nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs#L3054-L3060

This issue didn’t occur with .with(), but I honestly don’t yet understand why it happens with toSorted.

aapoalas: The reason was that both the slices were empty slices, and those should be allowed to compare equal as pointers.

I added a separate check for that, and cleaned up some unnecessary Results etc.

aapoalas:
Thank you, looks really good to me! And good job, really smooth going <3

PR #702, opened 2025-05-17
feat(ecmascript): Implement `String.raw`

eliassjogreen:
- **feat(ecmascript): Implement `String.raw`**
- **chore: Update metrics and expectations**


aapoalas:
I fixed the crash with the one tests, that was about using `arguments[1..]` without checking that the arguments actually has at least 1 element, and changed from using a Vec of scoped Values for the arguments scoping, to using the builtin ArgumentsList scoping that I set up some time ago.

PR #729, opened 2025-05-21
fix(ecmascript): Behaviour of `Array.prototype.includes` larger than length `fromIndex` argument

eliassjogreen:
Just going through the expectations and doing some easy fixes 😅


aapoalas (approved):
LGTM, nice find <3

PR #731, opened 2025-05-22
fix(ecmascript): Add max size (4gb) to array buffer

eliassjogreen:
- **fix(ecmascript): Add max size (4gb) to array buffer**
- **chore: Update metrics and expectations**


aapoalas (approved):
LGTM. Maybe add a note that eventually we should use the currently-experimental Allocator APIs like https://doc.rust-lang.org/beta/std/alloc/struct.Global.html#method.allocate and check throw an error on an Err result.

PR #741, opened 2025-05-27
chore: bump oxc 0.72.0

load1n9:


nova_vm/src/heap/element_array.rs
```diff
@@ -1793,60 +1793,48 @@ impl ElementArrays {
...
-                            }
-                        }
-                    }
-                    (PropertyKey::String(a), PropertyKey::String(b)) => {
-                        assert!(a.get_index() == b.get_index())
-                    }
-                    (PropertyKey::Symbol(a), PropertyKey::Symbol(b)) => {
-                        assert!(a.get_index() == b.get_index())
+        // Check at runtime that PropertyKey can be transmuted into Option<PropertyKey> without issues and will show up as Some.
+        debug_assert!(
```
aapoalas, line 1797: **question:** Why did this go into a debug_assert? These tests shouldn't really be done on every call of this function, only at compile time if at all possible.

The checks should actually be moved out to a different function as well, honestly.

tests/expectations.json
```diff
@@ -1009,6 +1009,7 @@
   "built-ins/Function/prototype/toString/getter-class-statement-static.js": "FAIL",
   "built-ins/Function/prototype/toString/getter-class-statement.js": "FAIL",
   "built-ins/Function/prototype/toString/getter-object.js": "FAIL",
+  "built-ins/Function/prototype/toString/line-terminator-normalisation-LF.js": "FAIL",
```
aapoalas, line 1012: **question:** Can you check why this fails now?

aapoalas (dismissed):
LGTM aside from the const->debug_assert change. Thank you! <3

PR #744, opened 2025-06-01
feat(ecmascript): %TypedArray%.prototype.set

yossydev:
I think there are still a few things we can do.

doc: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.set

### DIFF
--- a/nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
+++ b/nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
@@ -3353,14 +3353,14 @@ pub(crate) fn split_typed_array_buffers<'a, T: Viewable>(
     let target_slice = target.as_mut_slice(agent);
     let target_slice =
         byte_slice_to_viewable_mut::<T>(target_slice, target_byte_offset, target_byte_limit);
-    let targt_byte_length = core::mem::size_of_val(target_slice);
+    let target_byte_length = core::mem::size_of_val(target_slice);
     let target_ptr = target_slice.as_mut_ptr();
     let target_len = target_slice.len();
     let source_slice = source.as_mut_slice(agent);
     let source_slice = byte_slice_to_viewable_mut::<T>(
         source_slice,
         source_byte_offset,
-        source_byte_offset + targt_byte_length,
+        source_byte_offset + target_byte_length,
     );
     // SAFETY: Confirmed beforehand that the two ArrayBuffers are in separate memory regions.
     let target_slice = unsafe { std::slice::from_raw_parts_mut(target_ptr, target_len) };


nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3369,3 +3390,77 @@ fn subarray_typed_array<'a, T: Viewable>(
...
+    if target_offset.is_negative() {
+        return Err(agent.throw_exception_with_static_message(
+            ExceptionType::RangeError,
+            "invalid array length",
+            gc.into_nogc(),
+        ));
+    }
+    // 6. If source is an Object that has a [[TypedArrayName]] internal slot, then
+    if scoped_source.get(agent).is_object() {
+        if let Ok(source) = TypedArray::try_from(scoped_source.get(agent)) {
```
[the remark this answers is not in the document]
aapoalas: Been eagerly waiting for these :)

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3369,3 +3390,77 @@ fn subarray_typed_array<'a, T: Viewable>(
...
+            )
+            .unbind()?;
+        } else {
+            // 7. Else,
+            //  a. Perform ? SetTypedArrayFromArrayLike(target, targetOffset, source).
+            with_typed_array_viewable!(
+                scoped_o.get(agent),
+                set_typed_array_from_array_like::<T>(
+                    agent,
+                    scoped_o.get(agent),
```
aapoalas, line 3442: **issue:** Since you already have `scoped_o` as Scoped, it doesn't make sense to `get` its result and then immediately `scope` it again inside the `SetTypedArrayFromX` calls: Just pass the `Scoped<TypedArray>` directly instead, and that'll save you one unnecessary scoping.

nova_vm/src/ecmascript/builtins/indexed_collections/typed_array_objects/typed_array_intrinsic_object.rs
```diff
@@ -3369,3 +3390,77 @@ fn subarray_typed_array<'a, T: Viewable>(
...
+                    target_offset,
+                    scoped_source.get(agent),
```
aapoalas, line 3444: **issue:** Here too, you can just pass in the `scoped_source: Scoped<Value>`  directly; inside the call you'll need to do `let src = source.replace_self(agent, src);` after you've done `let src = to_object(agent, source.get(agent), gc.nogc());` but otherwise it should be all good.

aapoalas (changes requested):
A bit more work to do around the GC handling here. You could also use the existing optimised TypedArray copy functions here instead of using the generic `GetValueFromBuffer` and `SetValueFromBuffer` methods.

aapoalas (approved):
Alright, LGTM! I did a few minor fixes and refactorings, most importantly moving the CloneArrayBuffer out of the if-else branches.

There is an optimisation to be made there; if the two ABs are the same, then
a) if the TypedArrays point to different parts of it, we could just use copy-within. And actually even if there is overlap, I think we can still use `copy_within` if we have the same types.

Even if we can't use `copy_within`, the "scratch buffer" that we create using `CloneArrayBuffer` is actually unnecessary per-se: we could simply create a `Box<[SrcType]>` instead; that avoid storing anything on the engine heap and the scratch buffer would be dropped synchronously after the set call is done. But! That sort of optimisation can wait for a better day.

PR #747, opened 2025-06-02
fix(test262): Include Annex-B and Intl in skipped tests

### LINKED ISSUE #743: website's test262 page is misleading
Nova [skips all of Intl and Annex B](https://github.com/trynova/nova/blob/main/tests%2Ftest262_runner.rs#L39) but presents this percentage as its Test262 score on the website which misleads people :(

eliassjogreen:
Closes #743 and properly reports the correct count of skipped tests. Also updated the test262 submodule while I was at it.


aapoalas (approved):
Awesome, thank you <3

PR #756, opened 2025-06-08
fix: Number.greaterThan

DonIsaac:


nova_vm/src/ecmascript/types/language/number.rs
```diff
@@ -599,7 +599,7 @@ impl<'a> Number<'a> {
     }
 
     pub fn greater_than(agent: &mut Agent, x: Self, y: Self) -> Option<bool> {
-        Number::less_than(agent, y, x).map(|x| !x)
+        Number::less_than(agent, x, y).map(|x| !x)
```
DonIsaac, line 602: I think this won't work if they're equal

Issue #722, opened 2025-05-21
Implement Object ...rest assignment


aapoalas:
This crashes at `nova_vm/src/engine/bytecode/bytecode_compiler/assignment.rs:248:13:` for 93 tests.

komyg:
Hey, I'd like to contribute to this bug. Can you assign the issue to me?

Also, if it isn't too much trouble, can you tell me one or two of the 93 tests that are failing?

aapoalas:
Hey, thank you for the interest and welcome! I assigned the issue to you; an example test is `language/statements/for-await-of/async-func-decl-dstr-obj-rest-descriptors.js` and `language/statements/for-of/dstr/obj-rest-getter.js`.

Probably looking for "rest" or "obj-rest" in the `exceptions.json` will also find many more examples.

PR #758, opened 2025-06-14
Implement Object ...rest Assignment in for of loop

komyg:
Issue: https://github.com/trynova/nova/issues/722


load1n9 (changes requested):
LGTM but fix the metrics file

load1n9 (approved):
LGTM

komyg:
Hey @aapoalas thank you for the review!

I did the changes you requested and now the code is creating a new object for the rest assignment, and it is also excluding any existing properties.

For example, the code below will throw at `Y Possible`:

```js
const source = { x: 1, y: 2 };
let target = {};
let x;
({ x, ...target } = source);
if (target === source) {
  throw 'impossible';
}

if (target.x === 1) {
  throw 'X Impossible';
}

if (target.y === 2) {
  throw 'Y Possible';
}
```

However my changes don't feel very right to me, because I had to change the implementation of `AssignmentTargetPropertyIdentifier` to push the references of the processed properties into the reference stack, so that I could tell the `CopyDataPropertiesIntoObject` instruction which properties to omit.

This solves the problem you raised, but if I don't have a `rest` assignment, then I am adding unnecessary elemetns to the reference stack.

So I think there might be a better way to do this.

aapoalas:
> Hey @aapoalas thank you for the review!
> 
> I did the changes you requested and now the code is creating a new object for the rest assignment, and it is also excluding any existing properties.
> 
> For example, the code below will throw at `Y Possible`:
> 
> ```js
> const source = { x: 1, y: 2 };
> let target = {};
> let x;
> ({ x, ...target } = source);
> if (target === source) {
>   throw 'impossible';
> }
> 
> if (target.x === 1) {
>   throw 'X Impossible';
> }
> 
> if (target.y === 2) {
>   throw 'Y Possible';
> }
> ```
> 
> However my changes don't feel very right to me, because I had to change the implementation of `AssignmentTargetPropertyIdentifier` to push the references of the processed properties into the reference stack, so that I could tell the `CopyDataPropertiesIntoObject` instruction which properties to omit.
> 
> This solves the problem you raised, but if I don't have a `rest` assignment, then I am adding unnecessary elemetns to the reference stack.
> 
> So I think there might be a better way to do this.

Alright, I did a bit of a refactor on this: I had to bring in some of the `impl Compile` inline to get the necessary control in to inject the `GetValueKeepReference && PushReference` conditionally instead of unconditionally. (I then proceeded to outline them again, but this time into separate helper functions that would allow me to inject the `has_rest` condition as a parameter.)

All tests that mention `obj-rest` now pass. I think this is good to go now :)

Issue #727, opened 2025-05-21
Test failures related to `Array.prototype.copyWithin`


eliassjogreen:
These two test failures seem to indicate an off-by-one or similar error in the array `copyWithin` method. It probably needs a taking look at but should be fairly easy and fast to fix.

<details>
<summary>

### <code>"built-ins/Array/prototype/copyWithin/coerced-values-start-change-target.js": "FAIL"</code>

</summary>

```
[...]

PR #766, opened 2025-06-21
fix(vm): Fix element deletion when reducing array length

edwinlzs:
Resolves `built-ins/Array/prototype/copyWithin/coerced-values-start-change-target.js` under #727 

It was indeed caused by an off-by-one error, not in the `copyWithin` method itself but rather during the deletion of elements when assigning a smaller length to an array.

Thus this fix also happens to pass a couple other tests related to reducing array length under `Object/defineProperties`!


edwinlzs:
Hold on, I just realized the same error likely exists in `array_try_set_length` as well, let me fix that too.

aapoalas (approved):
LGTM! Thank you, and a very nice find! <3

edwinlzs:
You're welcome! Thanks for your work, I'm learning a lot from looking into this project!

aapoalas:
Alright, looks like all the tests and whatnot pass as well, so I'm merging this. Thank you again, and welcome to the project! Great work <3

PR #767, opened 2025-06-23
fix(vm): use correct value for Number.MIN_VALUE

lino-levan:
Rust's `f64::MIN` is the biggest negative number. JavaScript's `Number.MIN_VALUE` is the smallest representable positive number.


aapoalas (approved):
Oops :D

PR #773, opened 2025-06-25
Handle negative numbers correctly in `Number::is_odd_integer`

simonwuelker:
Using the modulo operator gives incorrect results for negative numbers, we should use `rem_euclid` instead.

See https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6c04fdf5c03305201cdb801f23198b4c for reference.

Fixes four test262 tests.


aapoalas (commented):
LGTM, nice little fix! Thank you for fixing this, and welcome to the project! :)

PR #790, opened 2025-07-08
fix(linter): Update dylint and related dependencies

eliassjogreen:


aapoalas:
> Seems to be working so you're welcome to take a look @aapoalas. Locally I got an error because I apparently lacked a zlib install in the location expected of me (nix...) but seems to work fine on CI and normal linux at least.

Works locally for me as well, building and running wise.

I tried out if this would still detect a mistaken function and it doesn't seem to; I created a `fn bad_get` in `operations_on_objects.rs` that just swapped `fn get`'s first and second parameter around, but the linter stayed silent even after I switched another function to use that version (ie. it definitely wasn't dead code, although that shouldn't matter for a linter anyway).

I wonder if the linter is still properly linting our code?

aapoalas (approved):
LGTM! Merging, thank you very much for getting this working <3

PR #793, opened 2025-07-10
fix(small_string)!: SmallString::from_str_unchecked

DonIsaac:
## What This PR Does
- fix: several calls to `SmallStr::from_str_unchecked` were checking string length (not byte length) before calling `SmallStr::from_str_unchecked`. This would cause panics if given a UTF-8 string with <= 7 characters but >= 7 bytes.
- perf!: use `assert_unchecked` in `SmallStr::from_str_unchecked` and `SmallStr::from_wtf8_unchecked` to elide panic branch when given too-large strings. This required a contract change; these functions are now `unsafe`.
  > note: this lines up with other `from_<foo>_unchecked` APIs on `core::str`
- feat: implement `PartialEq` for `str` and `&str`
- refactor: use a `MAX_LEN` constant instead of referring to the magic number `7` everywhere

### DIFF
--- a/nova_vm/build.rs
+++ b/nova_vm/build.rs
@@ -70,9 +70,9 @@ fn gen_builtin_strings() -> io::Result<Vec<u8>> {
         output.push_str("    r#");
         output.push_str(&replace_invalid_key_characters(string));
         if SmallString::try_from(string.as_str()).is_ok() {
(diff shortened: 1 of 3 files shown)


aapoalas (approved):
LGTM, thank you! <3 Really nice additions/changes!

PR #794, opened 2025-07-10
fix(vm): regex parsing

DonIsaac:
## What This PR Does
Fixes two related bugs in `RegExpConstructor`
- oxc's regex parser expects patterns to be a string literal. This PR wraps patterns in double quotes before being parsed.
- empty flags should be `None`, not `Some("")`. Oxc's parser throws an error when flags are an empty string

### Some Notes
wrapping patterns in double quotes unconditionally is both bad for performance and possibly incorrect when patterns also contain a double quote. Ideally we'd make a PR into `oxc_regular_expression` to suppport non-literals.


nova_vm/src/ecmascript/builtins/regexp/abstract_operations.rs
```diff
@@ -196,16 +196,22 @@ pub(crate) fn reg_exp_initialize<'a>(
...
     // 13. Let parseResult be ParsePattern(patternText, u, v).
-    match ConstructorParser::new(
-        &allocator,
-        &p.to_string_lossy(agent),
-        Some(&f_str),
-        Default::default(),
-    )
-    .parse()
-    {
+    let pattern_text = if p.len(agent) == 0 {
```
aapoalas, line 207: **thought:** I wonder if we should actually be using the `LiteralParser` after all? Does a "cooked" string used in a RegExp constructor like `const string = a + b; new RegExp(string)` become equivalent to a literal?

nova_vm/src/ecmascript/builtins/regexp/abstract_operations.rs
```diff
@@ -196,16 +196,22 @@ pub(crate) fn reg_exp_initialize<'a>(
         Ok(f) => f.as_str().into(),
         Err(f) => f.to_string_lossy(agent),
     };
+    let flags: Option<&str> = if f_str.is_empty() {
```
aapoalas, line 199: **praise:** This is a nice catch <3

tests/metrics.json
```diff
@@ -1,10 +1,10 @@
 {
   "results": {
     "crash": 105,
-    "fail": 9366,
-    "pass": 37783,
+    "fail": 9201,
```
aapoalas, line 4: **praise:** This is a really nice K/D ratio on a PR! <3

aapoalas (approved):
LGTM, though I wonder if using `ConstructorParser` at all is a mistake?

nova_vm/src/ecmascript/builtins/regexp/abstract_operations.rs
aapoalas: It seems like using LiteralParser is indeed better; I got a few extra passes with that.

PR #797, opened 2025-07-22
Implement Promise.all

komyg:


nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -212,11 +216,96 @@ impl PromiseConstructor {
...
+                ExceptionType::TypeError,
+                "First element is not a Promise",
+                gc.into_nogc(),
+            ));
+        };
+
+        let result_capability = PromiseCapability::new(agent, gc.nogc());
+        let result_promise = result_capability.promise().scope(agent, gc.nogc());
+
+        let result_callback_closure: for<'a, 'b, 'c, 'd, 'e, '_gc> fn(
```
aapoalas, line 275: **note:** This isn't actually a closure; it doesn't "close over" the result_promise or anything like that; this is just a normal function. That's why you can put it in `Behaviour::Regular`.

That being said, I don't currently plan on supporting Rust closures as builtin functions as they have some distinct issues related to garbage collection associated with them. Instead of having an actual closure, you'll want to go the `PromiseReactionHandler::PromiseAll` route together with a new heap vector in `struct Heap` that stores the data you need to "close over", which the `PromiseReactionHandler::PromiseAll` then holds a "reference" (just a u32 index newtype that implements `HeapMarkAndSweep`) to.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -212,11 +216,149 @@ impl PromiseConstructor {
...
-        _this_value: Value,
-        _arguments: ArgumentsList,
-        gc: GcScope<'gc, '_>,
+        this_value: Value,
+        arguments: ArgumentsList,
+        mut gc: GcScope<'gc, '_>,
     ) -> JsResult<'gc, Value<'gc>> {
-        Err(agent.todo("Promise.all", gc.into_nogc()))
+        // 1. Let C be the this value.
+        if this_value
```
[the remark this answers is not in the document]
komyg: Yes, this is mostly WIP at this point. I will mark it more clearly next time.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_all_record.rs
```diff
@@ -0,0 +1,152 @@
...
+    engine::context::{Bindable, GcScope, NoGcScope},
+    heap::{
+        CompactionLists, CreateHeapData, Heap, HeapMarkAndSweep, WorkQueues, indexes::BaseIndex,
+    },
+};
+
+#[derive(Debug, Clone, Copy)]
+pub struct PromiseAllRecordHeapData<'a> {
+    pub remaining_unresolved_promise_count: u32,
+    pub result_array: Array<'a>,
```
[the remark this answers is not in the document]
komyg: So far I am using `Array` just to make it work, but I can change to `ElementsVector` if you prefer.

What is the difference between the two?

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_all_record.rs
```diff
@@ -0,0 +1,164 @@
...
+        let promise_all_index = self.bind(gc.nogc());
+        let value = value.bind(gc.nogc());
+
+        let result_array = {
+            let data = promise_all_index.get_mut(agent);
+            data.result_array.bind(gc.nogc())
+        };
+
+        let elements = result_array.as_mut_slice(agent);
+        elements[index as usize] = Some(value.unbind());
```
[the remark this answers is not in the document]
komyg: This breaks, because I was originally pre-filling the `result_array` with `Value::Undefined`, so the previous value is always: `prev: Some(Undefined)` and the debug assertion breaks.

komyg:
Hey @aapoalas, I've resolved your comments, and I also fixed the `promise_constructor`.

The code does not implement the whole spec yet, but I think it is at a good starting point.

Can you review again, please?

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_all_record.rs
```diff
@@ -0,0 +1,138 @@
...
+    engine::context::{Bindable, GcScope, NoGcScope, bindable_handle},
+    heap::{
+        CompactionLists, CreateHeapData, Heap, HeapMarkAndSweep, WorkQueues, indexes::BaseIndex,
+    },
+};
+
+#[derive(Debug, Clone, Copy)]
+pub struct PromiseAllRecord<'a> {
+    pub(crate) remaining_unresolved_promise_count: u32,
+    pub result_array: Array<'a>,
```
aapoalas, line 23: **note:** None of these fields should be `pub`; at most `pub(crate)`. This is an internal implementation detail of the engine, after all.

aapoalas (changes requested):
The code mostly looks good (just one GC lifetime / use-after-free issue), but it's not exactly what the spec says the method should do. Maybe we could pair program this over the finish line?

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -211,18 +221,209 @@ impl PromiseConstructor {
...
+
+        let Some(iterator_record) = iterator_record.into_iterator_record() else {
+            return Err(throw_not_callable(agent, gc.into_nogc()));
+        };
+
+        // 8. If result is an abrupt completion, then
+        //         a. If iteratorRecord.[[Done]] is false, set result to Completion(IteratorClose(iteratorRecord, result)).
+        //         b. IfAbruptRejectPromise(result, promiseCapability).
+        // 9. Return ! result.
+        Self::perform_promise_all(
```
aapoalas, line 419: **issue:** Step 7 should be this:

> Let result be [Completion](https://tc39.es/ecma262/#sec-completion-ao)([PerformPromiseAll](https://tc39.es/ecma262/#sec-performpromiseall)(iteratorRecord, C, promiseCapability, promiseResolve)).

So, what this would look like here is something like this:

```rust
let result = Self::perform_promise_all(agent, iterator_record.unbind(), c.unbind(), promise_capability.unbind(), promise_resolve.unbind(), gc.reborrow()).unbind().bind(gc.nogc())
```

Then steps 8 and 9 need to be done:

```rust
// 8. If result is an abrupt completion, then
match result {
  Err(err) => {
    // a. If iteratorRecord.[[Done]] is false, set result to Completion(IteratorClose(iteratorRecord, result)).
    // todo: the error result somehow needs to tell us if iterator_record reached done; this could perhaps be an extra `iterator_done: &mut bool` parameter passed to `perform_promise_all` that is set to `true`.
    let result = if !iterator_done {
      let iterator_record = ...; // get scoped iterator_record basically.
      iterator_close_with_error(agent, iterator_record, result, gc.reborrow()).unbind().bind(gc.nogc())
    } else {
      Err(result)
    };
    let result = if_abrupt_reject_promise_m!(agent, result, promise_capability, gc);
    // 9. Return ! result.
    Ok(result)
  },
  Ok(result) => {
    // 9. Return ! result.
    Ok(result)
  },
}
```

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -211,18 +221,209 @@ impl PromiseConstructor {
...
+        mut gc: GcScope<'gc, '_>,
+    ) -> JsResult<'gc, Value<'gc>> {
+        let iterator_record = iterator_record.bind(gc.nogc());
+        let constructor = constructor.scope(agent, gc.nogc());
+        let result_capability = result_capability.bind(gc.nogc());
+        let promise_resolve = promise_resolve.scope(agent, gc.nogc());
+
+        let iterator = iterator_record.iterator.scope(agent, gc.nogc());
+        let next_method = iterator_record.next_method.scope(agent, gc.nogc());
+        //     1. Let values be a new empty List.
```
aapoalas, line 240: **nitpick:** Unnecessary indentation on the comment.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -211,18 +221,209 @@ impl PromiseConstructor {
         Ok(scoped_promise.get(agent).into_value())
     }
 
+    /// ### [27.2.4.1.2 PerformPromiseAll ( iteratorRecord, constructor, resultCapability, promiseResolve )](https://tc39.es/ecma262/#sec-performpromiseall)
+    fn perform_promise_all<'gc>(
+        agent: &mut Agent,
+        iterator_record: IteratorRecord,
+        constructor: Function,
```
aapoalas, line 228: **suggestion:** Just take `Scoped<Function>` as a parameter instead.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -211,18 +221,209 @@ impl PromiseConstructor {
...
+        result_capability: PromiseCapability,
+        promise_resolve: Function,
```
aapoalas, line 230: **suggestion:** Likewise, just take a `Scoped<Function>` instead.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -211,18 +221,209 @@ impl PromiseConstructor {
         Ok(scoped_promise.get(agent).into_value())
     }
 
+    /// ### [27.2.4.1.2 PerformPromiseAll ( iteratorRecord, constructor, resultCapability, promiseResolve )](https://tc39.es/ecma262/#sec-performpromiseall)
+    fn perform_promise_all<'gc>(
+        agent: &mut Agent,
+        iterator_record: IteratorRecord,
```
aapoalas, line 227: **suggestion:** You can probably also take the iterator and next_method as Scoped values instead of an unscoped `IteratorRecord`: the calling function needs to be able to call `IteratorClose` on `IteratorRecord` as well so it makes sense to scope the two values there and pass them as parameters, instead of scoping them both outside and inside the function.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -211,18 +221,209 @@ impl PromiseConstructor {
...
+            .create(promise_all.unbind())
+            .scope(agent, gc.nogc());
+
+        //     3. Let index be 0.
+        let mut index = 0;
+
+        //     4. Repeat,
+        loop {
+            let iterator_record = IteratorRecord {
+                iterator: iterator.get(agent),
```
aapoalas, line 273: **issue:** Missing binds here; best practice to always bind get results unless you're passing them immediately into a function that takes `GcScope`. Here `iterator_record` is now unbound and could end up being use-after-free.

Alternatively, do what you did with `promise_all` above which was to `bind` the struct itself instead of its field values.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -211,18 +221,209 @@ impl PromiseConstructor {
...
+                .unbind()?
+                .bind(gc.nogc());
+
+            // b. If next is done, then
+            let Some(next) = next else {
+                break;
+            };
+
+            // c. Append undefined to values.
+            if let Err(err) = result_array.get(agent).reserve(agent, 1) {
```
aapoalas, line 287: **nitpick:** You can avoid one `result_array.get(agent)` by assigning `let temp_array = result_array.get(agent).bind(gc.nogc())` and using that in for both `reserve` and `set_len`.

The compiler is fairly unlikely to realise it can deduplicate the `get` calls, so we should deduplicate them for it.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -211,18 +221,209 @@ impl PromiseConstructor {
...
+            // f. Let length be the number of non-optional parameters of the function definition in Promise.all Resolve Element Functions.
+            // g. Let onFulfilled be CreateBuiltinFunction(steps, length, "", « [[AlreadyCalled]], [[Index]], [[Values]], [[Capability]], [[RemainingElements]] »).
+            // h. Set onFulfilled.[[AlreadyCalled]] to false.
+            // i. Set onFulfilled.[[Index]] to index.
+            // j. Set onFulfilled.[[Values]] to values.
+            // k. Set onFulfilled.[[Capability]] to resultCapability.
+            // l. Set onFulfilled.[[RemainingElements]] to remainingElementsCount.
+            // m. Set remainingElementsCount.[[Value]] to remainingElementsCount.[[Value]] + 1.
+            // n. Perform ? Invoke(nextPromise, "then", « onFulfilled, resultCapability.[[Reject]] »).
+            // o. Set index to index + 1.
```
aapoalas, line 319: **nitpick:** I'd move the comments to colocate with the code that performs then, ie. this'd go before line 339 and the `m. Set remainingElementsCount.[[Value]] ...` would go before line 341.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -218,11 +230,186 @@ impl PromiseConstructor {
...
+        let constructor = constructor.scope(agent, gc.nogc());
+        let promise_capability = PromiseCapability::new(agent, gc.nogc());
+        let promise = promise_capability.promise().scope(agent, gc.nogc());
+
+        // 3. Let promiseResolve be Completion(GetPromiseResolve(C)).
+        let promise_resolve = get_promise_resolve(agent, constructor.get(agent), gc.reborrow());
+
+        // 4. IfAbruptRejectPromise(promiseResolve, promiseCapability).
+        let promise_capability = PromiseCapability {
+            promise: promise.get(agent),
```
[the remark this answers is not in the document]
aapoalas: Aight, that's an unfortunate effect of the reborrowing system using exclusive references `&mut` as its model: you need to do `let promise_resolve = get_promise_resolve(agent, constructor.get(agent), reborrow()).unbind().bind(gc.nogc());`. The `.unbind().bind(gc.nogc())` "dance" means that we take the `JsResult<'gc, Function<'gc>>` we got from the `get_promise_resolve` which is bound to the `GcScope` as _exclusive_, ie. no one is allowed to touch it for as long as the JsResult exists, and we create a new `JsResult<'static, Function<'static>>` from that and forget the original one. Now `GcScope` can be touched again, and we use that to create a new `JsResult<'gc, Function<'gc>>` from the static one but now using `gc.nogc()` which uses `GcScope` as _shared_, ie. others are also allowed to use it as long as they use it as shared as well.

This is the same "problem" that is why you cannot access a `Vec<u32>`'s data if you pass it through a function like `fn get_mut(vec: &Vec<u32>) -> &u32 { return &vec[0]; }`, and the reason for why Rust works like this has to do with `Cell` and other interior-mutable types.

aapoalas (changes requested):
Still needs a few fixes here or there, but I think this is generally good to merge now! Great work and thank you for sticking it out! <3

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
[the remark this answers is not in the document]
komyg: Ok, I think I understand now! Thank you for the explanation!

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_constructor.rs
```diff
@@ -211,18 +224,261 @@ impl PromiseConstructor {
...
     }
 
+    /// ### [27.2.4.1.2 PerformPromiseAll ( iteratorRecord, constructor, resultCapability, promiseResolve )](https://tc39.es/ecma262/#sec-performpromiseall)
+    fn perform_promise_all<'gc>(
+        agent: &mut Agent,
+        iterator_record: IteratorRecord,
+        constructor: Scoped<Function>,
+        result_capability: PromiseCapability,
+        promise_resolve: Scoped<Function>,
+        _iterator_done: &mut bool,
```
aapoalas, line 234: **nitpick:** Underscore prefixing is for unused variables; this is properly used and thus shouldn't have an underscore prefix.

PR #803, opened 2025-07-29
ci(release): Add linux arm target

eliassjogreen:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/#how-to-use-the-runners


aapoalas (approved):
Absolutely! <3

I should do another release...

PR #811, opened 2025-08-03
feat(vm): enums support

load1n9:


nova_vm/src/engine/bytecode/vm.rs
```diff
@@ -3185,6 +3184,22 @@ impl Vm {
                     ));
                 }
             }
+            #[cfg(feature = "typescript")]
+            Instruction::CreateTSEnum => {
```
aapoalas, line 3188: **issue:** This seems 100% equivalent to `Instruction::ObjectCreate`, maybe with an `Instruction::Load` added in.

Given that TS compiles down to JS, it should always be possible to implement all TS in pure bytecode, and we should aim to do so.

aapoalas: You may also want to look at `Instruction::ObjectCreateWithShape` for simple enums. If that seems useful to you, then it may also be a good idea to create the object shape as an "intrinsic" object shape (the name is bad, it should rather be "unique" or something). What that means is that the shape is "stand-alone", unique and cannot be recreated by individually adding properties to an empty object (with the same prototype). This saves memory when we expect an object with this shape to only exist once.

PR #828, opened 2025-08-20
feat(vm): implement typescript type assertion and ts instantiation expressions

load1n9:


nova_vm/src/engine/bytecode/bytecode_compiler.rs
```diff
@@ -2167,10 +2167,21 @@ impl<'a, 's, 'gc, 'scope> CompileEvaluation<'a, 's, 'gc, 'scope> for ast::Expres
...
                 None
             }
-            ast::Expression::JSXElement(_)
-            | ast::Expression::JSXFragment(_)
-            | ast::Expression::TSTypeAssertion(_)
-            | ast::Expression::TSInstantiationExpression(_) => unreachable!(),
+            #[cfg(feature = "typescript")]
+            ast::Expression::TSTypeAssertion(x) => {
+                x.expression.compile(ctx);
+                None
```
aapoalas, line 2173: **issue:** My own fool fault, but all of the TS expressions should actually just do `ast::Expression::TSFoo(x) => ts.expression.compile(ctx),` ie. they should return whatever the inner expression compilation returns.

aapoalas (dismissed):
LGTM! <3

PR #845, opened 2025-09-07
perf(engine): Optimize hot-paths of instruction methods

eliassjogreen:
For me these small changes pushed the performance of richards.js from ~95 to > ~100.


aapoalas (approved):
LGTM, thank you! <3

PR #862, opened 2025-09-18
feat(linter): Implement `immediately_bind_scoped` rule

eliassjogreen:
Needs #790 to be merged first.

Still a bit work in progress with some edge-cases which need ironing out, but it seems to mostly work now which is pretty cool.


aapoalas:
**praise:** Ooh this is really cool! <3

eliassjogreen:
Updated the lint to check the `take` method too and found two new issues 😉.

PR #872, opened 2025-10-04
Implement Promise All Settled

komyg:


nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
```diff
@@ -0,0 +1,292 @@
...
+        rootable::{HeapRootData, HeapRootRef, Rootable},
+    },
+    heap::{
+        CompactionLists, CreateHeapData, Heap, HeapMarkAndSweep, ObjectEntry, WorkQueues,
+        indexes::BaseIndex,
+    },
+};
+
+#[derive(Debug, Clone, Copy)]
+pub enum PromiseGroupType {
```
aapoalas: **praise:** Nice <3

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
```diff
@@ -0,0 +1,292 @@
...
+    ) {
+        let value = value.bind(gc);
+        let promise_all = self.bind(gc);
+        let data = promise_all.get_mut(agent);
+
+        let capability = PromiseCapability::from_promise(data.promise, true);
+        capability.reject(agent, value.unbind(), gc);
+    }
+
+    pub(crate) fn on_promise_all_settled_fulfilled(
```
aapoalas: **suggestion:** `allSettled` doesn't deserve two functions that do nearly the same thing: just have one function with a boolean parameter that controls what kind of `obj ` gets created.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_jobs.rs
```diff
@@ -284,20 +287,48 @@ impl PromiseReactionJob {
...
-                            index,
-                            argument.unbind(),
-                            gc.reborrow(),
-                        );
-                    }
-                    PromiseReactionType::Reject => {
-                        promise_all.on_promise_rejected(agent, argument.unbind(), gc.nogc());
-                    }
+                let record = promise_group.get(agent);
+                match record.promise_group_type {
```
aapoalas: **suggestion:** I'd move this inside of a general `promise_group.settle(agent, index, argument, reaction_type, gc)` handler. That then effectively becomes the one function shared between `allSettled` and `all` I suggested above. The only difference is that `all`'s reject side is there as well, and it's just on its own path.

aapoalas (changes requested):
Wow, this looks great! Only one real blocking nit to pick, and then a couple of suggestions on reducing some repetition in the resolving code.

Absolutely stellar stuff! <3

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
```diff
@@ -0,0 +1,298 @@
...
+    ) {
+        let value = value.bind(gc);
+        let promise_all = self.bind(gc);
+        let data = promise_all.get_mut(agent);
+
+        let capability = PromiseCapability::from_promise(data.promise, true);
+        capability.reject(agent, value.unbind(), gc);
+    }
+
+    pub(crate) fn on_promise_all_settled(
```
aapoalas: **issue:** I'd prefer unifying this function with `on_promise_all_fulfilled`; the only difference between this and that is creating the `obj`, and that can be easily extracted into a helper function that you call from the `settle` function before calling into `on_promise_all_fulfilled` with the result.

```rust
let value = match record.promise_group_type {
  PromiseGroupType::PromiseAll => match reaction_type {
    PromiseReactionType::Fulfill => {
      value
    }
    PromiseReactionType::Reject => {
      // A reject in Promise.all rejects the entire group immediately.
      self.reject(agent, value.unbind(), gc.into_nogc());
      return;
    }
  },
  PromiseGroupType::PromiseAllSettled => {
    self.to_settled_object(agent, reaction_type, value.unbind(), gc.nogc())
  }
};
self.fulfill(agent, index, value.unbind(), gc);
```

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    pub fn get_mut(self, agent: &mut Agent) -> &mut PromiseGroupRecord<'static> {
+        agent
+            .heap
+            .promise_group_records
+            .get_mut(self.get_index())
+            .expect("PromiseGroupRecord not found")
+    }
+
+    pub(crate) const fn _def() -> Self {
```
aapoalas: **nitpick:** This could maybe be a `const DEF: Self`.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
```diff
@@ -0,0 +1,298 @@
...
+    pub(crate) fn settle(
+        self,
+        agent: &mut Agent,
+        reaction_type: PromiseReactionType,
+        index: u32,
+        value: Value<'a>,
+        mut gc: GcScope<'a, '_>,
+    ) {
+        let value = value.bind(gc.nogc());
+        let record = self.get(agent);
```
aapoalas: **thought:** I think you could do a `self.get_mut(agent)` here and call the `take` method on the record even before matching on the group type: if this is a "reject immediately" type of group then we can even still reuse the `record` borrow to unconditionally access the `promise` in the `PromiseReactionType::Reject` branch; we'll have subtracted the remaining count by one "unnecessarily" but that's probably the rare case anyhow so it's fine.

aapoalas (changes requested):
Looks really damn good! But let's "deduplicate" the methods a bit more still; the PromiseAll / PromiseAllSettled handling should be exactly the same in both cases with PromiseAllSettled just wrapping the value in an object before calling the shared handling code.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
komyg: Hum... I would prefer to keep the code as is.

If we call `take` on this method, then we will have to pass the `result_array` and the `Option(promise)`  as arguments down to the `fulfill` and `reject` functions, and they already take many arguments...

In this case, I think that the code is clearer by getting these values where we need them in each of the function. Especially because `reject` doesn't care about the `result_array`.

aapoalas (dismissed):
LGTM, and great work! I have a feeling the next ones will be ridiculously easy with this work in place :)

aapoalas:
Oh, @komyg , the test results require updating.

komyg:
Hey @aapoalas, tests are passing now. 

I didn't merge the expectations updates before, because that always caused merge conflicts.

And yes, I intend to do `promise.any` next, and it should be pretty easy after what we built so far.

PR #874, opened 2025-10-13
Benches (Test CI)

niluxv:


niluxv:
Ah, yes, I did look at `boa`s benchmarks before I wrote the criterion benchmark. I added a notice at the top of the file (although I don't think my code was "derived work" in the legal sense, but I'm not a lawyer; fortunately legal considerations are irrelevant as their code is Unlicensed).

PR #876, opened 2025-10-19
Temporal intrinsic

sebastianjacmatt:


nova_vm/Cargo.toml
```diff
@@ -63,6 +65,7 @@ shared-array-buffer = ["array-buffer", "ecmascript_atomics"]
 weak-refs = []
 set = []
 typescript = []
+temporal = ["temporal_rs"]
```
aapoalas: **issue:** Feature is doubled.

nova_vm/src/ecmascript/builtins/temporal.rs
```diff
@@ -0,0 +1,250 @@
...
+            .unbind()
+            .to_string(agent, gc.reborrow())
+            .unbind()?
+            .as_bytes(agent)
+            != b"auto"
+        {
+            // b. Return auto.
+            return Ok(temporal_rs::parsers::Precision::Auto);
+        }
+        // SAFETY: not shared.
```
aapoalas: **issue:** You should be throwing an error here. "If not 'auto', throw a RangeError exception. Else, return 'auto'."

nova_vm/src/ecmascript/builtins/temporal.rs
```diff
@@ -0,0 +1,250 @@
...
+    }
+    if let Value::Integer(digits_value) = digits_value
+        && (0..=9).contains(&digits_value.into_i64())
+    {
+        return Ok(temporal_rs::parsers::Precision::Digit(
+            digits_value.into_i64() as u8,
+        ));
+    }
+    // 3. If digitsValue is not a Number, then
+    if !digits_value.is_number() {
```
aapoalas: **suggestion:**
```suggestion
    let Ok(digits_value) = Number::try_from(digits_value) else {
```

This should work to convert `digitsValue` into a number after this if branch; you just need to add the currently missing "throw RangeError" part inside the `else`.

aapoalas (changes requested):
A partial review.

nova_vm/src/ecmascript/builtins/temporal.rs
```diff
@@ -0,0 +1,250 @@
...
+/// The abstract operation GetDifferenceSettings takes arguments operation (since or until),
+/// options (an Object), unitGroup (date, time, or datetime), disallowedUnits (a List of Temporal units),
+/// fallbackSmallestUnit (a Temporal unit), and smallestLargestDefaultUnit (a Temporal unit) and returns either
+/// a normal completion containing a Record with fields [[SmallestUnit]] (a Temporal unit),
+/// [[LargestUnit]] (a Temporal unit), [[RoundingMode]] (a rounding mode),
+/// and [[RoundingIncrement]] (an integer in the inclusive interval from 1 to 10**9),
+/// or a throw completion. It reads unit and rounding options needed by difference operations.
+pub(crate) fn get_difference_settings<'gc, const IS_UNTIL: bool>(
+    agent: &mut Agent,
+    options: Object<'gc>,                 // options (an Object)
```
aapoalas: **suggestion:** No reason for the 'gc lifetime here.
```suggestion
    options: Object,                 // options (an Object)
```

nova_vm/src/ecmascript/builtins/temporal/duration.rs
```diff
@@ -0,0 +1,488 @@
+pub(crate) mod data;
```
aapoalas: **issue:** Missing MPLv2 headers.

nova_vm/src/ecmascript/builtins/temporal/duration/duration_constructor.rs
```diff
@@ -0,0 +1,181 @@
...
+        let new_target = new_target.bind(gc.nogc());
+        // 1. If NewTarget is undefined, throw a TypeError exception.
+        let Some(new_target) = new_target else {
+            return Err(agent.throw_exception_with_static_message(
+                ExceptionType::TypeError,
+                "calling a builtin Temporal.Duration constructor without new is forbidden",
+                gc.into_nogc(),
+            ));
+        };
+        let Ok(new_target) = Function::try_from(new_target.unbind()) else {
```
aapoalas: **note:** Here you do not want to unbind `new_target` before calling `try_from`: if left bound then the `Ok(new_target)` would still be bound to the GC lifetime and would be safe. If you didn't scope the result into a variable shadowing this name, it'd be possible for someone to accidentally use `new_target` after free.

nova_vm/src/ecmascript/builtins/temporal/duration/duration_constructor.rs
```diff
@@ -0,0 +1,181 @@
...
+                .unbind()?
+                .into_f64(agent) as i128
+        }
+        .bind(gc.nogc());
+        // 12. Return ? CreateTemporalDuration(y, mo, w, d, h, m, s, ms, mis, ns, NewTarget).
+        let duration = temporal_rs::Duration::new(y, mo, w, d, h, m, s, ms, mis, ns)
+            .map_err(|e| temporal_err_to_js_err(agent, e, gc.nogc()))
+            .unbind()?
+            .bind(gc.nogc());
+        create_temporal_duration(agent, duration.unbind(), Some(new_target.get(agent)), gc)
```
aapoalas: **note:** It shouldn't be necessary to unbind duration, as `temporal_rs::Duration` is just "plain old data" and does not carry the GC lifetime. In other words, types that are `trivially_bindable!` do not need unbind/bind at function call interfaces.

nova_vm/src/ecmascript/builtins/temporal/error.rs
```diff
@@ -0,0 +1,33 @@
+use crate::{
```
aapoalas: **nitpick:** Missing MPLv2 header.

nova_vm/src/ecmascript/builtins/temporal/instant.rs
```diff
@@ -0,0 +1,412 @@
...
+/// internal slots with valid values.
+fn create_temporal_instant<'gc>(
+    agent: &mut Agent,
+    epoch_nanoseconds: temporal_rs::Instant,
+    new_target: Option<Function>,
+    gc: GcScope<'gc, '_>,
+) -> JsResult<'gc, TemporalInstant<'gc>> {
+    // 1. Assert: IsValidEpochNanoseconds(epochNanoseconds) is true.
+    // 2. If newTarget is not present, set newTarget to %Temporal.Instant%.
+    let new_target = new_target.unwrap_or_else(|| {
```
aapoalas: **suggestion:** You could add a path for checking `new_target.is_none_or(|nt| nt == agent.current_realm_record().intrinsics().temporal_instant().into_function())`, ie. are we using the default constructor, and if so then just use the `CreateHeapData` directly. When there is no newTarget set or it is the default intrinsic, we know that the call is going to come out to just effectively a `CreateHeapData` with the default data, after which we use the `set_epoch_nanoseconds` to override the default data. We don't need to go that far and can just `CreateHeapData` with the `epoch_nanoseconds` value directly in that case.

nova_vm/src/ecmascript/builtins/temporal/instant/instant_constructor.rs
```diff
@@ -0,0 +1,241 @@
...
+    fn from<'gc>(
+        agent: &mut Agent,
+        _: Value,
+        args: ArgumentsList,
+        gc: GcScope<'gc, '_>,
+    ) -> JsResult<'gc, Value<'gc>> {
+        let item = args.get(0).bind(gc.nogc());
+        // 1. Return ? ToTemporalInstant(item).
+        let instant = to_temporal_instant(agent, item.unbind(), gc)?;
+        let instant = agent.heap.create(InstantRecord {
```
aapoalas: **praise:** Nice <3

nova_vm/src/ecmascript/builtins/temporal/instant/instant_prototype.rs
```diff
@@ -0,0 +1,696 @@
+use temporal_rs::{
```
aapoalas: **nitpick:** Missing MPLv2 header.

nova_vm/src/ecmascript/builtins/temporal/instant/instant_prototype.rs
```diff
@@ -0,0 +1,696 @@
...
+    ) -> JsResult<'gc, Value<'gc>> {
+        let duration = args.get(0).bind(gc.nogc());
+        // 1. Let instant be the this value.
+        let instant = this_value.bind(gc.nogc());
+        // 2. Perform ? RequireInternalSlot(instant, [[InitializedTemporalInstant]]).
+        let instant = require_internal_slot_temporal_instant(agent, instant.unbind(), gc.nogc())
+            .unbind()?
+            .bind(gc.nogc());
+        // 3. Return ? AddDurationToInstant(subtract, instant, temporalDurationLike).
+        const SUBTRACT: bool = false;
```
aapoalas: **praise:** Good idea to show the operation name like this <3

nova_vm/src/ecmascript/builtins/temporal/instant/instant_prototype.rs
```diff
@@ -0,0 +1,696 @@
...
+    fn since<'gc>(
+        agent: &mut Agent,
+        this_value: Value,
+        args: ArgumentsList,
+        gc: GcScope<'gc, '_>,
+    ) -> JsResult<'gc, Value<'gc>> {
+        let other = args.get(0).bind(gc.nogc());
+        let options = args.get(1).bind(gc.nogc());
+        // 1. Let instant be the this value.
+        let instant = this_value;
```
aapoalas: **issue (nonblocking):** `this_value` left unbound here. Not an issue since there's no possibility of a GC happening before it gets "moved" into the `difference_temporal_instant` call, but it's technically not quite correct :)

aapoalas (changes requested):
Alright, lots of good code but also a lot of little corners to file and polish.

But overall, this is quite good as-is already, so excellent work! There weren't any big issues, no overt GC lifetime problems that I detected, or anything like that. Just nice, clean (or what passes as clean in Nova :) ) good code <3

PR #883, opened 2025-11-05
Implement Promise Any

komyg:


nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
```diff
@@ -103,7 +112,64 @@ impl<'a> PromiseGroup<'a> {
...
+        let error = error.bind(gc.nogc());
+
+        let promise_group_record = promise_group.get_mut(agent);
+        let (result_array, promise_to_resolve) = promise_group_record.take_result_and_promise();
+
+        let elements = result_array.as_mut_slice(agent);
+        elements[index as usize] = Some(error.unbind());
+
+        if let Some(promise_to_resolve) = promise_to_resolve {
+            let aggregate_error = OrdinaryObject::create_object(
```
komyg: I am not sure if I am creating this error correctly.

I saw that we have some intrinsic objects, so maybe I should be using that instead of an ordinary object.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
```diff
@@ -103,7 +112,64 @@ impl<'a> PromiseGroup<'a> {
...
+                agent,
+                Some(
+                    agent
+                        .current_realm_record()
+                        .intrinsics()
+                        .aggregate_error_prototype()
+                        .into(),
+                ),
+                &[ObjectEntry::new_data_entry(
```
komyg: The spec also says that I should use `DefinePropertyOrThrow` when setting this. I hope this is the equivalent.

komyg:
Note: I will update the Test 262 statistics after the PR is approved, to avoid merge conflicts.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
```diff
@@ -64,20 +65,28 @@ impl<'a> PromiseGroup<'a> {
...
 
         match record.promise_group_type {
-            PromiseGroupType::PromiseAll => match reaction_type {
+            PromiseGroupType::All => match reaction_type {
                 PromiseReactionType::Fulfill => {
                     self.fulfill(agent, index, value.unbind(), gc.reborrow());
                 }
                 PromiseReactionType::Reject => {
-                    self.reject(agent, value.unbind(), gc.nogc());
+                    self.immediately_reject(agent, value.unbind(), gc.nogc());
```
aapoalas: **praise:** The method naming is now nice and uniform <3

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
aapoalas: **issue:** Right, so here's what you need to do: you need to create an `ErrorHeapData` containing an AggregateError kind and no message or reason, and then call the DefineProperty on it. Or if you want to be very optimal, you create the ordinary object you have here (although you need to add a "name" property with the value "AggregateError" in it) and the ErrorHeapData such that its `backing_object` points to that orinary object.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
```diff
@@ -103,7 +112,64 @@ impl<'a> PromiseGroup<'a> {
...
+        agent: &mut Agent,
+        index: u32,
+        error: Value<'a>,
+        gc: GcScope<'a, '_>,
+    ) {
+        let promise_group = self.bind(gc.nogc());
+        let error = error.bind(gc.nogc());
+
+        let promise_group_record = promise_group.get_mut(agent);
+        let (result_array, promise_to_resolve) = promise_group_record.take_result_and_promise();
```
aapoalas: **suggestion:** In the `take_result_and_promise` when you find the value to be 0, you could also check if the PromiseGroup is in the last slot on the heap and if so then pop it off the Vec (maybe also popping off any following ones that have a 0 on them).

That would help the GC (and you could optionally also subtract from the alloc counter, although I don't think that actually really makes sense), which is nice.

nova_vm/src/ecmascript/builtins/control_abstraction_objects/promise_objects/promise_abstract_operations/promise_group_record.rs
```diff
@@ -121,48 +187,36 @@ impl<'a> PromiseGroup<'a> {
...
-                    agent
-                        .current_realm_record()
-                        .intrinsics()
-                        .object_prototype()
-                        .into(),
+        let entries = match reaction_type {
+            PromiseReactionType::Fulfill => vec![
+                ObjectEntry::new_data_entry(
+                    BUILTIN_STRING_MEMORY.status.into(),
+                    BUILTIN_STRING_MEMORY.fulfilled.into(),
```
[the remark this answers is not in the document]
aapoalas: Oh, my bad, in that case let's keep it as one.

aapoalas (dismissed):
LGTM!

komyg:
> LGTM!

Thank you @aapoalas! 

I've added your suggestions.

PR #887, opened 2025-11-11
Implement Promise Race

komyg:


komyg:
Hey @aapoalas, in this implementation of `Promise.race` I am using the infrastructure that we already have for `PromiseGroup`. 

This means that we are creating a `PromiseGroupRecord` with all its elements, such as `result_array` and `remaining_elements_count`, which are not used by the `Promise.race` implementation:

```rs
pub struct PromiseGroupRecord<'a> {
    pub(crate) promise_group_type: PromiseGroupType,
    pub(crate) remaining_elements_count: u32,
    pub(crate) result_array: Array<'a>,
    pub(crate) promise: Promise<'a>,
}
```

We can keep it like this, or I can create a data structure specially for `Promise.race` that doesn't have any unused fields. This means also creating heap vectors to hold a `PromiseRaceRecord` struct. What do you prefer?

aapoalas (dismissed):
LGTM :)

komyg:
> Hmm.... Do we actually need any kind of PromiseGroup at all...? Or can we just use the normal PromiseResolve stuff?
>

Well, I used the `PromiseGroup`, because the initial part of the spec is exactly the same for `Promise.all`, `Promise.any`, `Promise.allSettled` and `Promise.race`. Plus, all of them take in an iterable as an argument.

The difference is in the _Perform Promise Race_ method, that doesn't use all the state that the other methods use. Even so, I am also using the `immediately_resolve` and `immediatelly_reject` functions from the `PromiseGroup`.

So if we can accept a slight more memory usage for `Promise.race`, I think we can continue as is.

As for the `Promise.Resolve`, I am not sure. The difference here is that `Promise.race` takes an array of promises, which is currently handled by `PromiseGroup`, and the `Promise.Resolve` will take a single value.

PR #901, opened 2025-11-25
chore(heap): Use SoAVec for Set storage

Reinsdyret:
Migrated set storage in heap from Vec<SetHeapData> to SoAVec<SetHeapData>.


aapoalas:
Some lints to fix at least.

PR #907, opened 2025-11-28
fix(ecmascript): Fix BigInt and Number equality comparison precision

baseballyama:
The current implementation converted both `BigInt` and `Number` to `f64` for comparison. This caused precision loss for large `BigInt` because `f64` only has 53 bits of precision. Instead of converting `BigInt` to `f64` (which loses precision), convert `f64` to an exact `BigInt` representation and compare BigInts directly.

### DIFF
--- a/tests/metrics.json
+++ b/tests/metrics.json
@@ -1,10 +1,10 @@
 {
   "results": {
-    "crash": 72,
-    "fail": 7405,
-    "pass": 39875,
-    "skip": 3325,
-    "timeout": 19,
+    "crash": 70,
+    "fail": 7451,
+    "pass": 39831,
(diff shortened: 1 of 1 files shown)


aapoalas (dismissed):
Hello, and welcome to the Nova project! Thank you for your time, effort, and code!

LGTM, great work and great commenting! This was a pleasure to read through <3 Absolutely stellar stuff!

aapoalas:
So it turns out that my perf improvement didn't make the test fast enough and I opted to add it to skipped tests instead. Oh well, my next PR _should_ finally resolve that issue :) But in the mean time, this PR is great, good to go, and I'll merge it once the tests pass.

PR #910, opened 2025-11-28
chore(just): Add just

tbashiyy:
# Background
When I execute commands like `cargo build --profile dev-fast && cargo run --bin test262 --profile dev-fast -- -u`, I check https://github.com/trynova/nova/blob/main/CONTRIBUTING.md everytime because I cannot remember commands.

# Proposal
Introduce just for improving developer experience when executing commands like build,run test262, and so on...
If we can use just, we can check essential commands for contributing like below :).

https://github.com/user-attachments/assets/5831bed8-79df-4733-bd12-a595e8a85177


tbashiyy:
OK!
I will open another PR for adding justfile to .gitignore!

PR #913, opened 2025-12-08
feat(linter): Rule which warns when you could've used `NoGcScope`

eliassjogreen:
A lint which checks if you are taking `GcScope` despite only needing `NoGcScope`. Fixed all of the edge cases I found until the engine had no false-positives and fixed a few places in the vm where we used `GcScope` where we could've used `NoGcScope`. I left two places with the `GcScope` because I was unsure if it needed it because of async or as a placeholder for the future, these are excluded with a `#[allow]` directive.


aapoalas (approved):
LGTM, thank you and this is absolutely freaking excellent <3

PR #916, opened 2025-12-14
refactor(cli): Librarify `nova_cli` to avoid code duplication

niluxv:
Right now this is mostly moving code around, so the benchmark runner (which I have in a separate branch) can share most of its code with the cli. I have no idea whether this library could be useful for users of nova as well, I don't really know their requirements, but I assume the current setup is not flexible enough for that. (For example, currently everything assumes there is only one realm.)

The benchmark runner needs to separately parse and execute scripts, while the cli can just immediately execute a parsed script, which is why I haven't touched the script parsing and execution logic of the cli. However, maybe we can store a list of parsed but not yet executed scripts in `HostDefined` (alongside the module map of already executed modules, or maybe merge them?), and then implement convenience commands for script/module parsing and execution for the `Instance`. Or maybe that is overly complicated and it is better to leave this work to the library user.

I would love to hear any suggestions, as I find myself really struggling to provide a flexible and convenient API.

### DIFF
--- a/nova_cli/src/lib/host_hooks.rs
+++ b/nova_cli/src/lib/host_hooks.rs
@@ -1,0 +1,190 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! The [`HostHooks`] implementation for the main thread.
+
+use std::{
+    cell::RefCell, collections::VecDeque, fmt::Debug, path::PathBuf, rc::Rc, sync::mpsc, thread,
+    time::Duration,
+};
+
+use nova_vm::{
+    ecmascript::{
+        Agent, ExceptionType, GraphLoadingStateRecord, HostDefined, HostHooks, Job, ModuleRequest,
+        Referrer, SharedDataBlock, String as JsString, finish_loading_imported_module,
+        parse_module,
+    },
+    engine::{Bindable, Global, NoGcScope},
+};
+
+pub enum HostToChildMessage {
+    Broadcast(SharedDataBlock),
+}
+
+pub enum ChildToHostMessage {
+    Joined,
+    Report(String),
+}
+
+pub struct CliHostHooks {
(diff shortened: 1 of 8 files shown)


nova_cli/src/lib/host_hooks.rs
```diff
@@ -0,0 +1,216 @@
...
+        agent: &mut Agent,
+        referrer: Referrer<'gc>,
+        module_request: ModuleRequest<'gc>,
+        _host_defined: Option<HostDefined>,
+        payload: &mut GraphLoadingStateRecord<'gc>,
+        gc: NoGcScope<'gc, '_>,
+    ) {
+        let specifier = module_request.specifier(agent);
+        let specifier = specifier.to_string_lossy(agent);
+        let specifier_target = if let Some(specifier) = specifier.strip_prefix("./") {
```
aapoalas: **thought:** I wonder if some of this code would belong in the ModuleMap now?

aapoalas (approved):
This looks good to me: I don't unfortunately have any particularly good ideas about what a more generic API would look like. Storing parsed-but-not-executed Scripts could indeed be done using `Global<Script>`s, but I don't have a handle on if that would work nicely for a benchmarking CLI or not.

I'd be happy to merge this as-is; I kind of expect that figuring out the correct API is something that can only be done through experiment.

nova_cli/src/lib/host_hooks.rs
aapoalas: I forgot to say: sorry it took so long for me to review this <3

aapoalas:
> Ok, indeed we can further experiment after this is merged; the no semver guarantee gives us flexibility there. And it would be quite a nightmare to merge/rebase this if someone makes significant changes to the cli code in the meantime...
> 
> How about I rebase this on top of main so we lose the merge commit, and squash my two commits together?

Sounds good to me!

niluxv:
>> How about I rebase this on top of main so we lose the merge commit, and squash my two commits together?
>
> Sounds good to me!

Done.

aapoalas:
@niluxv Sorry it took me so long to get back to you! I rebased the code again and will merge once the tests show the all green.

Thank you so much for this, this is great! <3

PR #917, opened 2025-12-15
fix(ecmascript): Property descriptor of `globalThis`

eliassjogreen:
Fixes this 262 test: https://github.com/tc39/test262/blob/main/test/built-ins/global/property-descriptor.js


aapoalas (approved):
LGTM, nice to find little things like this :)

PR #976, opened 2026-03-26
Handle Builtin Atomics Async Critical Section in the Main Thread

komyg:
Fixes this comment: https://github.com/trynova/nova/pull/956#pullrequestreview-3898474204

### DIFF
--- a/nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
+++ b/nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
@@ -1699,14 +1699,16 @@ fn enqueue_atomics_wait_async_job<const IS_I64: bool>(
     // 1. Let timeoutJob be a new Job Abstract Closure with no parameters that
     //    captures WL and waiterRecord and performs the following steps when
     //    called:
-    // let data_block = data_block.clone();
     let signal = Arc::new(AtomicBool::new(false));
+    let s = signal.clone();
     let handle = thread::spawn(move || {
         // SAFETY: buffer is a cloned SharedDataBlock; non-dangling.
         let waiters = unsafe { data_block.get_or_init_waiters() };
         let waiter_record = WaiterRecord::new_shared();
         let mut guard = waiters.lock().unwrap();
 
+        // Signal the main thread that we have the lock and are about to sleep.
+        s.store(true, StdOrdering::Release);
         guard.push_to_list(byte_index_in_buffer, waiter_record.clone());
 
         if t == u64::MAX {
--- a/nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
+++ b/nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
@@ -5,10 +5,7 @@
 use std::{
     hint::assert_unchecked,
     ops::ControlFlow,
-    sync::{
-        Arc,
-        atomic::{AtomicBool, Ordering as StdOrdering},
-    },
+    sync::Arc,
     thread::{self, JoinHandle},
     time::Duration,
 };
@@ -1421,6 +1418,7 @@ fn do_wait_critical<'gc, const IS_ASYNC: bool, const IS_I64: bool>(
     gc: NoGcScope<'gc, '_>,
 ) -> Value<'gc> {
     let slot = buffer.as_slice(agent).slice_from(byte_index_in_buffer);
+    let data_block = buffer.get_data_block(agent).clone();
     // 14. Let WL be GetWaiterList(block, byteIndexInBuffer).
     // 15. If mode is sync, then
     // a. Let promiseCapability be blocking.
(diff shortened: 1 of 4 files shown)


nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
```diff
@@ -1485,28 +1485,30 @@ fn do_wait_critical<'gc, const IS_ASYNC: bool, const IS_I64: bool>(
     // }.
     // 28. Perform AddWaiter(WL, waiterRecord).
     // 29. If mode is sync, then
+
+    let data_block = buffer.get_data_block(agent);
+
+    // Re-read value under critical section to avoid TOCTOU race.
```
aapoalas: **issue:** You're not within the critical section here, and we've already performed the same read above anyway.

aapoalas (changes requested):
Yeah, this doesn't unfortunately really do what is advertised or requested by the specification.

The point is that we should be doing the following steps for the async case especially:
1. Acquire WaiterList lock.
2. Read value from buffer.
3. Compare value to expected and return NotEqual if it is not equal.
4. Produce a new waiter and add it into the WaiterList.
5. Return a Promise tied to the new waiter.

Or something like that. Instead we do:
1. Read value from buffer, compare and return NotEqual if not equal.
2. Start a new thread and wait until the new thread has started up.
3. On the new thread, acquire the WaiterList lock and tell the main thread we've started up.
4. Still on the new thread, produce a new waiter and add it into the WaiterList.
5. On the main thread return a Promise tied to the new thread.

The issue is that between steps 1 and 3 it's possible for some foreign thread to change the value to not be our expected value, acquite the WaiterList lock, and wake up all waiters. Our `waitAsync` call will not be woken up by this because the foreign thread got to the WaiterList first, we are not in it yet, but we also managed to read the old expected value from the buffer. This ends up with our `waitAsync` Promise never resolving.

The fundamental mistake that I made in the original implementation was using a native thread for the waiter: that makes it really hard (or slow) to have the read happen within the critical section. At the end of the day the correct choice should be to leave the threading and timeouting to the embedder.

komyg:
Hey @aapoalas, thank you for the feedback and for the explanation.

I've updated the PR to try to handle the critical sections properly. Please let me know what you think.

Next, I am planning to finish the implementation so that we don't depend on threads, as you said.

aapoalas:
> Hey @aapoalas, thank you for the feedback and for the explanation.
> 
> I've updated the PR to try to handle the critical sections properly. Please let me know what you think.
> 
> Next, I am planning to finish the implementation so that we don't depend on threads, as you said.

Unfortunately the critical section handling still isn't fully correct. The reading of the data is now within the critical section (I think - I should bring the code in locally and really think it through to be sure), but there is still a race condition. What's happening now is:

1. Acquire the WaiterList lock on the main thread.
2. Read and compare value to expected, return `NotEqual` if we don't find the expected value. This is correct and good.
3. Create a new waiter and add it into the WaiterList.
4. Start a new thread, release the WaiterList lock, queue up a Job, and return a Promise..
5. On the new thread, acquire the WaiterList lock and go to sleep awaiting for a wake up signal.

The issue is that the steps 4 and 5 must happen in this order as the new thread cannot acquire the lock before the main thread has released it, and/but there is no guarantee that the new thread is actually going to be the thread that acquires the lock immediately after the new thread. Hence, the critical section actually ends after step 4, before the new thread is waiting for wakeup signals.

It is therefore possible that we reach the end of step 4, have determined that the value is the expected value, and that we should sleep, but then a foreign thread acquires the WaiterList lock before our waiter thread does, and changes the value and signals a wakeup. Then our waiter thread acquires the lock, goes to sleep and releases the lock again: but now it's entirely possible that no weakup call ever comes again; we missed the signal that was supposed to wake us up and we're now effectively at a deadlock.

I think it's _fine_ to merge this work, but it's definitely still not what the spec requires. The extra thread keeps harassing us :)

komyg:
> > Hey @aapoalas, thank you for the feedback and for the explanation.
> > 
> > I've updated the PR to try to handle the critical sections properly. Please let me know what you think.
> > 
> > Next, I am planning to finish the implementation so that we don't depend on threads, as you said.
> 
> Unfortunately the critical section handling still isn't fully correct. The reading of the data is now within the critical section (I think - I should bring the code in locally and really think it through to be sure), but there is still a race condition. What's happening now is:
> 
> 1. Acquire the WaiterList lock on the main thread.
> 2. Read and compare value to expected, return `NotEqual` if we don't find the expected value. This is correct and good.
> 3. Create a new waiter and add it into the WaiterList.
> 4. Start a new thread, release the WaiterList lock, queue up a Job, and return a Promise..
> 5. On the new thread, acquire the WaiterList lock and go to sleep awaiting for a wake up signal.
> 
> The issue is that the steps 4 and 5 must happen in this order as the new thread cannot acquire the lock before the main thread has released it, and/but there is no guarantee that the new thread is actually going to be the thread that acquires the lock immediately after the new thread. Hence, the critical section actually ends after step 4, before the new thread is waiting for wakeup signals.
> 
> It is therefore possible that we reach the end of step 4, have determined that the value is the expected value, and that we should sleep, but then a foreign thread acquires the WaiterList lock before our waiter thread does, and changes the value and signals a wakeup. Then our waiter thread acquires the lock, goes to sleep and releases the lock again: but now it's entirely possible that no weakup call ever comes again; we missed the signal that was supposed to wake us up and we're now effectively at a deadlock.
> 
> I think it's _fine_ to merge this work, but it's definitely still not what the spec requires. The extra thread keeps harassing us :)

I see, yeah sorry I didn't think of this case.

If it is alright with you, I think we can merge this and I will make a follow up PR that fixes this and the thread issue. What do you think?

aapoalas:
Sure thing!

aapoalas (approved):

PR #984, opened 2026-04-19
feat(ecmascript): PlainTime.prototype.toJSON

jesperkha:


nova_vm/src/ecmascript/builtins/temporal/plain_time/plain_time_prototype.rs
```diff
@@ -228,14 +238,35 @@ impl TemporalPlainTimePrototype {
...
 
+    /// ### [4.3.18 Temporal.PlainTime.prototype.toJSON ( )](https://tc39.es/proposal-temporal/#sec-temporal.plaintime.prototype.tojson)
+    fn to_json<'gc>(
+        agent: &mut Agent,
+        this_value: Value,
+        _args: ArgumentsList,
+        gc: GcScope<'gc, '_>,
+    ) -> JsResult<'gc, Value<'gc>> {
+        // 1. Let plainTime be the this value.
+        let value = this_value.bind(gc.nogc());
```
aapoalas: **nitpick:** Could start with `let gc = gc.into_nogc();`.

aapoalas (dismissed):
LGTM but fixing the nitpick would maybe be a nice cleanup: you can get rid of all the unbinds as well that way.

PR #988, opened 2026-04-19
feat(ecmascript): PlainTime.prototype.round

0Zeno:


nova_vm/src/ecmascript/builtins/temporal/plain_time/plain_time_prototype.rs
```diff
@@ -228,6 +238,98 @@ impl TemporalPlainTimePrototype {
...
+        };
+
+        // 10. Perform ? ValidateTemporalUnitValue(smallestUnit, time).
+        // 11. Let maximum be MaximumTemporalDurationRoundingIncrement(smallestUnit).
+        // 12. Assert: maximum is not unset.
+        // 13. Perform ? ValidateTemporalRoundingIncrement(roundingIncrement, maximum, false).
+        // 14. Let result be RoundTime(plainTime.[[Time]], roundingIncrement, smallestUnit, roundingMode).
+        let result = plain_time
+            .get(agent)
+            .inner_plain_time(agent)
```
aapoalas: **nitpick:** You could get this before all the `reborrow` work and store it on the stack where it cannot be GC'd and is safe: that'd mean probably no need to scope `plain_time`.

aapoalas (dismissed):
LGTM!

nova_vm/src/ecmascript/builtins/temporal/plain_time/plain_time_prototype.rs
0Zeno: Should be fixed now

PR #989, opened 2026-05-07
Remove OS Thread from Atomics Wait Async

komyg:


nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
```diff
@@ -1621,52 +1615,106 @@ fn create_wait_result_object<'gc>(
...
-    // but currently without the trait. The job trait will be added eventually
-    // and we can get rid of this lint exception.
-    #[allow(unknown_lints, can_use_no_gc_scope)]
-    pub(crate) fn run<'gc>(self, agent: &mut Agent, gc: GcScope) -> JsResult<'gc, ()> {
+    pub(crate) fn run<'gc>(self, agent: &mut Agent, gc: GcScope<'gc, '_>) -> JsResult<'gc, ()> {
         let gc = gc.into_nogc();
+
+        // SAFETY: buffer is a cloned SharedDataBlock; non-dangling.
+        let waiters = unsafe { self.0.data_block.get_or_init_waiters() };
+        // a. Perform EnterCriticalSection(WL).
```
aapoalas: **nitpick (but blocking):** I'd like to see here a reference to where these steps are coming from. Possibly also a link to the relevant spec method.

nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
```diff
@@ -1621,52 +1615,106 @@ fn create_wait_result_object<'gc>(
...
+            }
+            WaitResult::TimedOut => {
+                unwrap_try(promise_capability.try_resolve(
+                    agent,
+                    BUILTIN_STRING_MEMORY.timed_out.into(),
+                    gc,
+                ));
+            }
+        }
+        // c. Perform LeaveCriticalSection(WL).
```
aapoalas: **nitpick (blocking):** Step b is missing. I assume that'd be removing the waiter from the list, but I'd want to see it.

nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
```diff
@@ -1621,52 +1615,106 @@ fn create_wait_result_object<'gc>(
...
+        let promise_capability = PromiseCapability::from_promise(promise, true);
+        match result {
+            WaitResult::Ok => {
+                unwrap_try(promise_capability.try_resolve(
+                    agent,
+                    BUILTIN_STRING_MEMORY.ok.into(),
+                    gc,
+                ));
+            }
+            WaitResult::TimedOut => {
```
aapoalas: **thought:** The string could be created from some `WaitResult::to_string` method to combine the two paths into a single `try_resolve(agent, result.to_string(), gc)`.

nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
```diff
@@ -1621,52 +1615,106 @@ fn create_wait_result_object<'gc>(
...
+
+impl WaitAsyncTimeoutJob {
+    pub(crate) fn run<'gc>(self, _agent: &mut Agent, _gc: GcScope<'gc, '_>) -> JsResult<'gc, ()> {
+        if self.0.waiter_record.get_result().is_some() {
             return Ok(());
-        };
+        }
+
+        // SAFETY: buffer is a cloned SharedDataBlock; non-dangling.
+        let waiters = unsafe { self.0.data_block.get_or_init_waiters() };
```
aapoalas: **nitpick:** This seems like it could be a helper method.

aapoalas (changes requested):
Generally looks good to me, but I had a few nitpicks regarding the inline comments that I'd like to fix before merging :)

nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
komyg: So this is the implementation of the Job Abstract Closure for [WaitAsyncTimeoutJob](https://tc39.es/ecma262/#sec-enqueueatomicswaitasynctimeoutjob), when there is no timeout. 

AFAIK the spec only defines the case with a timeout, but we have separated them, so I don't think that there is a specifc spec reference that I can add.

Therefore, I removed the comments referencing the steps.

nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
komyg: See here: https://github.com/trynova/nova/pull/989#discussion_r3384479162

nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
komyg: Since this is an `unsafe` operation, IMO it would be better to have this inline, so that this fact is not hidden from the user by an abstraction.

komyg:
Hey @aapoalas, I've addressed your comments. Can you review again, please?

nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
aapoalas: That is the exactly wrong way of thinking about this and Rust safety guarantees in general, as I see it: safe wrapper functions around `unsafe` things makes local reasoning possible, and local reasoning makes code easy to reason about in general.

If an unsafe action cannot be guaranteed to be safe with local reasoning only, then we need to perform global reasoning (which is generally hard), and we then have to document our reasoning in the `SAFETY` comment.

If, on the other hand, we're sure that this _is_ safe without any global arguments then the SAFETY comment is noise, and the unsafety is noise. No user wants to see eg. SAFETY comments on `Vec::push`, they want safe APIs.

In this case here AFAIU the `WaitAsyncTimeoutJobInner`  is only ever created with a non-dangling SDB, meaning that `get_or_init_waiters` is always safe on it. There is 0 unsafety here, and needing the `unsafe` blocks and SAFETY comments is just visual noise.

aapoalas (changes requested):
I disagree about the safe helper method, but otherwise this looks good to :)

nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
komyg: Hum... I never thought of this this way, you raise a good point.

I've added the helper function.

komyg:
> I disagree about the safe helper method, but otherwise this looks good to :)

Thanks @aapoalas ! I've added the helper function you requested. Could you review again, please?

aapoalas:
Something in lint is failing :(

aapoalas:
Oh shit, dylint... Hmm

komyg:
Hey @aapoalas, I had to update the Rust toolchain to make the Dylint work. 

All steps are passing now. Could you review again, please?

nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
```diff
@@ -1621,52 +1620,94 @@ fn create_wait_result_object<'gc>(
...
 
-    // NOTE: The reason for using `GcScope` here even though we could've gotten
-    // away with `NoGcScope` is that this is essentially a trait impl method,
-    // but currently without the trait. The job trait will be added eventually
-    // and we can get rid of this lint exception.
-    #[allow(unknown_lints, can_use_no_gc_scope)]
-    pub(crate) fn run<'gc>(self, agent: &mut Agent, gc: GcScope) -> JsResult<'gc, ()> {
+    /// Implementation of the Job Abstract Closure for [WaitAsyncTimeoutJob](https://tc39.es/ecma262/#sec-enqueueatomicswaitasynctimeoutjob),
+    /// for the cases where no timeout is specified.
+    pub(crate) fn run<'gc>(self, agent: &mut Agent, gc: GcScope<'gc, '_>) -> JsResult<'gc, ()> {
```
eliassjogreen: **nitpick**: This is throwing a dylint warning about it being possible to use `NoGcScope` now. Is the old comment and lint ignore still relevant or could you switch to `NoGcScope`?

nova_vm/src/ecmascript/builtins/structured_data/atomics_object.rs
```diff
@@ -1621,52 +1620,94 @@ fn create_wait_result_object<'gc>(
...
+struct WaitAsyncTimeoutJobInner {
+    data_block: SharedDataBlock,
+    byte_index_in_buffer: usize,
+    waiter_record: Arc<WaiterRecord>,
+}
+
+pub(crate) struct WaitAsyncTimeoutJob(Box<WaitAsyncTimeoutJobInner>);
+
+impl WaitAsyncTimeoutJob {
+    pub(crate) fn run<'gc>(self, _agent: &mut Agent, _gc: GcScope<'gc, '_>) -> JsResult<'gc, ()> {
```
eliassjogreen: **nitpick**: Same as the other comment, this is throwing a dylint warning now.

eliassjogreen (changes requested):

aapoalas (dismissed):
LGTM: I'll fix the dylints, push to this PR, and merge once tests show all-green. Thanks for upgrading dylint <3 That's always something I'm afraid of doing :D

